PR #6, opened 2016-11-29
update installation to new r-pkgs home

dpastoor:


hadley:
Thanks - I fixed that and a couple of other urls myself.

PR #26, opened 2017-06-06
explicitly import testthat

dpastoor:
when using use_testthat on a new computer, fails as testthat not available. Suggest moving testthat to Imports so it will be installed


hadley:
I'm pretty sure it should be in suggests. Can you please file an issue if you believe otherwise?

PR #31, opened 2017-07-21
handle windows absolute paths in escape_path

dpastoor:


hadley:
Would you mind adding a test?

R/ignore.R
```diff
@@ -25,5 +25,7 @@ use_build_ignore <- function(files, escape = TRUE, base_path = ".") {
 escape_path <- function(x) {
   x <- gsub("\\.", "\\\\.", x)
   x <- gsub("/$", "", x)
+  # maintain absolute path on windows to keep double \\ even after escaping
+  x <- gsub("(^\\D:)", "\\1\\\\", x)
```
hadley: I can't really follow the logic here

dpastoor: this was a hacky? way of getting around the original gsub problems, discussed in the pkgdown issue. Basically on windows, this escape_path function previously would escape in such a way that the resultant path would be like

`C:\Users\hadley\...`

this causes the PCRE regex compilation error, which wants

`C:\\Users\hadley\...`

thus this regex will replace, if it detects a windows absolute path delineated by `<Driveletter>:\` and replace it with `<Driveletter>:\\` such that the regex doesn't explode.

dpastoor:
this may be too much of a code smell to accept, especially if you are aware of a more elegant way to handle this.

In regards to a test, if this still seems reasonable, I would be happy to add some tests. One question is, would you like a windows only 'real' test that actually grabs a full path, runs it through this regex and introspects it? I am a little unsure how best to implement this such that it would be run against any persons computer - eg on appveyor. 

It would be easy to add some fake paths, like `C:\\User\\Hadley` that are representative of windows paths and could be tested without directly touching the file system. Would you consider that sufficient?

hadley:
Yes, that would be perfect. `escape_path()` is an isolate component that can be easily tested.

dpastoor:
@hadley ready for a review. I can also squash these to a single commit if you prefer?

R/ignore.R
```diff
@@ -22,5 +22,14 @@ use_build_ignore <- function(files, escape = TRUE, base_path = ".") {
...
   x <- gsub("/$", "", x)
+
+  # at this point the string replacement for an absolute windows path would
+  # approximately be: C:\Users\hadley\...
+  # this causes the PCRE regex compilation error, which wants
+  # C:\\Users\hadley\...
+  #thus this regex will replace, if it detects a windows absolute path delineated
+  # by <Driveletter>:\ and replace it with <Driveletter>:\\ such that the regex
+  # doesn't explode.
+  x <- gsub("(^\\D:)", "\\1\\\\", x)
```
hadley: I think `[A-Z]` would be clearer than `\\D`, and I'd usually put the anchor outside of the parens.

tests/testthat/test-escape_path.R
```diff
@@ -0,0 +1,27 @@
+context("escape_path")
```
hadley: My convention is to name the test file the same as the R file - so this should be `test-ignore.R`. Can you please change?

R/ignore.R
```diff
@@ -22,5 +22,14 @@ use_build_ignore <- function(files, escape = TRUE, base_path = ".") {
 escape_path <- function(x) {
   x <- gsub("\\.", "\\\\.", x)
   x <- gsub("/$", "", x)
+
+  # at this point the string replacement for an absolute windows path would
+  # approximately be: C:\Users\hadley\...
+  # this causes the PCRE regex compilation error, which wants
+  # C:\\Users\hadley\...
```
hadley: I still don't understand why you need an extra `\` here. Can you please rewrite the comment to specifically to address that? Ideally it would be at most 3 lines.

dpastoor: Given the pattern `<driveletter>:\user\...`, you get a PCRE regex compilation error. This is evident in https://github.com/hadley/pkgdown/issues/307 which was impetus for this PR

dpastoor:
@hadley merged all upstream changes, and addressed your comments

R/ignore.R
```diff
@@ -22,5 +22,10 @@ use_build_ignore <- function(files, escape = TRUE, base_path = ".") {
 escape_path <- function(x) {
   x <- gsub("\\.", "\\\\.", x)
   x <- gsub("/$", "", x)
+
+  # pattern `<driveletter>:\user\...` causes the PCRE regex compilation error,
```
hadley: Why? I don't understand what the problem is

dpastoor: When normalizePath() is called upstream, given a full path, plus when it goes through escape path, the resulting string will be written to the .Rbuildignore:

`C:\User\hadley\...`

per the discussion https://github.com/hadley/pkgdown/issues/307, The next time the .Rbuildignore is parsed, such as when calling `devtools::build()` it will present with the following error (taken from #pkgdown/307) 

```
Error in grepl(e, files, perl = TRUE, ignore.case = TRUE) : 
  invalid regular expression '^C:\Users\gabriel_odom\Documents\GitHub\mvMonitoring\mvMonitoring\docs$'
Execution halted
```

In order to have a valid regex, we need the first driveletter to have a double slash `C:\\...`

Therefore, line 29 will now add back an extra `\` if the pattern is detected.

hadley: Why is it only the drive letter that needs the extra escape?

Why isn't line 23 already adding the extra escapes?

hadley:
The root cause is that we're generating absolute paths on windows, not that we're failing to escape correctly. That also appears to be a problem, but I don't think that your patch is the correct fix.

dpastoor:
thats fair enough, just want to get it fixed in some capacity for windows users :-)

PR #53, opened 2017-08-25
Use clipr to add code to clipboard

### LINKED ISSUE #52: integrate with clipr?
It might be useful for things where you have to copy paste if the text was automatically added to the clipboard with [clipr](https://github.com/mdlincoln/clipr).

It is true if there is more than one thing to copy paste only the last one would be on the clipboard, but at least it is something. Also some people (me at least) use utilities that can save and retrieve clipboard history.

jimhester:
This makes it easy to paste into the source files where needed.

It looks like all current uses of `code_block()` are places where it makes sense to put the code on the clipboard. If we think this may change in the future we could either have an option to `code_block()`, or have a helper function which calls `clipr::write_clip()` and `code_block()` internally and switch the current uses of `code_block()` to the helper.

Fixes #52

### DIFF
--- a/R/style.R
+++ b/R/style.R
@@ -16,6 +16,7 @@ code_block <- function(...) {
   block <- paste0("  ", c(...), collapse = "\n")
   if (clipr::clipr_available()) {
     clipr::write_clip(paste0(c(...), collapse = "\n"))
+    message("copied to clipboard:")
   }
   cat_line(crayon::make_style("darkgrey")(block))
 }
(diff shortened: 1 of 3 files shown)


R/style.R
```diff
@@ -14,6 +14,9 @@ done <- function(...) {
 
 code_block <- function(...) {
   block <- paste0("  ", c(...), collapse = "\n")
+  if (clipr::clipr_available()) {
+    clipr::write_clip(paste0(c(...), collapse = "\n"))
```
hadley: Maybe need message here to say "copied to clipboard"?

hadley:
Thanks!

Issue #48, opened 2017-08-16
Badging for project lifecycle


cboettig:
I greatly enjoy using the `use_*` commands to get the markdown for my README badges.  Curious what your team thinks of the http://www.repostatus.org/ badges, and if you would consider a l'lil function to generate the appropriate badge.  I know there's some risk of these becoming stale since they are just static badges, but personally I find the categories quite helpful, particularly on the sunset era of a package.

hadley:
I like the idea, but I don't love repostatus - it just feels a bit too fine grained to me.

Another option is https://github.com/badges/stability-badges but they don't feel quite right for the R community.

cderv:
Just for pointing other interesting links. Here is an initiative for stability badges for R : https://github.com/joethorley/stability-badges based on shield.io services.

cboettig:
The `stability-badges` makes sense to me once I realized I should think of them as package lifecycle terms, but I think `stability` is the wrong term.  I see the motivation: e.g. `plyr` would be `Locked`, in the sense that commits would only patch bugs; perhaps closest to what repostatus calls `Inactive`, and `unstable` is approximately `WIP`, e.g. pre-CRAN release.  

Yet when I first looked at stability-badges, I was greatly confused because I was not thinking of the in lifecycle terms but just as a generic indicator of the API.  I think the `unstable` & `stable` are the most confusing:  on the face of it, it sounds like `stable` is a pledge not never to break the API and never to bump the major version.   Closer inspection shows that it does not make that claim, but then it becomes a bit unclear how this differs from an unstable package -- is an `unstable` package just a `stable` package that fails to bump the major version when it makes such a change?  (This also gnaws at my dislike of semantic versioning -- sure there are big changes and small changes, changes that do or don't alter the API; but any change can break things, and I think the promise that "these changes won't hurt" is misleading or dangerous).  

 Focusing on the term *stability*, using `unstable` or `stable`, in my opinion, puts too much emphasis on classifying how the API changes and makes it easy to miss the lifecycle aspect (e.g. it is easy to miss the implication that `stable` is a more mature version of `unstable`, and instead think this is just a message of how much the maintainer gives a crap about the idea of backwards compatibility.).

If the those badges were called `lifecycle` badges instead of `stability` badges, maybe replacing `unstable`  with the more life-cycle focused term `beta` and `locked` with a term that better indicates "this is an old package that gets limited maintenance priority / only 'security' updates / bug patches" I'd feel much better about it...

[...]

PR #56, opened 2017-08-28
Added Repo Status.

muschellij2:
Added `use_repo_status_badge` to the `NEWS.md` and the code related.  Uses standard badges from http://www.repostatus.org/.


hadley:
Per #48, I just don't love repostatus, so I'd like to discuss this more before selecting a service.

muschellij2:
Agreed - but is there a detriment to adding this function or would a better name be appropriate.  I was thinking something where you could have  an option like `service = c("repostatus", "stability-badge"))` similar to `code_coverage`.

hadley:
Unless there is strong motivation otherwise I'd prefer to be as prescriptive as possible.

Issue #68, opened 2017-09-11
Implement `use_depsy_badge`


gvegayon:
[Depsy](http://depsy.org/about) is a NSF funded project that consists on 

> [A] free website launched in November 2015 that aims to “measure the value of software that powers science”. ([nature, Jan 2016](http://www.nature.com/news/the-unsung-heroes-of-scientific-software-1.19100))

In simple terms, Depsy provides the __unsung heroes of scientific software__ with a simple way of measuring the value of their contribution to the scientific community.

In practice, Depsy ranks software (published on PyPI and CRAN) and provides information about usage and citation of those software projects. Ultimately, such rank can be reported on a project website with a badge, e.g. the `devtools` Depsy badge: [![Research software impact](http://depsy.org/api/package/cran/devtools/badge.svg)](http://depsy.org/package/r/devtools).

I think having a `use_depsy_badge` function would be nice. If you think so too, I can provide a PR with such functionality based on this issue as a wrapper of [use_badge](https://github.com/r-lib/usethis/blob/master/R/badge.R).

hadley:
A PR would be 🎉!

PR #70, opened 2017-09-14
New function `use_depsy_badge`

gvegayon:
This solves https://github.com/r-lib/usethis/issues/68

Nothing fancy, just the `use_depsy_badge` function, which was implemented based on `use_cran_badge`.


hadley:
Thanks!

hadley:
BTW if you use `Fixes #68` in a commit message the corresponding issue will get automatically closed when the PR is merged.

gvegayon:
Thanks! I'll use that in the future.

PR #75, opened 2017-09-26
Update license-mit.md

benmarwick:
Tidy up the MIT license text template so that it get the current year and the user's name in there when they run `usethis::use_mit_license("My Name")`


hadley:
Thanks!

PR #76, opened 2017-09-27
Minor typos

jonocarroll:
I noticed the `simulated` typo in the README and figured I'd run `devtools::spell_check()` over the rest, which caught a few more. 

`callr` was added to `DESCRIPTION` but this may be unnecessary.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,8 +20,7 @@ Imports:
     git2r,
     rmarkdown,
     rstudioapi,
-    whisker,
-    callr
+    whisker
 Suggests: 
     covr,
     testthat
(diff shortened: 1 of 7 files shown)


DESCRIPTION
```diff
@@ -20,7 +20,8 @@ Imports:
     git2r,
     rmarkdown,
     rstudioapi,
-    whisker
+    whisker,
```
hadley: Can you please remove this?

jonocarroll: Done. I didn't catch that it was only used in the example. Cheers.

hadley:
Thanks!

Issue #77, opened 2017-09-27
Coveralls doesn't work


muschellij2:
Getting pandoc error for coveralls badge

```
[![Coverage Status](https://img.shields.io/coveralls/muschellij2/mimosa.svg)](https://coveralls.io/r/muschellij2/mimosa?branch=master)
```

```
pandoc: Could not fetch https://img.shields.io/coveralls/muschellij2/mimosa.svg
TlsExceptionHostPort (HandshakeFailed Error_EOF) "img.shields.io" 443
Error: pandoc document conversion failed with error 67
```
when using `img.shields.io` from `use_coverage`.  

When using the coveralls badge, this doesn't happen:
```
[![Coverage status](https://coveralls.io/repos/github/muschellij2/mimosa/badge.svg)](https://coveralls.io/r/muschellij2/mimosa?branch=master)
```
there no error.  Same for if you put a branch:
```
[![Coverage status](https://coveralls.io/repos/github/muschellij2/mimosa/badge.svg?branch=master)](https://coveralls.io/r/muschellij2/mimosa?branch=master)
```

Submitting PR

PR #78, opened 2017-09-27
Fixing Coveralls Badge

muschellij2:
Fixing #77 badge and also added some newlines for reading.


DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: usethis
 Title: Automatically Set Up Commonly Used Package and Analysis Components
-Version: 0.1.0
+Version: 0.1.0.9000
```
hadley: Can you please revert this change?

NEWS.md
```diff
@@ -1,3 +1,8 @@
+# usethis 0.1.0.9000
+
+* Adding simple newlines for readability in `use_github` and fixing
```
hadley: I think this change is small enough and reflects stuff that's only in the devel version that there's no need for a news bullet.

Issue #86, opened 2017-10-10
Pass argument `open = ` to `use_template()` from wrappers


mixtrak:
Some functions to add useful infrastructure to a package also then open the files created. For example, use_news_md() will create and then open the news.md file, and similar behaviour is found in use_readme_rmd(), use_package_docs(), use_vignette() and possibly others.
I have written a wrapper function which initialises a new package with infrastructure and then switches to the new RStudio project in a new RStudio session. However, as a side-effect, all the files mentioned above get opened in my old RStudio session.
It would be nice to be able to pass an `open = FALSE` argument to the underlying code. For example, here is the code for use_package_doc:

```R
function (pkg = ".") 
{
    pkg <- as.package(pkg)
    use_template("packagename-package.r", file.path("R", paste(pkg$package, 
        "-package.r", sep = "")), data = list(name = pkg$package), 
        open = TRUE, pkg = pkg)
}
<environment: namespace:devtools>
```

Note the `open = TRUE` on the fourth line of function body code, which seems to be an argument of the internal {devtools} function devtools:::use_template. Perhaps this argument could be exposed to users of the higher-level functions.

hadley:
Would you be interested in a writing a pull request to this effect? You'd just need to make sure that every function that calls `use_template()` with `open = TRUE` gains an `open` argument so that you can control this behaviour.

mixtrak:
Would I! (yes)

mixtrak:
Pull request submitted. Only potential issue is some weird tests I made - I can revert that commit if you like, or you can do it on merge.

PR #91, opened 2017-10-14
Chooseopen: passing "open = T/F" arguments to use_template()

mixtrak:
As per issue #86. Some infrastructure-creation functions automatically open the resulting files in the current session, which is not always the desired behaviour. This is because these functions wrap `use_template()` with a hard-coded `open = TRUE` argument.

In this branch, I added `open = TRUE` arguments to the relevant wrapper functions, and replaced the hardcoded `open = TRUE` in the body of the functions with `open = open`.

The more recent commit is writing some automated tests to ensure these functions still execute without error.

### DIFF
--- a/R/cran.R
+++ b/R/cran.R
@@ -9,12 +9,12 @@
 #'
 #' @export
 #' @inheritParams use_template
-use_cran_comments <- function(base_path = ".") {
+use_cran_comments <- function(base_path = ".", open = TRUE) {
   use_template(
     "cran-comments.md",
     data = list(rversion = paste0(version$major, ".", version$minor)),
     ignore = TRUE,
-    open = TRUE,
+    open = open,
     base_path = base_path
   )
 
--- a/R/news.R
+++ b/R/news.R
@@ -4,11 +4,11 @@
 #'
 #' @inheritParams use_template
 #' @export
-use_news_md <- function(base_path = ".") {
+use_news_md <- function(base_path = ".", open = TRUE) {
   use_template(
     "NEWS.md",
     data = package_data(base_path),
-    open = TRUE,
+    open = open,
     base_path = base_path
   )
 }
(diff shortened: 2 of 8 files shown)


mixtrak:
Ugh. Did the build fail because of my weird tests? I can revert that commit if you like.

Also thanks for the opportunity to contribute. I wrote a [rambling blog post](https://mixtrak.tumblr.com/post/166397286686/r-package-development-fun) documenting my process and what I learned with all the messing around, on the off chance anyone's interested.

mixtrak:
Alrighty, updated and regenerated the Roxygen2 docs. Let me know if you want me to nuke the tests, or if they can be fixed.

R/cran.R
```diff
@@ -7,14 +7,15 @@
 #' other packages, you also need to describe the steps you have taken to
 #' ensure that it does not break existing code on CRAN.
 #'
+#' @param open Open the new file in the text editor. Defaults to TRUE.
```
hadley: It's better to not copy and paste documentation and instead document in one place and use `@inheritParams()`.

tests/testthat/test-use-template.R
```diff
@@ -13,3 +13,21 @@ test_that("find_template errors if template missing", {
   expect_error(find_template("xxx"), "Could not find template")
 })
 
+
+# use_template ------------------------------------------------------------
+
+test_that("no errors by passing open = FALSE to use_template wrappers", {
```
hadley: I don't think this test adds anything, so you can remove it.

mixtrak:
PR updated:
* Tests and documentation changes nuked with `git reset` to initial commit
* Docs re-generated with Roxygen2 and commit amended to include them. Since these functions inherit params from `use_template()` and now have the `open =TRUE` argument in their formals, it turns out I didn't need to edit docs in the first place.

hadley:
Thanks!

Issue #98, opened 2017-10-27
question: import UQ/UQS/!!/!!! in tidy_eval.R template?


ijlyttle:
I have been using the tidyeval framework in some of my packages and I have found it useful to import the unquoting operators, for internal package use,  something like this:

```
#' @importFrom rlang UQ !!
#' @importFrom rlang UQS !!!
```

As usual, I don't _really_ know what I'm doing - but could this be useful for the `tidy_eval.R` template?

If so, happy to make a PR.

hadley:
You should import `UQ` and `UQS` to avoid R CMD warnings. A PR would be great!

PR #101, opened 2017-10-31
adds import of UQ, UQS, !!, and !!!

ijlyttle:
Addresses #98

Added the syntactic-sugar versions, too - please let me know if they should not be there.

Also, seemed to be more than six functions there, so thought I would remove the count.

If you think it would be constructive to export these, I can amend to add

```
#' @export UQ UQS !! !!!
```

as well as a sentence to the descriptive paragraph.

### DIFF
--- a/inst/templates/tidy-eval.R
+++ b/inst/templates/tidy-eval.R
@@ -1,6 +1,6 @@
 #' Tidy eval helpers
 #'
-#' These six functions provide tidy eval-compatible ways to capture
+#' These functions provide tidy eval-compatible ways to capture
 #' symbols (`sym()`, `syms()`, `ensym()`), expressions (`expr()`,
 #' `exprs()`, `enexpr()`), and quosures (`quo()`, `quos()`, `enquo()`).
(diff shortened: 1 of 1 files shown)


inst/templates/tidy-eval.R
```diff
@@ -16,6 +16,7 @@
 #' @export           sym syms ensym
 #' @importFrom rlang expr exprs enexpr
 #' @export           expr exprs enexpr
+#' @importFrom rlang UQ UQS !! !!!
```
hadley: There's no need to import `!!` and `!!!` since those functions are never actually called

ijlyttle: Thanks for the explanation! Will amend accordingly.

ijlyttle:
I was going to leave a note saying things were ready to go, but no need ;)

hadley:
For future reference, if you include `Fixes #xyz` in a commit message (or the text of the PR), the corresponding issue will be automatically closed when the PR is merged.

PR #102, opened 2017-10-31
Set an explicit default width of 100% for README figs

jennybc:
Meant to combat figs that run off the page for README.html that CRAN creates and hosts.

Suggests the use of, e.g.,

```{r echo = FALSE}
knitr::include_graphics("man/figures/i-already-exist.png")
```

for pre-made figs.

### DIFF
--- a/inst/templates/omni-README
+++ b/inst/templates/omni-README
@@ -9,7 +9,8 @@ output: github_document
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>",
-  fig.path = "man/figures/README-"
+  fig.path = "man/figures/README-",
+  out.width = "100%",
 )
 ```
 {{/Rmd}}


hadley:
Do we really want them that wide?

jennybc:
Hmmm ... I just saw this as a huge improvement over the status quo in some of my projects.

I'll explain how I got here, in case it's all dysfunction on my end:

  * I often use quick and dirty measures to make a figure look better 😬. At the time of `ggsave()`, I make it physically larger than necessary, so that it looks nice and sharp wherever I need to use it later. Such as in a README. Ergo I need to specify figure width.
  * I often put two figure side-by-side via `fig.show = 'hold', out.width = '49%'`, which reinforced my choice of 49 + 49 = 98 ~= 100.

hadley:
Hmmm, maybe I'm getting confused about how they're sized by default. Can you point me to the diff for the gapminder README? (which was the motivation IIRC)

jennybc:
Sure:

https://github.com/jennybc/gapminder/commit/dd232a298a8bb66c984812a1762bf18c743ca9ff

jennybc:
Thanks! Even if 100% is not quite the thing, now the syntax is there in case someone wants to adjust, delete, or move to a specific chunk.

PR #110, opened 2017-11-10
revdep: git ignore .noindex files, used on macOS

gaborcsardi:
See https://github.com/r-lib/revdepcheck/issues/77
for details.


hadley:
Thanks!

PR #123, opened 2017-11-24
tidyeval template: removes `.env` from rlang imports

### LINKED ISSUE #122: tidyeval template: import .env from rlang
I don't have a strong opinion on this but I notice that (as of today) rlang does not export `.env` - so I think (as of today) the template should not import it.

I gather that this remains an open question (tidyverse/rlang#269), just going for consistency now.

Also making a PR for this.

ijlyttle:
Fixes #122 - rlang is not currently exporting `.env`

Also, corrects typo on `:=`. This was giving me a warning, removing single quotes brings this in line with syntax in pipe.R.


lionel-:
Could you redocument please?

ijlyttle:
I have redocumented, but there is no effect.

I think this is because the changes were made to a `.R` file in `inst/templates` rather than in `R`.

If there is documentation I need to add or revise, or if I am misunderstanding something, please let me know.

lionel-:
oops you're totally right, thanks! Could someone merge this please? I don't have commit rights for the r-lib org.

PR #136, opened 2017-12-19
Allow numbers in the name of .R files

### LINKED ISSUE #130: check_file_name contradicts itself
To create a new R file `R/abc.R`, I can run `use_r("abc")`, which works fine.

If the filename is to contain an integer, like `R/abc1.R`, `use_r("abc1")` trows an error

```
Error: 'abc1' is not a valid file name. It should:
* Contain only ASCII letters, numbers, '-', and '_'
```

This seems like a contradiction. What is the policy here? Are numbers in file names allowed or not?

jennybc:
Fixes #130

### DIFF
--- a/R/r.R
+++ b/R/r.R
@@ -25,7 +25,7 @@ check_file_name <- function(name) {
 }
 
 valid_file_name <- function(x) {
-  grepl("^[[:alpha:]_-]+$", x)
+  grepl("^[[:alnum:]_-]+$", x)
 }
 
 find_r_name <- function(name = NULL) {


jennybc:
Good example to help me get a feel for where you want more tests: I could add one or more tests around this but seems low risk / low priority. Confirm?

hadley:
More tests are always good, but low priority.

jennybc:
> More tests are always good, but low priority.

In this case, I dithered over whether to test the helper. Which seemed really silly. Or to test `use_r()`. But it's hard/impossible to test wrt typical usage, because the tricky bits are all about getting putative filename based on the active file. Hence no new test.

No reply necessary, just explaining how I'm thinking about it.

PR #138, opened 2017-12-20
Remove stray sentence fragment

jennybc:
housekeeping


jennybc:
I am drunk with power now.

PR #144, opened 2017-12-21
Add use_tidy_contributing() to tidyverse.R

### LINKED ISSUE #143: use_tidy_contributing
Assuming it’s _possible_ that someone would want to use `use_tidy_contributing()` for a non-tidyverse package (e.g. `r-lib`), it would make more sense to build the Travis URL in the same way it’s done for `use_travis()`
```r
url <- file.path("https://travis-ci.org", gh$username, gh$repo)
```
As opposed to hardcoding it as `http://travis-ci.org/tidyverse, name` (where name is the package name, and I just skip using the `gh` package altogether. 
_But_, it would be `gh$org` in the case of tidyverse and/or r-lib, right?

Alternately, I could use `remotes` from `desc::desc_get_remotes`, but, since I'm developing off of a fork, I can't tell whether or not that would actually work 😬!

batpigandme:
* fixes #143 
* uses tidy_contributing.md template

### DIFF
--- a/R/tidyverse.R
+++ b/R/tidyverse.R
@@ -18,6 +18,9 @@
 #' * `use_tidy_versions()`: pins all dependencies to require at least
 #'   the currently installed version.
 #'
+#' * `use_tidy_contributing()`: creates tidyverse contributing guidelines from
+#'    tidy_contributing template.
+#'
 #' @md
 #' @name tidyverse
 NULL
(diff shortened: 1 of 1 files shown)


hadley:
Can you please add a bullet to NEWS? It should briefly describe the change (starting with name of the function), and crediting yourself with `(@yourname, #issuenumber)`.

(HAH!)

hadley:
And re-document 😉

batpigandme:
@hadley Yeah— sorry I just got back from the vet. I actually did a better version of this on a diff. branch at like 6am. Let me PR that one instead.

PR #139, opened 2017-12-20
Add use_blank_slate()

jennybc:
Closes #133.

Implements ~`use_freshstarts()`~ `use_blank_slate()`, to suppress save/reload of user's workspace upon R restarts. 

Handles `scope = "project"` but not `scope = "user"`. Reasons given in https://github.com/r-lib/usethis/issues/133#issuecomment-352918047.

I needed to use work done in PRs #145 (Make scoped_temporary_package() easier to work with) and #146 (Less promiscuous Rbuildignoring) so those need approval first.

PR #145, opened 2017-12-22
Make scoped_temporary_package() easier to work with

### LINKED ISSUE #141: How to develop with `withr::defer()`
@hadley added this function, presumably because it's handy in the tests:

``` r
scoped_temporary_package <- function(dir = tempfile(), env = parent.frame()) {
  old <- proj$cur
  withr::defer(proj_set(old), envir = env)

  utils::capture.output(create_package(dir, rstudio = FALSE, open = FALSE))
  invisible(dir)
}
```

But when you're *developing* the test, you can't actually call `scoped_temporary_package()`. You get this:

```
 Error in withr::defer(proj_set(old), envir = env) : 
  attempt to defer event on global environment 
```

Which makes life rather difficult. I think I've run up against this in some other recent experiment with `withr::defer()`. What to do? Is there something that could or should be different about `withr::defer()`? Or should I finesse this inside `scoped_temporary_package()`?

cc @jimhester

jennybc:
Fixes #141: makes it possible to call `scoped_temporary_package()` in the console. Nicer for test development.

Move `scoped_temporary_package()` to `tests/testthat/helper.R`.

### DIFF
--- a/R/proj.R
+++ b/R/proj.R
@@ -61,14 +61,6 @@ proj_set <- function(path = ".", force = FALSE) {
   invisible(old)
 }
 
-scoped_temporary_package <- function(dir = tempfile(), env = parent.frame()) {
-  old <- proj$cur
-  withr::defer(proj_set(old), envir = env)
-
-  utils::capture.output(create_package(dir, rstudio = FALSE, open = FALSE))
-  invisible(dir)
-}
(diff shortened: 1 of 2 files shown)


jennybc:
Subsumed by #139

PR #146, opened 2017-12-22
Less promiscuous Rbuildignoring

### LINKED ISSUE #127: create_package doesn't work on CRAN
Because `.Rbuildignore` line added by RStudio also removes `template.Rproj`

jennybc:
Fixes #127 create_package doesn't work on CRAN

I just got bitten by this very hard, while writing tests for various function around RStudio Projects (coming soon, in a different PR). Relieved to get to the bottom of it! The `.Rproj` template wasn't surviving in the built package. The insidious part is I could pass tests with `test()` but then fail with `check()` ☹️.

I deliberately added a failing test first (https://github.com/r-lib/usethis/pull/146/commits/ae230172efa55efff3cc8ca10b6f18b1ae3d0140), then the fix for usethis itself.

I also went ahead and tightened up the line usethis adds to `.Rbuildignore` when it creates an RStudio project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 .Rhistory
 .RData
 revdep/.cache.rds
-.Rbuildignore
 revdep/checks
 revdep/library
 revdep/checks.noindex
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -1,4 +1,4 @@
-^.*\.Rproj$
+^usethis\.Rproj$
 ^\.Rproj\.user$
 ^revdep$
 ^\.travis\.yml$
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -2,7 +2,6 @@
 .Rhistory
 .RData
 revdep/.cache.rds
-.Rbuildignore
```
jennybc: Not sure why this was here but `.Rbuildignore` is being tracked, so it has had no effect anyway.

hadley:
I don't think this will work because RStudio automatically adds the promiscuous ignore when you re-open the project. Instead, I think we should just rename `template.Rproj` to have a different extension.

jennybc:
Subsumed by #139

PR #150, opened 2017-12-22
Adds use_tidy_contributing w/ check for travis

### LINKED ISSUE #143: use_tidy_contributing

batpigandme:
* Fixes #143 
* Add use_tidy_contributing() (uses `gh::gh_tree_remote()` for travis link)
* Not-exported, but adds `check_uses_travis()`
* Bullet in NEWS


batpigandme:
This passes `R CMD check` fine. I'm not sure how/if I turned appveyor on somehow (I've never used it before, but from the [build history](https://ci.appveyor.com/project/r-lib/usethis/history) it looks like @jennybc turned it on just a little while ago). Not sure if it's related to #146 at all, but FWICT, the `.Rproj` file causing the error (below) should be in the `.Rbuildignore` still.

`Project file "usethis.Rproj" is not a valid MSBuild project file which must be an XML.`

@hadley, not sure why this PR is on appveyor, but the previous wasn't ☹️!

hadley:
I think @jennybc broke appveyor, so don't worry about it.

hadley:
Thanks!

R/tidyverse.R
```diff
@@ -110,3 +113,22 @@ use_tidy_eval <- function() {
...
+
+
+#' @export
+#' @rdname tidyverse
+#' @inheritParams use_template
+use_tidy_contributing <- function() {
+  check_uses_travis()
+
+  name <- project_name()
+  gh <- gh::gh_tree_remote(base_path)
```
jennybc: What is `base_path` here? It's causing a new NOTE re: `no visible binding for global variable`.

batpigandme: Oh, shoot— sorry, one sec.

PR #151, opened 2017-12-22
Add use_tidy_issue_template()

batpigandme:
* Add minimal ISSUES.md to templates
* Add use_issues() to tidyverse.R


R/tidyverse.R
```diff
@@ -132,3 +134,14 @@ use_tidy_contributing <- function() {
     data = list(package = name, travis_url = travis_url)
   )
 }
+
+
+#' @export
+#' @rdname tidyverse
+#' @inheritParams use_template
+use_issues <- function() {
```
jennybc: Should this be checking if the project uses GitHub?

R/tidyverse.R
```diff
@@ -132,3 +134,14 @@ use_tidy_contributing <- function() {
...
+  use_template(
+    "ISSUES.md",
+    "ISSUE_TEMPLATE.md"
```
jennybc: Is there a reason to not just name the template itself `ISSUE_TEMPLATE.md`?

jennybc: FWIW the docs seem to feature `issue_template.md`. But I am sure it does not matter.

https://help.github.com/articles/creating-an-issue-template-for-your-repository/

jennybc: If the names differ, it should probably follow the convention of `tidy_*` files. In fact, the current PR seems to be `use_tidy_issues()`.

batpigandme: The use of `ISSUES.md` is a case of me being (possibly overly) literal! I found the same material you did, Jenny, re. the magic of `ISSUE_TEMPLATE.md`, but thought maybe there was a reason to use a different name for a template and its output (@hadley, not sure whether or not this _post hoc_ assessment is totally off…)
image

Totally agree re. `tidy_*` schema, and will change this to follow that (I think, initially, I was considering making it sufficiently general as to  be useful for non-tidyverse 📦).

hadley: We should make all the names consistent - i.e. `use_tidy_issue_template()` and `inst/templates/ISSUE_TEMPLATE.md`

At some point we should probably consider if (for the tidyverse) if these special files should all live in `.github`

PR #153, opened 2017-12-25
Windows-proof same_contents()

### LINKED ISSUE #148: Reconsider comparison method in same_contents()
Current way that `write_over()` evaluates whether proposed file content is same as current content:

https://github.com/r-lib/usethis/blob/30091521686cb8ffffd8f8d440a34db43f8741cf/R/write.R#L55-L63

I think this may not work on Windows (might explain [this failure](https://ci.appveyor.com/project/r-lib/usethis/build/1.0.2)). Will investigate more soon.

jennybc:
Fixes #148 Reconsider comparison method in `same_contents()`

Fixes `same_contents()` and `write_over()` on Windows, by keeping proposed text as a character vector. This avoids OS-specific line ending issues when deciding if proposed text is same as current. With this, all tests now pass on Windows.

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -22,7 +22,7 @@ write_union <- function(base_path, path, new_lines, quiet = FALSE) {
 }
 
 write_over <- function(base_path, path, contents) {
-  stopifnot(is.character(contents), length(contents) == 1)
+  stopifnot(is.character(contents), length(contents) > 0)
 
   full_path <- file.path(base_path, path)
   dir.create(dirname(full_path), showWarnings = FALSE)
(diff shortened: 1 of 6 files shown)


R/write.R
```diff
@@ -22,7 +22,7 @@ write_union <- function(base_path, path, new_lines, quiet = FALSE) {
 }
 
 write_over <- function(base_path, path, contents) {
```
hadley: Can you add a comment here, just to make clear that `content` should now be a vector?

jennybc: OK added this comment

``` r
## `contents` is a character vector of prospective lines
write_over <- function(base_path, path, contents) {
  ...
}
```

jennybc: Is this OK to merge? I'm going to try to bring the overall coverage up and it will be nice to have AppVeyor/Windows fully online.

PR #155, opened 2017-12-26
Set .here for projects that don't meet other criteria

### LINKED ISSUE #154: Make sure all new projects meet one project criterion
Re: self-consistency.

Should all new projects meet at least one of the criteria usethis uses to recognize a project?

https://github.com/r-lib/usethis/blob/973bcabceb36a73094d6a96094006b2a57fe0865/R/proj.R#L3-L9

In practice, I think this means creating a `.here` file whenever `rstudio = FALSE`.

cc @hadley

jennybc:
Fixes #154 Make sure all new projects meet one project criterion

Also added `is_proj()`. These are little odds and ends that look helpful as I write a bunch of tests, which will come in a separate PR.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -59,6 +59,8 @@ create_project <- function(path = ".",
 
   if (rstudio) {
     use_rstudio()
+  } else {
+    writeLines(character(), file.path(path, ".here"))
   }
   if (open) {
     open_project(path, name, rstudio = rstudio)
--- a/R/proj.R
+++ b/R/proj.R
@@ -15,6 +15,8 @@ proj_find <- function(path = ".") {
   )
 }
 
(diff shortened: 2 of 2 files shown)


R/create.R
```diff
@@ -59,6 +59,8 @@ create_project <- function(path = ".",
 
   if (rstudio) {
     use_rstudio()
+  } else {
+    writeLines(character(), file.path(path, ".here"))
```
hadley: Probably should eventually extract out into `use_here()` or similar, but no need for this PR.

jennybc: Yeah. That brings up whether usethis should embrace the here package in addition to rprojroot. As in, use rprojroot internally but deal more with here for user-facing stuff.

Issue #111, opened 2017-11-14
Add more tests


hadley:
Using `testthat::expect_known_output()` once available

hadley:
And make `yesno()` throw an error if not `interactive()`

PR #175, opened 2017-12-29
Rename yesno() to nope(); use our own interactive()

### LINKED ISSUE #142: Think about behaviour of `yesno()` inside tests

jennybc:
Fixes #142 (Think about behaviour of `yesno()` inside tests), where by "think about" I actually meant "make less annoying".

You also touched on this here (https://github.com/r-lib/usethis/issues/111#issuecomment-353196290), though I am not throwing an error. I'm just assuming the user has NOT authorized whatever action was being proposed. It feels like an error would be a net negative from the testing point of view.

I find `nope()` makes it a bit easier to remember that `TRUE` means "no" and `FALSE` means "yes".

The [branch where I'm accumulating tests](https://github.com/r-lib/usethis/tree/add-tests) passes with this.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -143,7 +143,7 @@ check_not_nested <- function(path, name) {
     stop(message, call. = FALSE)
   }
 
-  if (yesno(message, " This is rarely a good idea. Do you wish to create anyway?")) {
+  if (nope(message, " This is rarely a good idea. Do you wish to create anyway?")) {
     stop("Aborting project creation", call. = FALSE)
   }
 }
--- a/R/create.R
+++ b/R/create.R
@@ -14,7 +14,6 @@ create_package <- function(path = ".",
                            fields = getOption("devtools.desc"),
                            rstudio = rstudioapi::isAvailable(),
                            open = interactive()) {
-
   path <- normalizePath(path, mustWork = FALSE)
 
   name <- basename(path)
@@ -45,7 +44,6 @@ create_package <- function(path = ".",
 create_project <- function(path = ".",
                            rstudio = rstudioapi::isAvailable(),
                            open = interactive()) {
-
   path <- normalizePath(path, mustWork = FALSE)
 
   name <- basename(path)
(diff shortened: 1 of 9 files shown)


R/utils.R
```diff
@@ -3,14 +3,21 @@ can_overwrite <- function(path) {
 
   if (!file.exists(path)) {
     TRUE
-  } else if (interactive() && !yesno("Overwrite `", name, "`?")) {
```
hadley: I think this could be simplified if:

```R
if (!file.exists(path)) {
  TRUE 
} else {
  nope("Overwrite `", name, "`?")
}
```

But maybe I'm trying to be too clever? Either way, the logic here isn't that clear to me.

R/create.R
```diff
@@ -144,7 +144,8 @@ check_not_nested <- function(path, name) {
     stop(message, call. = FALSE)
   }
 
-  if (yesno(message, " This is rarely a good idea. Do you wish to create anyway?")) {
+  if (
```
hadley: The indenting here is now a bit confusing

R/utils.R
```diff
@@ -3,14 +3,21 @@ can_overwrite <- function(path) {
...
     FALSE
   }
 }
 
-yesno <- function(...) {
+## TRUE: user selected one of the "no's" or selected nothing or not interactive
+##   or in tests
+## FALSE: user selected one of the "yes's"
+nope <- function(...) {
+  if (!interactive() || is_testing()) {
```
hadley: Should it still display the message somehow? Should non-interactive usage be an error? 

The three uses cases above each handle non-interactive sessions in a different way. Maybe `nope()` should just do `stop("`nope()` called in non-interactive session")` to force developer to deal with it?

jennybc:
I will take all the feedback and come back. Sounds like it might be good to fully separate "is the user there?" from the menu presented by `nope()`.

jennybc:
Check again @hadley? This seems pretty promising. All tests still pass, current ones and the new ones in my branch.

I still need to go through and adjust usage and few things you pointed out above. But I think our `interactive()` and the revamped `nope()` represent progress.

R/utils.R
jennybc: Your proposal works if you're happy to hit a `stop()` in non-interactive session, as opposed to returning `FALSE`. It causes a few test failures, but just because we're failing for a new reason (new message). It doesn't really change outcomes.

R/utils.R
jennybc: I've made non-interactive usage an error. I haven't rationalized the functions wrt interactivity yet but will.

PR #180, opened 2017-12-30
Add is_package() and check_is_package()

jennybc:
I decided to delegate the definition of a package to rprojroot.

I think every function I applied `check_is_package()` to is a no-brainer. We will soon find out if it's annoying that any of these refuse to work on non-package projects.

There are probably more functions that should call `check_is_package()`, but this still represents progress.

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -9,7 +9,6 @@
 #'   from `getOption("devtools.desc")`.
 #' @export
 use_description <- function(fields = NULL) {
-
   name <- project_name()
   check_package_name(name)
 
--- a/R/package.R
+++ b/R/package.R
@@ -16,6 +16,7 @@
 #' use_package("dplyr", "suggests")
 #' }
 use_package <- function(package, type = "Imports") {
+  check_is_package("use_package()")
   use_dependency(package, type)
 
   switch(tolower(type),
--- a/R/package.R
+++ b/R/package.R
@@ -16,7 +16,6 @@
 #' use_package("dplyr", "suggests")
 #' }
 use_package <- function(package, type = "Imports") {
-  check_is_package("use_package()")
   use_dependency(package, type)
 
   switch(tolower(type),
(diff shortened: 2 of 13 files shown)


R/description.R
```diff
@@ -9,6 +9,9 @@
 #'   from `getOption("devtools.desc")`.
 #' @export
 use_description <- function(fields = NULL) {
+  if (file.exists(file.path(proj_get(), "DESCRIPTION"))) {
+    check_is_package("use_description()")
+  }
```
jennybc: Necessary because this function is used to create DESCRIPTION in the first place.

R/package.R
```diff
@@ -16,6 +16,7 @@
 #' use_package("dplyr", "suggests")
 #' }
 use_package <- function(package, type = "Imports") {
+  check_is_package("use_package()")
```
hadley: Projects should be able to `use_package()` too, I think

R/proj.R
```diff
@@ -17,6 +17,31 @@ proj_find <- function(path = ".") {
 
 is_proj <- function(path = ".") !is.null(proj_find(path))
 
+is_package <- function(base_path = proj_get()) {
+  res <- tryCatch(
+    rprojroot::find_package_root_file(path = base_path),
+    error = function(e) NA
```
hadley: I think it's slightly better to use `NULL` in situations like this, because `NA` is a vector.

R/proj.R
```diff
@@ -17,6 +17,31 @@ proj_find <- function(path = ".") {
...
+  )
+  !is.na(res)
+}
+
+check_is_package <- function(whos_asking = NULL) {
+  if (!is_package()) {
```
hadley: I think it's cleaner to do an early return here:

```R
if (is_package()) {
  return(invisible())
}
```

jennybc:
Made all those changes and added tests.

PR #210, opened 2018-01-16
Link lifecycle badges to definitions

### LINKED ISSUE #193: Linking lifecycle badges to stage definitions
The lifecycle badges are great but unlike the repostatus badges they link to standalone svg images rather than stage definitions. I think linking the lifecycle badges to their definitions would be a nice improvement.

jennybc:
Fixes #193

### DIFF
--- a/R/badge.R
+++ b/R/badge.R
@@ -38,33 +38,34 @@ use_depsy_badge <- function() {
 
 #' Create a life cycle badge
 #'
-#' * Experimental: very early days, a lot of churn in search of a good API.
-#'   Not on CRAN. Might not go anywhere. Use with care.
-#' * Maturing: API roughed out, but finer details likely to change. Will strive
-#'   to maintain backward compatibility, but need wider usage in order to get
-#'   more feedback.
-#' * Dormant: not currently working on it, but plan to come back to in the
-#'   future.
-#' * Stable: we're happy with the API, and the unlikely to be major changes.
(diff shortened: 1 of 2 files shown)


jennybc:
Usage

``` r
> use_lifecycle_badge("reTired")
● Add a lifecycle badge by adding the following line to your README:
Copying code to clipboard:
  [![lifecycle](https://img.shields.io/badge/lifecycle-retired-orange.svg)](https://www.tidyverse.org/lifecycle/#retired)
```

PROBLEM: If you render a `README.[Rr]?md` with such a badge, the badge svg is hyperlinked to <https://www.tidyverse.org>, i.e. the `lifecycle/#STAGE` is dropped in the `.html`.

Why?

jennybc:
cc @batpigandme

hadley:
Quite possibly a pkgdown bug

batpigandme:
I'm happy to look into this more closely— is it possible it's something in the tweak anchors region?

https://github.com/r-lib/pkgdown/blob/c94e6b2d6ecc69c1a2e4093cd5736aa97dac94f9/R/html.R#L1

hadley:
It'll probably be easiest for me to look into, since I don't currently remember the code well enough to even point you in the right direction.

jennybc:
I can't reproduce this problem now, so it must have been user error 🎉

PR #213, opened 2018-01-16
Don't try to browse local folder on RStudio server

### LINKED ISSUE #202: browseURL() on RStudio Server
The `browseURL()` usage in `use_course()` doesn't have the desired effect of opening the newly created directory on RStudio server. It attempts to open something like "8787/files/rematch2-master".

Reported by @mdsumner, who might care to elaborate here.

This is related to, but not the same as #129, in which `browseURL()` is being used to open an RStudio Project in, say, `create_package()`.

jennybc:
Fixes #202 browseURL() on RStudio Server

### DIFF
--- a/R/rstudio.R
+++ b/R/rstudio.R
@@ -104,6 +104,13 @@ in_rstudio <- function(base_path = proj_get()) {
   normalizePath(proj) == normalizePath(base_path)
 }
 
+in_rstudio_server <- function() {
+  if (!rstudioapi::isAvailable()) {
+    return(FALSE)
+  }
+  identical(rstudioapi::versionInfo()$mode, "server")
+}
+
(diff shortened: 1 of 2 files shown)


R/rstudio.R
```diff
@@ -104,6 +104,13 @@ in_rstudio <- function(base_path = proj_get()) {
   normalizePath(proj) == normalizePath(base_path)
 }
 
+in_rstudio_server <- function() {
+  if (!rstudioapi::isAvailable()) {
+    return(FALSE)
+  }
+  isTRUE(rstudioapi::versionInfo()$mode == "server")
```
hadley: I'd write `identical(rstudioapi::versionInfo()$mode, "server")`

Issue #215, opened 2018-01-17
create_from_github() should share many arguments with use_github()


jennybc:
`create_from github()` is capable of creating a GitHub repository, so it should be possible to specify `private` and `host`. `create_from_github()` should also be setting one or more remotes, therefore it should have a `protocol` argument. Finally auth is definitely required, so `credentials` and `auth_token` are also necessary.

Re-framed version of #114. Should be solved in the context of #214 GitHub connectivity.

cboettig:
Minor issue: in `create_from_github()`: if `path` doesn't exist, I get an error that it doesn't exist, while the docs imply the path will be created.  Also the behavior is a bit unexpected, in that:

```
git clone git@remote_address/repo_name path/to/local_name
```

Creates the new repo in the directory `to` with name `local_name`, while `create_from_github` does not allow the local directory to have a different name (this is related to a bigger conceptual issue I'll follow up in #214, essentially that I want to start by copying a repo called `my_template` from which I want to create repos called `project_1`, `project_2`, ..., which afik the current structure doesn't permit).

jennybc:
> Minor issue: in `create_from_github()`: if `path` doesn't exist, I get an error that it doesn't exist, while the docs imply the path will be created. 

Current version (now merged btw) has no `path` argument, only `destdir`. But perhaps everything you say applies to that checking now. I just want to make sure you're using the most recent version because this feedback is super timely /time-senstive.

jennybc:
@cboettig You're correct about the naming constraints. It is an intentional compromise/simplification for the time being. But if you feel really strongly, maybe we'll do something different before submission (imminent).

Basically, I want auto-naming to work well. I care more about that being able to override the name, because I perceive that it comes up more often. And with the current interface, you get to pick one. Long-term, I totally agree we should find a way to have both and a nice UI.

cboettig:
@jennybc Thanks! sorry I was testing `master` prior to merge.  saw the merge, just re-installed `master` branch and testing that now.

jennybc:
Address by this PR: #253 

@ijlyttle @cboettig I'd love it if you'd try this PR out.

@ijlyttle Does it work on GHE for you?

@cboettig I know this doesn't tick all your boxes -- it's aimed at typical single user use, not course workflow -- but hopefully still of interest.

cboettig:
:+1:

ijlyttle:
[...]

Issue #214, opened 2018-01-17
GitHub connectivity


jennybc:
Organizing various thoughts and issues around GitHub. Won't all be completed in v1.2.0 but this should make it easier to discuss and plan.

You = the usethis user
Them = anyone else, but often an instructor, collaborator, other dev
Here's the relevant universe of Git repos (them vs. you * GitHub vs. local):

<img width="638" alt="them-you-github-local" src="https://user-images.githubusercontent.com/599454/35029450-cf98e0e0-fb0f-11e7-87e1-ec9daed91a09.png">

How usethis functions help to make the necessary connections:

<img width="267" alt="screen shot 2018-01-16 at 10 53 34 pm" src="https://user-images.githubusercontent.com/599454/35029515-23402f0a-fb10-11e7-8cf1-341d79eb95e6.png">

`create_from_github(repo = THEM_GITHUB, destdir = YOU_LOCAL_PARENT_DIR, fork = TRUE, ...)`
TODO: add THEM_GITHUB as remote `fork-origin` of YOU_LOCAL repo

<img width="243" alt="screen shot 2018-01-16 at 10 56 09 pm" src="https://user-images.githubusercontent.com/599454/35029584-7cc997c8-fb10-11e7-9b06-9efef1694e8b.png">

`create_from_github(repo = THEM_GITHUB, destdir = YOU_LOCAL_PARENT_DIR, fork = FALSE, ...)`

<img width="117" alt="screen shot 2018-01-16 at 10 56 45 pm" src="https://user-images.githubusercontent.com/599454/35029603-926a9244-fb10-11e7-9462-2a0408943be7.png">

`create_from_github(repo = YOU_GITHUB, destdir = YOU_LOCAL_PARENT_DIR, fork = FALSE, ...)`

<img width="83" alt="screen shot 2018-01-16 at 10 57 32 pm" src="https://user-images.githubusercontent.com/599454/35029631-ab7c9034-fb10-11e7-9b9d-99389a833b8e.png">

`use_github(...)`

jennybc:
> Essentially, I want to go from box 1 to box 1 (or equivalently, box 2 to box 2).

Is there even an existing way to do that in the browser or via the API? Create GitHub repo A from GitHub repo B without forking?

I do understand your goal btw.

cboettig:
No, there's no way to do this in the GitHub API.  My understanding is that was a major reason why GitHub Classroom was created in the first place, it provides a web API to do exactly this (and little else). 

[...]

PR #216, opened 2018-01-17
Progress on create_from_github()

### LINKED ISSUE #114: create_from_github should not assume authentication.
Minor issue from #109

Currently, `create_from_github()` will throw an opaque error if the user has not put a valid `GITHUB_PAT` environmental variable, since without authentication the API will not return a `permissions` chunk.  If any behavior doesn't need auth, then it should work without this, otherwise, it should be documented that you should set `GITHUB_PAT` as an env var, and the function should probably have an argument to specify the token directly, consistent with `use_github()` etc.

jennybc:
Fixes #114. For now. In some sense.

Basically this function currently only supports cloning from a public repo. If a PAT is detected, then forking is possible. To make it do more .... will require more time than we have for this release. It needs to have a lot of the same functionality as `use_github()`, but going in the opposite direction (#215 `create_from_github()` should share many arguments with `use_github()`). I've also created a more general issue for GitHub connectivity #214.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -77,45 +77,94 @@ create_project <- function(path,
   invisible(TRUE)
 }
 
-#' Create a project from a github repository
+#' Create a repo and project from GitHub
+#'
+#' Creates a new local Git repository from a repository on GitHub. If you have
+#' pre-configured a GitHub personal access token (PAT) as described in
+#' [gh::gh_whoami()], you will get more sensible default behavior for the `fork`
+#' argument. You cannot create a fork without a PAT. Currently only works for
+#' public repositories. A future version of this function will likely have an
+#' interface closer to [use_github()], i.e. more ability to accept credentials
+#' and more control over the Git configuration of the affected remote or local
+#' repositories.
+#'
+#' @seealso [use_course()] for one-time download of all files in a Git repo,
+#'   without any local or remote Git operations.
 #'
 #' @inheritParams create_package
-#' @param repo Full name of repository: `owner/repo`
-#' @param fork Create a fork before cloning? Defaults to `TRUE` if you can't
-#'   push to `repo`, `FALSE` if you can.
+#' @param repo GitHub repo specification in this form: `owner/repo`. The second
+#'   part, i.e. the GitHub repo name, will be the name of the new local repo.
+#' @inheritParams use_course
+#' @param fork Create and clone a fork? Or clone `repo` itself? Defaults to
+#'   `TRUE` if you can't push to `repo`, `FALSE` if you can.
+#' @param rstudio Initiate an [RStudio
+#'   Project](https://support.rstudio.com/hc/en-us/articles/200526207-Using-Projects)?
+#'    Defaults to `TRUE` if in an RStudio session and project has no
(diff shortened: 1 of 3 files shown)


R/create.R
```diff
@@ -77,45 +77,94 @@ create_project <- function(path,
...
+#'    Defaults to `TRUE` if in an RStudio session and project has no
+#'   pre-existing `.Rproj` file. Defaults to `FALSE` otherwise.
 #' @export
-create_from_github <- function(repo, path, fork = NA, open = TRUE) {
+#' @examples
+#' \dontrun{
+#' create_from_github("r-lib/usethis")
+#' }
+create_from_github <- function(repo,
+                               destdir = NULL,
```
jennybc: I'm now using `destdir` like `use_course()`. The name of the new directory = repo is determined by `repo` of `owner/repo`. So the user is really specifying a destination or parent directory. It has same default behaviour, which is to put the new repo in `~/Desktop`.

R/create.R
```diff
@@ -77,45 +77,94 @@ create_project <- function(path,
...
+                               fork = NA,
+                               rstudio = NULL,
```
jennybc: Gained `rstudio` argument to match `create_package()` and `create_project()`.

R/create.R
```diff
@@ -77,45 +77,94 @@ create_project <- function(path,
...
   if (is.na(fork)) {
-    fork <- !repo_info$permissions$push
+    perms <- repo_info$permissions
+    if (is.null(perms)) {
+      # if permissions are absent, there's no PAT and we can't fork
+      fork <- FALSE
+    } else {
+      # fork only if can't push to the repo
+      fork <- !isTRUE(perms$push)
+    }
```
jennybc: This is the main fix: permissions won't be present if we can't pick up a PAT automatically.

R/proj.R
```diff
@@ -4,6 +4,7 @@ proj_crit <- function() {
   rprojroot::has_file(".here") |
     rprojroot::is_rstudio_project |
     rprojroot::is_r_package |
+    rprojroot::is_git_root |
```
jennybc: I guess this was an oversight? I hadn't noticed till now but the lack of this caused some really puzzling results in my manual tests, until I figure it out. I've always been assuming that git repos would be seen as a project in the `proj_get()/proj_set()` sense here.

jennybc:
@cboettig If you can take a look, I'd also love your confirmation that this function is less frustrating now. As you saw in other issues, it needs more work but that will have to come later.

R/create.R
hadley: 👍

R/proj.R
hadley: I think this was deliberate, but I forgot why 😬

jennybc: Maybe it had to do with some form of deliberate nesting? I added it because we need to be able to recognize a newly cloned repo as a project, even if it's not a package. Alternatively, I could set a `.here` sentinel in that case 🤔

Decision: I leave `rprojroot::is_git_root` in here. The universe will remind us if/why that is a bad idea.

PR #223, opened 2018-01-21
use_version function

### LINKED ISSUE #188: Idea: function to increment version number
Would it be within the scope of this package to have a function similar to `usethis::use_dev_version()` but instead it could increment the ‘major’, ‘minor’ or ‘patch’ accordingly.

If it is of interest I would be happy to conjure something up.

EmilHvitfeldt:
Fixes #188 

I would like to resolve #188 and I think I have a solution ready here. 

It defaults to an interactive menu where the current version is shown and you are asked which level and how much you would like to increment with. Furthermore does it have arguments for faster non-interactive use. :)


jennybc:
@hadley Will you indicate your opinion at a high level? If you basically support the goal, I can shepherd the PR further along.

hadley:
Seems like a little too much interactivity right now - I think if you want to increment in some non-standard way, you'll need to use the arguments. Also need to adjust `use_dev_version()` to use the same underlying code.

Overall seems like a good idea, although the implementation needs some work.

jennybc:
@EmilHvitfeldt Let me know if you want more concrete advice before proceeding. Otherwise, just ping me when things have advanced.

I came across this function today, which you should probably look into: `desc::desc_bump_version()`.

EmilHvitfeldt:
Thanks @jennybc! Yes I would like some more concrete advice.  

So right now I feel the direction is:
- A function `use_version ` with 1 layer of interactivity with Mayor, Minor, Patch or Development. With arguments available for nonstandard increment and interactivity-avoidance.
- Adjustment to `use_dev_version()`, possible wrapper around `use_version ` with level = Development.

I will look at `desc::desc_bump_version()` Again. I remember looking at it earlier and deciding against using it.

What convention are using with regard to bumping of .9000. What increment levels should remove .9000 or should it be a separate action?

jennybc:
A few thoughts:

  * I would hard-wire an increment of 1. That eliminates one interactive query. We just want to cover the most common development tasks.
  * Consider naming the argument `bump`, i.e. `use_version(bump = "minor")`.
  * Valid values for `"bump"` would be `c("dev", "patch", "minor", "major")`.
  * I'd present choices approximately like so

```
Current version is 1.2.0.9000
What part to increment?

1: Major --> 2.0.0
2: Minor --> 1.3.0
3: Patch --> 1.2.1
4: Dev --> 1.2.0.9001
```

or

```
Current version is 1.0.0
What part to increment?

1: Major --> 2.0.0
2: Minor --> 1.1.0
3: Patch --> 1.0.1
4: Dev --> 1.0.0.9000
```

> What convention are using with regard to bumping of .9000. What increment levels should remove .9000 or should it be a separate action?

It seems like, if you have a dev version number, bumping the patch version should take you out of a dev version number.

Does this help? At this point, you've probably thought about this more than I have.

EmilHvitfeldt:
Okay I think this is it. I have

- hardcoded the increment to 1.
- removed the second layer of interactivity.
- presented the resulting version numbers in the first layer of interactivity on the following format

> Current version is 2.0.0
> what part to increment? 
> 
> 1: major --> 3.0.0
> 2: minor --> 2.1.0
> 3: patch --> 2.0.1
> 4: dev --> 2.0.0.9000

with lowercase words to match the arguments to lessen confusing.
- changed the argument name to "which".

I tried using the `desc_bump_version()` function but I could not avoid printing method while keeping the version update. To alleviate that I included a function `bump_version` in the helpers.R that did the same underlying task.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: usethis
 Title: Automate Package and Project Setup
-Version: 1.2.0.9000
+Version: 1.1.1.9000
```
jennybc: Please remove this change.

jennybc:
> I tried using the desc_bump_version() function but I could not avoid printing method while keeping the version update

I don't quite understand this and would like to. Do you mean it's impossible to just get a version string back from the desc package vs. actually changing the version? I think I'm beginning to see the problem....

R/helpers.R
```diff
@@ -237,3 +237,28 @@ view_url <- function(..., open = interactive()) {
   }
   invisible(url)
 }
+
+bump_version <- function(ver, choice) {
```
jennybc: I'd stick with same vocabulary and input values. So "which" instead of "choice" and `c("major", "minor", "patch", "dev")` as possible values.

tests/testthat/test-use-version.R
```diff
@@ -0,0 +1,97 @@
+context("use_version.R")
+
+test_that("use_version() increments Mayor by 1 and resets other fields", {
```
jennybc: "major"

tests/testthat/test-use-version.R
```diff
@@ -0,0 +1,97 @@
...
+    use_description_field(name = "Version", value = "1.1.1.9000", overwrite = TRUE)
+  )
+  capture_output(use_version("patch"))
+  expect_identical(
+    as.character(desc::desc_get_version(proj_get())),
+    "1.1.2"
+  )
+})
+
+test_that("use_version() increments patch by 1 leaving other fields alone", {
```
jennybc: Is this a duplicate of the test above?

jennybc:
I'd shift the all the case-by-case testing to be testing the `bump_version()` helper. Then it can be much more succinct, i.e. multiple expectations inside one test, and won't need a skip. Then you really only need one test of `use_version()` to make sure it's wired up correctly.

R/version.R
```diff
@@ -0,0 +1,48 @@
...
+  types <- c("major", "minor", "patch", "dev")
+
+  ver <- desc::desc_get_version(proj_get())
+
+  vers <- c(bump_version(ver, 1),
+            bump_version(ver, 2),
+            bump_version(ver, 3),
+            bump_version(ver, 4))
+
+  if(is.null(which)) {
```
jennybc: needs to be `if (is.null(which) && interactive()) { ... }`

EmilHvitfeldt:
> I don't quite understand this and would like to. Do you mean it's impossible to just get a version string back from the desc package vs. actually changing the version? I think I'm beginning to see the problem....

Yes the desc package will calculate the updated version string and update the DESCRIPTION at the same time.

jennybc:
Thanks to @gaborcsardi:

There is a way to use desc's version incrementation logic w/o changing the version of the active package:

https://github.com/r-lib/desc/issues/61#issuecomment-366791937

Are you game to put this inside `bump_version()`? I would like to NOT maintain a separate version of this logic but, instead, get it from desc.

EmilHvitfeldt:
That is amazing! I put it inside `bump_version()`.

jennybc:
Thanks! Will you add a bullet to the dev version section of NEWS? I might groom this a bit after the merge but will cc you on those commits, so you're in the loop.

EmilHvitfeldt:
I'm a little unsure of how I would change the NEWS.md file since my forked directory is behind and doesn't have the up to date NEWS.md.

jennybc:
> I'm a little unsure of how I would change the NEWS.md file since my forked directory is behind and doesn't have the up to date NEWS.md.

OK I'll take it from here then.

Next time, you should create a branch in your fork, then do your pull request work in that branch. If master changes during the process, you would pull those changes into master in your fork, then merge master into your branch.

But this PR is fine. We are developing a guide for this, but it's not in position yet.

Thanks!

PR #229, opened 2018-01-26
Make the default Title title case

jimhester:
Currently the default field is not title case, so if you check a newly created package with `--as-cran` you get a NOTE.

    The Title field should be in title case, current version then in title case:
    ‘What the Package Does (one line, title case)’
    ‘What the Package Does (One Line, Title Case)’

Maybe the note is helpful in that it reminds you to change the title, but maybe not and it would be nicer to avoid the note.


jennybc:
Good point.

PR #237, opened 2018-02-08
Update pipe.R

CorradoLanera:
devtools::install() return the following warning:
```
missing file link '%>%'
```
which makes Travis crash at:
```
checking whether package ‘XYZ’ can be installed ... ERROR
```

use of rdname works correctly

### DIFF
--- a/inst/templates/pipe.R
+++ b/inst/templates/pipe.R
@@ -1,6 +1,6 @@
 #' Pipe operator
 #'
-#' See \code{magrittr::\link[magrittr]{\%>\%}} for details.
+#' See \code{magrittr::\link[magrittr]{pipe}} for details.
 #'
 #' @name %>%
 #' @rdname pipe


jennybc:
I can't reproduce this problem. Are you sure your devtools is current? Can you be more explicit about what you're doing? Can you point me to a travis log file with the error you're getting?

Both forms of link work for me, so I suppose either is fine.

@hadley A link to the `pipe` topic does seems more robust than to `%>%`, but perhaps less evocative. Do you care?

CorradoLanera:
[This](https://travis-ci.com/CorradoLanera/CrossClustering/builds/65289048) Trevis job on [this](https://github.com/CorradoLanera/CrossClustering/commit/17f8cb04ca37826eaca216efc7b6e9b6d9a0ddb2) commit returned in the Trevis error:
```
checking whether package ‘CrossClustering’ can be installed ... ERROR
Installation failed.
```
On the other hand, for the same commit (on my pc: win10 + RStudio 1.1.419 + R 3.4.3 + devtools 1.13.4) `devtools::check()` results in 0 errors, 0 warnings, 0 notes and I can "install & reload" the package without problems. I can't figure out why on my pc I had no problem but Travis can't "install" the package.
After some googling I have read [here](http://r-pkgs.had.co.nz/check.html) the suggestion to explicitly run `devtools::install()` in case of that error. 

Running `devtools::install()` resulted in a single warning:
```
Rd warning: C:/Users/corra/OneDrive/CrossClustering/man/pipe.Rd:10: missing file link '%>%'
```
which wasn't throw with `pipe` instead of `\%>\%`.

[here](https://travis-ci.com/CorradoLanera/CrossClustering/builds) there is all the Travis job history.

I'm very new to Travis and I had some trouble to understand I have to merge something to make it run jobs again on a new commit so before I was able to re-run Travis I made some changes and more than one commit (you can track them [here](https://github.com/CorradoLanera/CrossClustering/commits/develop)). Now Travis works properly.

jennybc:
I used `use_pipe()` here in this package [in a branch](https://github.com/r-lib/usethis/tree/pipe-doc-test) as an experiment and have no trouble with R CMD check, including [on Travis](https://travis-ci.org/r-lib/usethis/builds/338853380?utm_source=github_status&utm_medium=notification). I think there must be something else going on in your situation that is the root cause of your difficulty. Maybe looking at that branch and commit will reveal what is different from your situation.

CorradoLanera:
I set `use_pipe()` again in a different branch as you suggested me and no trouble with Travis nor AppVeyor. So, I don't understand what was the problem but it seems not related with `use_pipe()`. I agree with you that `%>%` is more evocative and possibly better than `pipe`, but `devtools::install()` continue to throw the same warning:

```r
> devtools::install()
Installing CrossClustering
"C:/PROGRA~1/R/R-3.4.3/bin/x64/R" --no-site-file --no-environ --no-save --no-restore --quiet CMD INSTALL  \
  "C:/Users/corra/OneDrive/CrossClustering" --library="C:/Users/corra/Documents/R/win-library/3.4" --install-tests 

* installing *source* package 'CrossClustering' ...
** R
** data
** tests
** preparing package for lazy loading
** help
*** installing help indices
  converting help for package 'CrossClustering'
    finding HTML links ... done
    ARI_contingency                         html  
    CrossClustering                         html  
    PermSignificanceARI                     html  
    SignificanceARI                         html  
    chain_effect                            html  
    is_zero                                 html  
    max_proportion_function                 html  
    pipe                                    html  
Rd warning: C:/Users/corra/OneDrive/CrossClustering/man/pipe.Rd:10: missing file link '%>%'
    prune_zero_tail                         html  
    toy                                     html  
    twomoons                                html  
    which_cluster                           html  
    worms                                   html  
** building package indices
** testing if installed package can be loaded
*** arch - i386
*** arch - x64
* DONE (CrossClustering)
In R CMD INSTALL
Reloading installed CrossClustering
```
I don't know if it can cause some throuble.

PR #241, opened 2018-02-12
Tolerate .Rproj files in the session temp dir

jennybc:
I suffer from this frequently when other activities have left a stray `.Rproj` file in the session temp directory. ~~Is this kosher? No one should be putting anything of value there, right?~~

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -197,6 +197,14 @@ check_not_nested <- function(path, name) {
     return()
   }
 
+  ## special case: allow nested project if
+  ## 1) is_testing()
+  ## 2) proposed project name matches magic string we build into test projects
+  ## https://github.com/r-lib/usethis/pull/241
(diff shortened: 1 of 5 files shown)


hadley:
Hmmm, this is a little scary — I think I'd rather fix it with a special case in the function that checks if you're already in a project

jennybc:
I suppose I could exploit `is_testing()` here.

So, by special case, do you mean proceed with the creation of a nested project, without explicit user permission, if it appears we are running tests?

hadley:
I was thinking look for `/tmp` in the path, and skip checks if so. `is_testing()` would be another option.

jennybc:
I implemented the special case approach. I skip the "nested project ... are you sure?" check if we are testing AND a special string is built into the project name.

I think it would be awfully hard to do that by accident and I'm not deleting any files now.

Issue #239, opened 2018-02-11
support for binder?


uribo:
[Binder](https://mybinder.org/) (beta) provides a reproducible execution environment based on the GitHub repository. Also, supports badges  and generates a link to launch the repository in the following format:

```
[![Binder](https://mybinder.org/badge.svg)](https://mybinder.org/v2/gh/<user>/<repository>/master)
```

[![Binder](http://mybinder.org/badge.svg)](http://mybinder.org/v2/gh/binder-examples/dockerfile-rstudio/master)

I think that badge display is useful for a everyone in a single report or article. 
So I would like to add a `use_binder_badge()` function like `usethis::use_lifecycle_badge()`.

If you are positive about this feature, can I submit a draft proposal as pull request? 
What I'm thinking is adding a *Dockerfile* that can launch an rstudio session (use the official [imagefile](https://github.com/binder-examples/dockerfile-rstudio/blob/master/Dockerfile)), if there is no *Dockerfile* or *runtime.txt*, along with the code output for the badge.

batpigandme:
Point of clarification: are you proposing the integration of support for the binder _badge_ or binder on the whole? (No judgement either way, just trying to make sure I understand correctly 👍).

uribo:
Thank you for your reply. I will prepare for the pull request soon.

jennybc:
Thanks @uribo. Can you answer the question @batpigandme posed, before proceeding? Getting a PR for a badge is pretty different than for a Dockerfile. I'm not saying the latter is necessarily unwelcome, but it's a bigger responsibility re: maintenance.

uribo:
Sorry @jennybc @batpigandme 

What I mentioned was a request for a function to display a binder badge.
The latter topic, I would like to discuss when usethis package will integrate with docker in the future.

PR #242, opened 2018-02-12
Support a binder badge

uribo:
My Proposal is here #239 

Added new function `use_binder_badge()` that prints out [Binder](https://mybinder.org/) badge.


R/badge.R
```diff
@@ -78,6 +78,32 @@ stages <- c(
   questioning = "blue"
 )
 
+#' Create a Binder badge
+#'
+#' This prints out the markdown which will display a Binder "badge", indicating
+#' launch build your repository source codes on Binder <https://mybinder.org/>.
```
jennybc: This part of the sentence still doesn't make sense to me:

> indicating launch build your repository source codes on Binder

I visited <https://mybinder.org> and tried to understand what's going on. Does this sound correct?

> indicating that your repository can be launched in an executable environment on <https://mybinder.org/>.

If that's OK, let's use that wording (or something similar).

R/badge.R
```diff
@@ -78,6 +78,32 @@ stages <- c(
...
+#'
+#' @export
+use_binder_badge <- function() {
+
+  if (uses_git(path = proj_get()) == TRUE) {
```
jennybc: You don't need the `== TRUE`. `uses_git()` already returns either `TRUE` or `FALSE`.

Also, I wonder if this should be `uses_github()` in this case?

jennybc:
Will you also add a bullet to NEWS.md?

R/badge.R
uribo: Thank you for your suggestion. That is the tidy phrase :)

R/badge.R
uribo: Hmm. Indeed, I might as well change to `uses_github()` as you say.

[Here](https://github.com/r-lib/usethis/blob/d12b3f14c24768c65aa2abf147f1a2951c500a14/R/github.R#L72), `uses_github()` is used the same way.

uribo:
Done. But I'm sorry, lowered coverage.

jennybc:
Thanks!

PR #247, opened 2018-02-13
Adds single function for creating tidy GitHub templates

batpigandme:
* Adds `use_tidy_github()` as a stand-alone function for tidyverse templates in the `.github/` directory.
* Adds test for same (following the model/copying and pasting from Jenny's individual tidy function tests).
* Does not: 
  - Delete the other functions yet, since I wanted to make sure this was ~the right direction first…
  - Add a _new_ NEWS bullet, since it's not user-facing.


NEWS.md
```diff
@@ -6,7 +6,7 @@
 
 * `create_package()` and `create_project()` return a normalized path, even if target directory does not pre-exist (#227, #228).
 
-* "Tidy" GitHub templates are placed in a `.github/` subdirectory (#165 @batpigandme)
+* "Tidy" GitHub templates are placed in a `.github/` subdirectory (#165 @batpigandme).
```
jennybc: I know I wrote this sentence. But perhaps it should say "tidyverse GitHub templates". The templates themselves are neither tidy nor untidy.

R/tidyverse.R
```diff
@@ -32,6 +32,10 @@
 #' * `use_tidy_community()`: wrapper function around contributing, issues,
 #'   support, and code of conduct functions.
 #'
+#' * `use_tidy_github()`: creates standard tidyverse: issue template,
+#'   support resources document, contributing guidelines, and code of conduct in
+#'   an `.github/` subdirectory.
+#'
```
jennybc: creates standard tidyverse ?missing word?: issue template, ...

R/tidyverse.R
```diff
@@ -203,3 +207,45 @@ use_tidy_community <- function() {
   use_tidy_support()
   use_tidy_coc()
 }
+
+
+#' @export
+#' @rdname tidyverse
+use_tidy_github <- function() {
+  check_uses_github()
+  check_uses_travis()
```
jennybc: Are we going to be sad that travis must be setup before one can use this?

jennybc:
> Does not:
> Delete the other functions yet, since I wanted to make sure this was ~the right direction first…

I defer to @hadley re: what he wants.

> Does not:
> Add a new NEWS bullet, since it's not user-facing.

Before this gets merged, I think the addition of this function should appear in NEWS. And, if it happens, the removal of the individual functions.

R/tidyverse.R
batpigandme:
```
creates standard tidyverse (then the colon was for the beginning of a list) issue template, support resources document…
```
I could easily just kill that colon though. It _does_ look super confusing in monospace.

I swapped "a" and "an" like three or four times (including changing the order to `a subdirectory, .github/` to avoid the issue entirely)…I'll put it back to "a."

batpigandme:
Made changes for _most_ of @jennybc's comments.

Proposal: Remove the current `use_tidy_community()` and rename this new function (`use_tidy_github()`) `use_tidy_community()`. They, in effect, do the same thing, and it occured to me that `use_tidy_github()` ~sounds like it involves more github setup than it actually does.

Once we decide whether to remove the individual functions and/or change the name up, I'll add a NEWS bullet reflecting same.

_Sorry if this half-baked PR thing is a pain. 😔 In retrospect it might have been easier for you if I worked this one over more in issues, and referenced code in my branch…_

PR #249, opened 2018-02-14
Work on use_github_labels()

jennybc:
* Export ~~`gh_labels()`~~ `tidy_labels()` which returns named vector of in-house standard labels and colours
  * Add `labels` argument, in case anyone wants something other than ~~`gh_labels()`~~ `tidy_labels()`
  * Fix buglet that prevented modification of exactly 1 colour + handle a length-0 edge case
  * Add "good first issue" and "help wanted", with GitHub default colours (now tweaked)
  * ~~Adopt GitHub default colours for "bug" (just a different red) and "feature" (the light blue they use for "enhancement")~~
  * Beefed up the docs
  * Styled with styler

This repo has had this version of `use_github_labels()` called on it, so here's the effect:
https://github.com/r-lib/usethis/labels


R/github-labels.R
```diff
@@ -1,22 +1,41 @@
...
-#' This function creates new labels and changes colours (if needed).
-#' It does not generally remove labels, unless you explicit ask to
-#' remove GitHub's defaults.
+#' @description `use_github_labels()` creates new labels and/or changes label
+#' colours. It does not generally remove labels, unless you explicitly ask to
+#' remove the default GitHub labels that are not present in the labels you
+#' provide via `labels`.
 #'
-#' @param delete_default If `TRUE`, will remove default labels.
+#' `gh_labels()` returns the labels and colours commonly used by tidyverse
```
hadley: Should this include `tidy` since it's a tidyverse convention?

R/github-labels.R
```diff
@@ -1,22 +1,41 @@
-#' Create standard github labels for labelling issues.
+#' Manage GitHub issue labels.
 #'
-#' This function creates new labels and changes colours (if needed).
-#' It does not generally remove labels, unless you explicit ask to
-#' remove GitHub's defaults.
+#' @description `use_github_labels()` creates new labels and/or changes label
+#' colours. It does not generally remove labels, unless you explicitly ask to
```
hadley: unless explicitly requested with `remove = TRUE`?

R/github-labels.R
```diff
@@ -1,22 +1,41 @@
...
+#' @rdname use_github_labels
 #' @export
-use_github_labels <- function(delete_default = FALSE,
+#' @examples
+#' \dontrun{
+#' ## typical use in, e.g., a new tidyverse project
+#' use_github_labels(delete_default = TRUE)
+#' }
+use_github_labels <- function(labels = gh_labels(),
+                              delete_default = FALSE,
```
hadley: Should this be `remove = FALSE`?

R/github-labels.R
```diff
@@ -25,51 +44,51 @@ use_github_labels <- function(delete_default = FALSE,
...
 
-  if (delete_default) {
-    default <- vapply(labels, "[[", "default", FUN.VALUE = logical(1))
-    def_labels <- setdiff(cur_labels[default], names(gh_labels))
+  if (delete_default && length(cur_labels) > 0) {
+    default <- vapply(cur_labels, "[[", "default", FUN.VALUE = logical(1))
+    def_labels <- setdiff(cur_label_names[default], names(labels))
 
     if (length(def_labels) > 0) {
       done("Removing default labels: ", collapse(value(def_labels)))
```
hadley: default -> other ?

R/github-labels.R
```diff
@@ -81,11 +100,17 @@ use_github_labels <- function(delete_default = FALSE,
...
-  "bug" =         "e02a2a",
-  "feature" =     "009800",
-  "reprex" =      "eb6420",
-  "wip" =         "eb6420",
-  "docs" =        "0052cc",
-  "performance" = "fbca04"
-)
+#' @rdname use_github_labels
+#' @export
+gh_labels <- function() {
```
hadley: I don't love the current colours, but I find the new colours ugly 😞

R/github-labels.R
jennybc: Are you suggesting that `remove = TRUE` requests to drop labels that are not present in `labels`, regardless of whether they are GitHub defaults?

Based on my interactions with usethis labels while working on this, that would create more work. For example, pretty sure I would have accidentally deleted the [non-package label](https://github.com/r-lib/usethis/labels/non-package) by now.

R/github-labels.R
jennybc: `gh_labels()` --> `tidy_labels()` in 7d7a544719593807769a86af06e4052ca2ff697b

R/github-labels.R
jennybc: Reworded to be mention `delete_default = TRUE`

R/github-labels.R
jennybc: Now reads "Removing labels", which should stand the test of time.

jennybc:
> Would the docs for this function be a good place to park out descriptions of when to use the different labels?

Yes, I opened #252 for that. We could merge this in the meantime.

I reverted the color changes and picked a nicer violet for "good first issue". This repo has had this version of `use_github_labels()` called on it, so here's the effect:
https://github.com/r-lib/usethis/labels

Can I merge? It's easy to tweak colours in future.

R/github-labels.R
jennybc: I reverted the changes to existing colours.

PR #254, opened 2018-02-15
Consult existing dependencies in use_dependency()

### LINKED ISSUE #99: use_dependency() should check for existing imports
E.g. when using:

```r
use_dependency("roxygen2", "Suggests")
```

on a package that already imports/depends `roxygen2` it will give a check error:

```
Package listed in more than one of Depends, Imports, Suggests, Enhances:
  ‘roxygen2’
```

jennybc:
Fixes #99 `use_dependency()` should check for existing imports

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -38,7 +38,6 @@ use_template <- function(template,
                          ignore = FALSE,
                          open = FALSE,
                          package = "usethis") {
-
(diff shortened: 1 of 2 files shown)


jennybc:
That is fair.

I went through some much cleaner versions, but unfortunately they didn't get the actual job done 🙂

Issue #177, opened 2017-12-29
Use fs when released


hadley:

jennybc:
I'm waiting for the Mac OS issue to resolve before pushing hard on this.

hadley:
I think it's ok to push ahead using fs here, and then push @jimhester to fix the fs problem 😉

Issue #198, opened 2018-01-10
Path reporting


jennybc:
In the long-run, consider messaging about paths relative to a relevant path.

For functions that operate on current project,  that's `proj_get()`.
For functions that take a path, it's ... less clear and would ideally be specified by caller.

Motivation: I'm writing `use_course()`, which has no notion of project. And the user can specify a directory where the ZIP file should be written, then unpacked. It's awkward that `can_overwrite()` forcibly removes path in this message: `"Overwrite pre-existing file ", value(basename(path)), "?"`

jennybc:
More ugliness, at least in `use_course()` and especially on Windows, comes from uneven treatment in different messages re: whether `~` has been expanded or not and replicated file path separators.

hadley:
Maybe use `fs::path_rel(path, proj_get())` everywhere? The `basename()` stuff I did was a crude hack.

hadley:
Probably not perfect, but I reviewed the `use_course()` code path and reduced some of the duplication (hopefully without losing anything important).

PR #255, opened 2018-02-16
Work on edit_*() functions

### LINKED ISSUE #181: edit_git_ignore("user") should explain how to put file into force
Just creating a global gitignore file is not enough. `edit_git_ignore("user")` should message about or complete the remaining setup:

https://stackoverflow.com/questions/7335420/global-git-ignore

### LINKED ISSUE #182: Be more flexible re: name of global gitignore file
Mine is at `~/.gitignore_global`, not `~/.gitignore`.

I though I was typical, but [it looks like I am wrong](https://twitter.com/JennyBryan/status/947225672090468352). However, some people *do* use that filename and most tutorials / blog posts seem to favour it.

Right now, if you have `~/.gitignore_global`, `edit_git_ignore("user")` doesn't pick it up and instead creates and opens a new, empty file.

Handy way to check location of global gitignore, if you have access to command line git:

``` sh
jenny@2015-mbp usethis $ git config --get core.excludesfile
/Users/jenny/.gitignore_global
```

jennybc:
Path handling will get even better once I tackle #177 (Use fs when released). Please view this as incremental progress in order to achieve this:

  * `edit_git_ignore(scope = "user")` should message if user-level `.gitignore`is created *de novo*, as `edit_r_profile()` and `edit_r_environ()` do. Until you edit `core.excludesfile` in git config, the new gitignore file has no effect. But in order to do that, you need to separate path formation from path creation & editing.
  * Fixes #181 (`edit_git_ignore("user")` should explain how to put file into force).

Fixes #182 (Be more flexible re: name of global gitignore file).

Return path, invisibly, from all `edit_*()` functions.

I think overall it is more convenient to edit a path than to edit a subpath within a basepath, given the helpers `scoped_path()`, `scoped_git_path()`, and `proj_path()`. Hence this change:

  * `edit_file(base_path, path)` --> `edit_file(full_path)`

### DIFF
--- a/R/edit.R
+++ b/R/edit.R
@@ -7,6 +7,8 @@
 #' * `edit_git_ignore()` opens `.gitignore`
 #' * `edit_rstudio_snippets(type)` opens `~/R/snippets/{type}.snippets`
 #'
+#' @return Path to the file, invisibly.
+#'
 #' @param scope Edit globally for the current __user__, or locally for the
 #'   current __project__
 #' @name edit
@@ -15,8 +17,7 @@ NULL
 #' @export
 #' @rdname edit
 edit_r_profile <- function(scope = c("user", "project")) {
-  file <- file.path(scope_dir(scope), ".Rprofile")
-  edit_file(file)
+  file <- edit_file(scoped_path(scope, ".Rprofile"))
   todo("Restart R for changes to take effect")
   invisible(file)
 }
@@ -24,8 +25,7 @@ edit_r_profile <- function(scope = c("user", "project")) {
 #' @export
 #' @rdname edit
 edit_r_environ <- function(scope = c("user", "project")) {
-  file <- file.path(scope_dir(scope), ".Renviron")
-  edit_file(file)
+  file <- edit_file(scoped_path(scope, ".Renviron"))
   todo("Restart R for changes to take effect")
   invisible(file)
 }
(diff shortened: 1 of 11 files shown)


R/helpers.R
```diff
@@ -201,30 +201,26 @@ create_directory <- function(base_path, path) {
...
+  full_path <- path.expand(path)
+  create_directory(dirname(dirname(full_path)), basename(dirname(full_path)))
 
   if (!file.exists(full_path)) {
     file.create(full_path)
   }
 
   if (!interactive() || is_testing()) {
-    todo("Edit ", value(path))
+    todo("Edit ", value(basename(path)))
```
jennybc: I will make this & the one below better when working on #177 (Use fs when released) and #198 (Path reporting).

R/edit.R
```diff
@@ -15,65 +17,100 @@ NULL
...
 #' @rdname edit
-#' @param type Snippet type. One of "R", "markdown", "C_Cpp", "Tex",
-#'   "Javascript", "HTML", "SQL"
-edit_rstudio_snippets <- function(type = "R") {
-  edit_file("~", paste0(".R/snippets/", tolower(type), ".snippets"))
-  invisible()
+edit_git_ignore <- function(scope = c("user", "project")) {
+  file <- scoped_git_path(scope, ".gitignore")
+  if (scope == "user" && !file.exists(file)) {
+    todo("Tell git about this new global gitignore file for changes to take effect!")
```
hadley: I did not know this!

jennybc: Maybe I should just call `git2r::git2r::config()` for them in this case and be done with it?

hadley: Yeah, I think that's reasonable

PR #257, opened 2018-02-16
Function to help configure a GitHub PAT

### LINKED ISSUE #248: browse_github_pat / use_github_pat
This was originally a devtools pr https://github.com/hadley/devtools/pull/1665, but I think it belongs more in usethis. Essentially a simple helper to make creating a GitHub Personal Access Token easier.

jennybc:
Fixes #248 `browse_github_pat` / `use_github_pat`

Still a WIP. Currently, pretty faithful to [the PR made by @jeroen](https://github.com/r-lib/devtools/pull/1665) on devtools.

This should finish quickly, so @hadley, happy to get any opinions you have:

  * Name: `[create|browse|use]_github_[pat|token]()`? usethis functions that deal with the PAT accept it via the `auth_token` argument (Is that what we want? gh uses `.token`.)
  * Default scopes?
  * Default nickname (a.k.a. "description")?
  * Shall I immediately message about using `edit_r_environ()`? Even call it?

I know I need to rationalize PAT docs, check if interactive, use in-house UI functions, etc.

Screenshot:

screen shot 2018-02-15 at 11 53 50 pm


hadley:
I think `browse_` would be the name most in keeping with the current behaviour.

Do we want to use [the API](https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization) to automate creation even further? Maybe it's worth pushing that off into gh, and considering later as part of an overhaul of how gh does auth?

jennybc:
> Do we want to use the API to automate creation even further? Maybe it's worth pushing that off into gh, and considering later as part of an overhaul of how gh does auth?

Yeah, I am not inclined to get into this with usethis for this release. I'd like to just make the web flow easier for an individual this time.

PR #272, opened 2018-02-19
Refactor proj_get() and proj_set()

### LINKED ISSUE #265: Behaviour of proj_set() on a non-project folder
Right now it is a silent no-op if `path` is not a project. This behaviour appears to be used internally in `proj_get()` but maybe this should be refactored. This seems weird to me:

``` r
> setwd("~")
> proj_get()
Error: Current working directory, '/Users/jenny',  does not appear to be inside a project or package.
> proj_set("~/tmp/aaa")
> proj_get()
Error: Current working directory, '/Users/jenny',  does not appear to be inside a project or package.
```

hadley:
Fixes #265

### DIFF
--- a/R/proj.R
+++ b/R/proj.R
@@ -83,7 +83,8 @@ proj_set <- function(path = ".", force = FALSE) {
   new_proj <- proj_find(path)
   if (is.null(new_proj)) {
     stop(
-      "Path ", value(path), "does not appear to be inside a project or package.",
+      "Path ", value(path),
+      " does not appear to be inside a project or package.",
       call. = FALSE
     )
   }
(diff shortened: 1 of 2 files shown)


jennybc:
@hadley With my update, I can pass *test* locally, but not *check*.

The problem is this:

https://github.com/r-lib/usethis/blob/424764e236542edda560836d77ee4a0701afa818/R/usethis-package.R#L9

We need to be able to attach usethis even if current working directory is not inside a project.

jennybc:
The error in R CMD check comes when usethis is attached prior to running examples.

I'm not sure what working directory is at that point and why it is not inside a project when I run R CMD check locally but is inside a project on travis/appveyor.

But I think it's clear you should be able to attach usethis even outside a project (?) and just get a message to that effect.

hadley:
What if we just didn't `proj_set()` on load?

PR #275, opened 2018-02-20
remove extra path separators

dchiu911:
- Removes extra slash between `github` and `gh$username`
- Removes extra slash between `gh$repo` and `?branch=master&svg=true`


jennybc:
Thanks!

PR #285, opened 2018-02-23
Add spell checker support

jeroen:
Add function to setup the automatic spell checker. This spell check runs at the end of CMD check and prints potential typos to the terminal.

It will also automatically create a WORDLIST which is a custom dictionary of approved words. No need to 'abuse' 'punctuation' for this.

Additional benefit is that it (at least currently) overrules the CRAN spell checker ;)


jennybc:
Thanks for this! As discussed, I will first concentrate on getting current release out the door.

PR #301, opened 2018-02-27
Make rmarkdown a soft dependency

jimhester:
We have previously tried to keep the stringr / stringi dependency out of devtools, however it snuck back in via rmarkdown. It looks like this was always meant to be a soft dependency (hence the calls to `check_installed()` but it was accidentally put in the wrong place in the DESCRIPTION.

I also moved styler as well, since it too can be a soft dependency.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,16 +30,16 @@ Imports:
     git2r,
     httr,
     rematch2,
-    rmarkdown,
     rprojroot,
     rstudioapi,
-    styler,
     whisker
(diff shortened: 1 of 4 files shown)


jennybc:
Do the travis/appveyor checks assure us we have all the right checks in place? Or should I do something more explicit locally re: running check w/o rmarkdown and styler in the library search path?

jennybc:
If I make rmarkdown and styler unavailable, there are some problems. I'll push more to this PR.

PR #306, opened 2018-03-05
Seperating use_github and use_github_links documentation

### LINKED ISSUE #303: use_url idea
With the use of `use_github`, does `usethis` have enough information to populate the URL section in the description? 

If so, I think that would be a nice feature, especially for packages that end up on CRAN but have development on github.

EmilHvitfeldt:
I'm hoping this would adequately resolve #303. 

I did some local testing to see if I was able to update the pkgdown page, with no success.


jennybc:
Thanks!

PR #317, opened 2018-03-27
DOCS update and fixes #128

### LINKED ISSUE #128: use_git() results in "Error in path.expand(path) : invalid 'path' argument"
```
> use_git()
✔ Initialising Git repo
✔ Adding '.Rhistory', '.RData' to './.gitignore'
✔ Adding files and committing
Error in path.expand(path) : invalid 'path' argument
```
This is with usethis v1.1.0 on OSX with RStudio 1.1.383 and R 3.4.2.  create_package did succeed, and the package directory is a subdirectory of RStudio's current working directory. 

Please let me know if I should provide additional info or this is not the best place to ask about this.

muschellij2:
Added a slight fix for issue #128 where it checks if paths are not `NULL` or technically `length(paths) > 0`, but may not be the root of the problem (interactive use only).  Also some docs update with spacing in links.


jennybc:
Thanks for this!

I decided to use an existing helper `git_uncommitted()` we have do check this (064f6383fef41feb10374015006f09cd8b829346). I still suspect #128 is more about crossed wires re: working directory vs. current project, as opposed to `use_git()` not having any changed files to commit. But it does no harm to check.

PR #321, opened 2018-04-02
Change test-pkg to test.pkg.

### LINKED ISSUE #319: examples using create_package not functionnal
In `use_github` example (and maybe other places), there is a call to `create_package("test-pkg")`. However, 
``` r
usethis::create_package("test-pkg")
#> Error: 'test-pkg' is not a valid package name. It should:
#> * Contain only ASCII letters, numbers, and '.'
#> * Have at least two characters
#> * Start with a letter
#> * Not end with '.'
```
I think the hyphen `-` does not follow the naming rule. It seems the example should be changed to a functional example. I can help with a PR if you want, trying to find all similar example. For `use_github`, the example is a _dont run_ one, it may be the reason this is still present.

batpigandme:
* Make `create_package()` example a valid package name.
* Fixes #319.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -48,9 +48,9 @@
 #' @export
 #' @examples
 #' \dontrun{
-#' create_package("test-pkg") # creates package in current working directory
+#' create_package("test.pkg") # creates package in current working directory
 #'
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks!

PR #323, opened 2018-04-04
Simplify the default context

jimhester:
This removes the test- prefix and .R extension, to match the generated
context in https://github.com/r-lib/testthat/pull/738

### DIFF
--- a/R/test.R
+++ b/R/test.R
@@ -27,13 +27,13 @@ use_testthat <- function() {
 #'   the name of the file open in the source editor.
 #' @export
 use_test <- function(name = NULL, open = TRUE) {
-  name <- find_test_name(name)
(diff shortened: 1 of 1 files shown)


jimhester:
A question for this PR is if we should remove the `context()` call from the template entirely. The upside would be an easier future transition to context-less test files. The downside would be it would require the auto-generated context from https://github.com/r-lib/testthat/pull/738, so test files created with `use_test()` would fail with CRAN testthat until the PR is merged (imminent) and the next version of testthat is released (who knows when).

jimhester:
Also this PR is in response to https://github.com/r-lib/testthat/pull/738#issuecomment-378294679

jimhester:
https://github.com/r-lib/usethis/issues/325

PR #339, opened 2018-04-26
Expose overwrite parameter in use_github_links

### LINKED ISSUE #316: use_github_links needs to expose the overwrite argument
If you already have a URL or BugReports field, and you call `use_github_links()` you get

> Error: BugReports has a different value in DESCRIPTION. Use overwrite = TRUE to overwrite.

But `use_github_links()` does not have an `overwrite` argument.

jimhester:
Previously you would get the error 

> Error: URL has a different value in DESCRIPTION. Use overwrite = TRUE to overwrite.

But there was no way to actually pass `overwrite = TRUE`.

Fixes #316

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -155,9 +155,12 @@ use_github <- function(organisation = NULL,
 }
 
 #' @export
+#' @param overwrite By default, `use_github_links()` will not overwrite
+#'   existing fields. Set to `TRUE` to overwrite existing links.
 #' @rdname use_github
 use_github_links <- function(auth_token = NULL,
-                             host = "https://api.github.com") {
+                             host = "https://api.github.com",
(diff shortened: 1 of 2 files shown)


jennybc:
Want to edit commit msg to autoclose #316?

jennybc:
I think part of the reason I didn't get into this back in the day, is that I wondered if pre-existing URL should be overwritten or appended to. Do you thoughts about that?

jimhester:
The reason I encountered this was moving the repo and wanting to update the URLs appropriately, so in that case I would want to overwrite.

I think appending is kind of special case, most repositories don't have multiple sites AFAIK.

jennybc:
Thanks!

PR #364, opened 2018-05-25
Sharpen wording around the code of conduct

### LINKED ISSUE #352: Clarify code of conduct
Make it a bit more clear that code of conduct applies to contributors to the package, not users. Maybe

> By participating in this project you agree to abide by its terms.

->

> By contributing to this project you agree to abide by its terms.

jennybc:
Fixes #352 

  * Specify the name of the project
  * "participating" --> "contributing to" in the README sentence
  * Removed `.md` from `README.md` to avoid the whole `.Rmd` vs `.md` thing

@batpigandme We are already using Contributor Covenant 1.0 and not 1.4, fwiw.

### DIFF
--- a/R/tidyverse.R
+++ b/R/tidyverse.R
@@ -201,10 +201,13 @@ use_tidy_coc <- function() {
     ".github/CODE_OF_CONDUCT.md"
   )
 
-  todo("Don't forget to describe the code of conduct in your README.md:")
+  todo("Don't forget to describe the code of conduct in your README:")
   code_block(
-    "Please note that this project is released with a [Contributor Code of Conduct](.github/CODE_OF_CONDUCT.md).",
-    "By participating in this project you agree to abide by its terms."
+    paste0("Please note that the ", value(project_name()),
+           " project is released with a ",
+           "[Contributor Code of Conduct](.github/CODE_OF_CONDUCT.md). ",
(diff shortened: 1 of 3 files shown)


R/tidyverse.R
```diff
@@ -201,10 +201,13 @@ use_tidy_coc <- function() {
     ".github/CODE_OF_CONDUCT.md"
   )
 
-  todo("Don't forget to describe the code of conduct in your README.md:")
+  todo("Don't forget to describe the code of conduct in your README:")
   code_block(
-    "Please note that this project is released with a [Contributor Code of Conduct](.github/CODE_OF_CONDUCT.md).",
-    "By participating in this project you agree to abide by its terms."
+    paste0("Please note that the ", value(project_name()),
```
hadley: Should we make this just "tidyverse packages"?

batpigandme: In that case, would r-lib packages use the traditional `use_code_of_conduct()`?

hadley: Oh good point - in that case maybe we should pull out the repeated code into a function?

(Another random thought: should `code_block()` use `glue()`?)

jennybc: > (Another random thought: should code_block() use glue()?)

YES. Will do.

PR #365, opened 2018-05-25
Don't help people depend on tidyverse or devtools

### LINKED ISSUE #172: use_package("tidyverse|devtools") should generate a warning
And suggest using the individual core packages instead

jennybc:
Fixes #172 use_package("tidyverse|devtools") should generate a warning

So far, I'm just refusing to add the dependency and not adding all the individual packages. I don't really think we want people depending on every single package in the tidyverse either. I think it's OK to ask people to figure out what they need, yes?

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -16,7 +16,6 @@
 #' }
 use_package <- function(package, type = "Imports") {
   refuse_package(package, verboten = "tidyverse")
-  refuse_package(package, verboten = "devtools")
 
   use_dependency(package, type)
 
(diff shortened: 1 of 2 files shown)


R/package.R
```diff
@@ -15,6 +15,9 @@
 #' use_package("dplyr", "suggests")
 #' }
 use_package <- function(package, type = "Imports") {
+  refuse_package(package, verboten = "tidyverse")
+  refuse_package(package, verboten = "devtools")
```
jennybc: Is it too early to do this for devtools, i.e. before more of the pieces have made it to CRAN?

hadley: Yeah, I think it's too early.

jennybc:
@batpigandme Yes, it would be good to have the body of that email as a blog post, so we have a definitive link and explanation for why there should be few / no reverse dependencies on the meta-packages. I won't use it in this PR but will use it elsewhere.

batpigandme:
Cool – draft post that is literally the email and a line introducing it is here: https://github.com/tidyverse/tidyverse.org/pull/161

PR #367, opened 2018-05-26
WIP re: options

### LINKED ISSUE #159: Review all uses of getOption("devtools.SOMETHING")
I suspect this is not implemented uniformly right now:

consult a usethis option then, if absent, consult a devtools option

### LINKED ISSUE #233: create_package() should document options
Currently in `package?devtools`

jennybc:
WIP do not merge

Closes #233 create_package() should document options, closes #159 Review all uses of getOption("devtools.SOMETHING")

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -29,6 +29,7 @@ use_description <- function(fields = NULL) {
   name <- project_name()
   check_package_name(name)
 
+  ## the definitive source of user-supplied info: in this call or via options
   fields <- fields %||%
     getOption("usethis.description") %||%
     getOption("devtools.desc") %||%
@@ -48,22 +49,15 @@ build_description <- function(name, fields = list()) {
 }
 
 build_description_list <- function(name, fields = list()) {
-  author <- getOption("devtools.desc.author") %||%
-    'person("First", "Last", , "first.last@example.com", c("aut", "cre"))'
-  license <- getOption("devtools.desc.license") %||% "What license it uses"
-  suggests <- getOption("devtools.desc.suggests")
-
   defaults <- list(
     Package = name,
     Version = "0.0.0.9000",
     Title = "What the Package Does (One Line, Title Case)",
     Description = "What the package does (one paragraph).",
-    "Authors@R" = author,
-    License = license,
-    Suggests = suggests,
+    "Authors@R" = 'person("First", "Last", , "first.last@example.com", c("aut", "cre"))',
+    License = "What license it uses",
     Encoding = "UTF-8",
-    LazyData = "true",
-    ByteCompile = "true"
+    LazyData = "true"
   )
 
   # Override defaults with user supplied options
--- a/R/create.R
+++ b/R/create.R
@@ -26,7 +26,7 @@
 #'   in RStudio, working directory is set to the new project.
 #' @export
 create_package <- function(path,
-                           fields = getOption("devtools.desc"),
+                           fields = NULL,
                            rstudio = rstudioapi::isAvailable(),
                            open = interactive()) {
   path <- normalizePath(path, mustWork = FALSE)
--- a/R/license.R
+++ b/R/license.R
@@ -96,13 +96,13 @@ license_data <- function(name, base_path = proj_get()) {
 
 
 find_name <- function() {
-  name <- getOption("devtools.name")
-  if (!is.null(name) && name != "Your name goes here") {
+  name <- getOption("usethis.full_name")
+  if (!is.null(name)) {
     return(name)
   }
 
-  name <- getOption("usethis.full_name")
-  if (!is.null(name)) {
+  name <- getOption("devtools.name")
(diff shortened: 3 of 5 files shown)


R/description.R
```diff
@@ -48,22 +58,15 @@ build_description <- function(name, fields = list()) {
 }
 
 build_description_list <- function(name, fields = list()) {
-  author <- getOption("devtools.desc.author") %||%
```
hadley: We're abandoning these individual options in favour of the omnibus `usethis.description`? If yes, need to call out in the NEWS (although it probably doesn't affect many people)

R/description.R
```diff
@@ -1,38 +1,48 @@
...
 use_description <- function(fields = NULL) {
   name <- project_name()
   check_package_name(name)
 
-  fields <- fields %||%
-    getOption("usethis.description") %||%
-    getOption("devtools.desc") %||%
-    list()
+  ## the definitive source of user-supplied info: in this call or via options
+  fields <- utils::modifyList(
```
hadley: It might be worth doing some light type checking here - e.g. that it's a list with names?

hadley: And should all this logic be moved into `build_description()`? I think that would make the tests simpler since you wouldn't need so many temporary projects

R/description.R
jennybc: Oh, I failed to see that this wasn't just grabbing the "author" element of the "devtools.desc" list.

But it does seem like something we could probably drop. Or switch entirely to "devtools.desc.XYZ". Supporting usethis and devtools options, in individual and ombibus form seems unnecessary.

R/description.R
jennybc: I'll try to move things around.

Part of the reason I tested via `use_description()`, though, was to test that directly specified fields override options & defaults, but in a targetted manner.

hadley: Yeah, I think it would be better to move all that code out so you can test independently of the writing.

R/description.R
hadley: Yeah, I'm fine with dropping it.

jennybc:
I think this does everything you asked @hadley:

  * Drop support of individual devtools' options and document this in NEWS.
  * Add check that user's `fields` is a dictionary-ish list.
  * Consolidate
    -  data: option- and package-based defaults exposed via `use_description_defaults()`
    -  logic: user's `fields` > usethis option > devtools option > usethis defaults, enacted in `build_description_list()`
  * Tests now focus on `build_description_list()`

R/description.R
```diff
@@ -1,74 +1,89 @@
...
 #' use_description()
+#'
+#' use_description(fields = list(Language = "es"))
+#'
+#' use_description_defaults()
 #' }
 use_description <- function(fields = NULL) {
   name <- project_name()
   check_package_name(name)
+  if (!is.null(fields)) check_is_named_list(fields)
```
hadley: If you moved this one line lower you could drop the conditional.

R/description.R
```diff
@@ -1,74 +1,89 @@
...
-    Description = "What the package does (one paragraph).",
-    "Authors@R" = author,
-    License = license,
-    Suggests = suggests,
-    Encoding = "UTF-8",
-    LazyData = "true",
-    ByteCompile = "true"
+build_description_list <- function(fields = list()) {
+  defaults <- use_description_defaults()
+  defaults <- utils::modifyList(
```
hadley: It feels to me like this belongs in `use_description_defaults()`

tests/testthat/test-use-description.R
```diff
@@ -0,0 +1,81 @@
+context("use_description")
+
+test_that("build_description_list() defaults to values built into usethis", {
+  withr::with_options(
```
hadley: I think it's easier to read if you use `local_options()`

tests/testthat/test-use-description.R
```diff
@@ -0,0 +1,81 @@
...
+    }
+  )
+})
+
+test_that("build_description_list(): user's fields > usethis defaults", {
+  d <- build_description_list(
+    fields = list(Title = "aaa", URL = "https://www.r-project.org")
+  )
+  ## from user's fields
+  expect_identical(d$Title, "aaa")
```
hadley: Why test two fields from each case? Isn't one enough?

R/description.R
jennybc: I actually made `use_description_defaults()` "data only" on purpose. It's the only easy way to see the usethis field defaults w/o reading source. Unless you care deeply, will leave this.

tests/testthat/test-use-description.R
jennybc: AHA! done.

tests/testthat/test-use-description.R
jennybc: One is an overwrite, one is novel. But I did remove one test care re: default values.

PR #375, opened 2018-05-30
Refactor project data helpers

jennybc:
Main points:

  * Moved the GitHub stuff out of `package_data()` and into `project_data()`, so that it's also available for non-package projects.
  * Use `github_owner()` and `github_repo()` helpers I recently added. (*They were added because we were doing this all over the place and it's awkward because two pieces of info.*)
  * Always refer to GitHub "owner" because works equally well for individual users and orgs. It also matches GitHub API terminology.
  * Consult  `project_data()` in `project_name()`, instead of re-implementing similar logic re: DESCRIPTION or basename of path.

Question: we have an `is_proj()` helper. Should we use it in `project_data()` instead of being so easy-going? `is_proj()` will return `TRUE` if the target path fulfills any of our `proj_crit()`, which we pass along to rprojroot.

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -71,29 +71,27 @@ find_template <- function(template_name, package = "usethis") {
 
 project_data <- function(base_path = proj_get()) {
   if (is_package(base_path)) {
-    package_data(base_path)
+    data <- package_data(base_path)
   } else {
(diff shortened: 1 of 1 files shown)


hadley:
Yes, that sounds like a good idea.

PR #376, opened 2018-05-31
Get organized: owner/repo = repo_spec

jennybc:
Main points:

  * Continue to standardize user-facing args and internal vocabulary around GitHub:
    - r-lib/usethis: owner = r-lib, repo = usethis, **repo_spec = r-lib/usethis**
  * Add more helpers around GitHub owner, repo, repo_spec

Questions:

  * Is it bad to use non-exported functions in default arg values? General question.
  * Should the [`browse_*()` family](https://github.com/r-lib/usethis/blob/master/R/browse.R) switch over to `repo_spec` from `package`? Right now, target repo is driven by GitHub info found in URL field of DESCRIPTION. But maybe it should be driven by the local remote associated with GitHub? Or try one then the other?
    - Pro: everything would work for non-package projects
    - Con: less favourable if you're working with someone else's package. But how relevant is that?

Happy to discuss/review during 1-on-1.


R/create.R
```diff
@@ -141,15 +141,9 @@ create_from_github <- function(repo,
...
-  repo <- strsplit(repo, "/")[[1]]
-  if (length(repo) != 2) {
-    stop(
-      code("repo"), " must be of form ",
-      value("owner/reponame"), call. = FALSE
-    )
-  }
-  owner <- repo[[1]]
-  repo <- repo[[2]]
+  ## TODO: do I need to do anything re: repo --> repo_spec?
```
jennybc: Do I need to do much/anything about this? `create_from_github()` is mostly for interactive use, probably has few users, `repo`/`repo_spec` is the first argument (so often unnamed) and partial matching works in our favour.

R/tidyverse.R
```diff
@@ -255,43 +254,30 @@ use_tidy_style <- function(strict = TRUE) {
...
-#'   owner = "r-lib", repo = "usethis",
-#'   from = "08a560d", to = "2018-05-14"
-#' )
+#' use_tidy_thanks("r-lib/usethis", from = "08a560d", to = "2018-05-14")
 #' }
-use_tidy_thanks <- function(owner = NULL,
-                            repo = NULL,
-                            from = releases(owner = owner, repo = repo)[[1]],
+use_tidy_thanks <- function(repo_spec = github_repo_spec(),
+                            from = releases(repo_spec)[[1]],
```
jennybc: I dislike using unexported functions for the defaults. But also like revealing the default.

R/create.R
hadley: I don't think you need to

R/tidyverse.R
hadley: Should we add this as an argument to `use_github_labels()` too?

jennybc:
We've decided to leave the `browse_*()` family alone for now, but it remains on the radar for future change.

R/tidyverse.R
jennybc: I don't think we need to add `repo_spec` to `use_github_labels()`. I makes sense to me that it works on the active project.

I think `use_tidy_thanks()` is different and pretty unique because you are likely to use it while in another project, e.g. composing a post in the tidyverse.org project.

PR #377, opened 2018-05-31
Update use_tidy_eval()

### LINKED ISSUE #343: `use_tidy_eval()` creates broken link
`use_tidy_eval()` creates help page with the following sentence: "To learn more about tidy eval and how to use these tools, read <http://rlang.tidyverse.org/articles/tidy-evaluation.html>". Mentioned link is broken.

Moreover, there is no articles in the current `rlang` site.

### LINKED ISSUE #419: Remove UQ/UQS from use_tidy_eval
Since they’re deprecated

jennybc:
Fixes #343 `use_tidy_eval()` creates broken link
Fixes #419 Remove UQ/UQS from use_tidy_eval

I'm asking you @lionel- to review because it's a good occasion to re-evaluate whether the imports and re-exports in [inst/templates/tidy-eval.R](https://github.com/r-lib/usethis/blob/use-tidy-eval/inst/templates/tidy-eval.R) still make sense. For example, my impression is that `UQ` and `UQS` should not be here. Can you offer advice so I can move forward and/or make a separate PR?


lionel-:
Should we point to the meta-programming chapters of the next adv-r edition rather than to the rlang website?

About the tidy eval reexports, I think this would be the new model: https://github.com/tidyverse/dplyr/blob/master/R/utils-tidy-eval.R except that we are going to remove these exports:

```
exprs enexpr enexprs
```

The last two should rarely be used and the first clashes with a base bioconductor function.

I will send a PR with an updated template.

hadley:
It really should point to a website that does not yet exist that has a page that discusses the quote/unquote pattern, a page that provides a bunch of recipes, and a page that gives an overview of the theory (with pointers to adv-r for more details)

jennybc:
Well, let's just fix the imports/exports, use a link that exists today, and update it when a better target exists.

jennybc:
Any update @lionel-? We are inching closer to a release here, though it's not exactly imminent.

jennybc:
@lionel- @hadley 

OK I just took a stab at this.

  * Used https://github.com/tidyverse/dplyr/blob/master/R/utils-tidy-eval.R as inspiration.
  * Template links to top-level of rlang site and metaprogramming section of Adv-R.
  * Removed roxygen2 pre-req and Suggest it instead.

R/tidyverse.R
```diff
@@ -137,10 +137,7 @@ dep_version <- function(x) {
 #' @export
 #' @rdname tidyverse
 use_tidy_eval <- function() {
-  if (!uses_roxygen()) {
-    stop_glue("{code('use_tidy_eval()')} requires that you use roxygen.")
-  }
-
+  use_dependency("roxygen2", "Suggests")
```
jennybc: I don't see why we required evidence of prior use of roxygen2. Seems like we can just make sure it is, at least, in Suggests and carry on.

R/tidyverse.R
```diff
@@ -137,10 +137,9 @@ dep_version <- function(x) {
 #' @export
 #' @rdname tidyverse
 use_tidy_eval <- function() {
-  if (!uses_roxygen()) {
-    stop_glue("{code('use_tidy_eval()')} requires that you use roxygen.")
-  }
+  check_is_package("use_tidy_eval()")
 
+  use_dependency("roxygen2", "Suggests")
```
jennybc: I don't see why we required evidence of prior use of roxygen2. Seems like we can just make sure it is, at least, in Suggests and carry on.

Conversely, I think we should make sure we're dealing with a package.

inst/templates/tidy-eval.R
```diff
@@ -1,15 +1,44 @@
 #' Tidy eval helpers
 #'
-#' These functions provide tidy eval-compatible ways to capture
-#' symbols (`sym()`, `syms()`, `ensym()`), expressions (`expr()`,
-#' `exprs()`, `enexpr()`), and quosures (`quo()`, `quos()`, `enquo()`).
-#' To learn more about tidy eval and how to use these tools, read
-#' <http://rlang.tidyverse.org/articles/tidy-evaluation.html>
+#' @description
```
jennybc: I tried to do what @lionel- said here:

https://github.com/r-lib/usethis/pull/377#issuecomment-393437046

lionel-:
@hadley Would it be possible to make `?dplyr::quo` and `?rlang::quo` point to the same documentation page? From discussions on Twitter, it may be likely that people don't notice or follow links and miss out on the documentation.

hadley:
The latest version of this file is at https://github.com/tidyverse/ggplot2/blob/master/R/utilities-tidy-eval.R

I think it's a little to aggressive to use `@import rlang` in a template that is used in situations outside of our control.

PR #379, opened 2018-05-31
Put a floor on clipr, rprojroot

### LINKED ISSUE #351: Aggressively state minimum versions of dependencies
@hadley's May 17/18 workshop reveals usethis still has some undocumented minimum version requirements for its Imports. The specific one he mentioned is rprojroot.

jennybc:
Fixes #351 Aggressively state minimum versions of dependencies

clipr and rprojroot are concrete examples where we use specific functions that I can see were added at a specific version, based on reading NEWS.

I contemplated every other package in Imports and Suggests and couldn't find a specific reason to state a minimum version.

How aggressive should I be?
Any vague notions that I should put a floor on anything else?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -21,7 +21,7 @@ Depends:
     R (>= 3.1)
 Imports: 
     backports (>= 1.1.0),
-    clipr,
+    clipr (>= 0.3.0),
     clisymbols,
     crayon,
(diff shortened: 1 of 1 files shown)


hadley:
In the future, we should have some systematic way to decide on the minimum version, but lets not think about that now.

PR #380, opened 2018-06-01
Refactor use_course() and gain progress reporting

### LINKED ISSUE #276: Refactor use_course()
Some code shared by @jeroen in a Slack conversation about `use_course()`

``` r
library(curl)

download_with_cd <- function(url, destdir = "."){
  h <- new_handle(noprogress = FALSE)
  tmp <- tempfile()
  curl::curl_fetch_disk(url, tmp, handle = h)
  headers <- curl::parse_headers_list(handle_data(h)$headers)
  filename <- stringr::str_match(headers[["content-disposition"]], "filename=.*")[1]
  filename <- sub("filename=", "", filename)
  outfile <- normalizePath(file.path(destdir, filename), mustWork = FALSE)
  file.rename(tmp, outfile)
  return(outfile)
}

download_with_cd("https://cran.ocpu.io/ggplot2/data/diamonds/csv", destdir = "~")
```

Other notes:

  * If you use `curl_fetch_disk` you can check the `status` code also from `handle_data()`.
  * Using `curl_download()` might be easier -- automatically throws an error if the result is not http 200
  * the trick is all functions in curl take a handle, and you can always use `handle_data()` to access the metadata after the request

jennybc:
Fixes #276 Refactor use_course()

Thanks @jeroen for all the pointers!

Key things:

  * Explicit use of a handle is how we hook into progress
    reporting, even though curl_download() doesn't expose
  * curl_download() has a built-in check for http status 200
  * Using a custom progressfunction because default doesn't
    always look right in RStudio
  * Switch to fs functions for file system work

### DIFF
--- a/R/course.R
+++ b/R/course.R
@@ -191,7 +191,7 @@ download_zip <- function(url, destdir = getwd(), pedantic = FALSE) {
     message(
       "A ZIP file named:\n",
       "  ", value(base_name), "\n",
-      "will be written to this folder:\n",
+      "will be copied to this folder:\n",
       "  ", value(base_path), "\n",
       "Prefer a different location? Cancel, try again, and specify ",
       code("destdir"), ".\n"
@@ -204,7 +204,7 @@ download_zip <- function(url, destdir = getwd(), pedantic = FALSE) {
 
   if (!can_overwrite(full_path)) {
     ## TO DO: it pains me that can_overwrite() always strips to basename
-    stop("Aborting download", call. = FALSE)
+    stop("Aborting.", call. = FALSE)
   }
 
   done(
--- a/R/course.R
+++ b/R/course.R
@@ -172,12 +172,14 @@ NULL
 
 download_zip <- function(url, destdir = getwd(), pedantic = FALSE) {
   stopifnot(is_string(url))
-  dl <- curl::curl_fetch_memory(url)
+  h <- curl::new_handle(noprogress = FALSE, progressfunction = progress_fun)
+  tmp <- fs::file_temp("usethis-use-course-")
+  curl::curl_download(
+    url, destfile = tmp, quiet = FALSE, mode = "wb", handle = h
+  )
+  check_is_zip(h)
 
-  stop_for_status(dl$status_code)
-  check_is_zip(dl)
-
-  cd <- content_disposition(dl)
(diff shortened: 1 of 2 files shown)


jennybc:
Quick screenshot snagged while downloading some hefty course materials from @hadley 

screen shot 2018-06-01 at 2 22 31 pm

This updates in place during the download.

R/course.R
```diff
@@ -328,6 +331,19 @@ parse_content_disposition <- function(cd) {
   )
 }
 
+progress_fun <- function(down, up) {
```
jennybc: The default curl progress bar gets mangled in (some versions of?) RStudio, so I took @jeroen's advice to copy an approach from tesseract.

https://github.com/ropensci/tesseract/blob/6514c5862a6e6991c45048449b20fd24babf4bb4/R/tessdata.R#L63-L74

jennybc:
@jeroen I can't formally request a review from you and I think I followed your advice fairly faithfully, but feel free to comment.

jeroen:
Yes I like it, nice how it saves it under the server suggested filename. Maybe somewhere insert a `cat("\n")` after the download has finished to print the next thing to a new line?

screen shot 2018-06-01 at 11 37 43 pm

R/course.R
```diff
@@ -188,28 +191,27 @@ download_zip <- function(url, destdir = getwd(), pedantic = FALSE) {
     message(
       "A ZIP file named:\n",
       "  ", value(base_name), "\n",
-      "will be downloaded to this folder:\n",
+      "will be written to this folder:\n",
```
hadley: copied?

R/course.R
```diff
@@ -188,28 +191,27 @@ download_zip <- function(url, destdir = getwd(), pedantic = FALSE) {
...
+    if (nope("Is it OK to write this file here?")) {
+      stop("Aborting", call. = FALSE)
     }
   }
-  full_path <- file.path(base_path, base_name)
+  full_path <- fs::path(base_path, base_name)
 
   if (!can_overwrite(full_path)) {
     ## TO DO: it pains me that can_overwrite() always strips to basename
     stop("Aborting download", call. = FALSE)
```
hadley: Reword because it's not the download anymore?

PR #383, opened 2018-06-05
Refashion use_tibble()

jennybc:
Make `use_tibble()` do less. Specifically, do not add an actual dummy `.R` file and don't include any directives that trigger a new export in NAMESPACE or an `.Rd` file.

Do produce the necessary roxygen directive and provide support re: where it should be stored. Beef up the help.


jimhester:
LGTM!

PR #385, opened 2018-06-06
Switch to fs

### LINKED ISSUE #177: Use fs when released

jennybc:
Fixes #177 Use fs when released

  * Switches to fs (<https://fs.r-lib.org>) for all file system operations. This implies that backports is no longer a dependency (was using for `dir.exists()`).
  * Other general refactoring of path handling. There were many remaining opportunities to advance the transition to the "active project" approach, kicked off in 5375bd447ea73ed26ee6a558841e61845bde694c.
  * Miscellaneous cleanup, continue to use `glue::glue()` more.
 
Implications of the switch to fs that require the most thought:

  * Different default behaviour around user's home directory on Windows. Documented in [`fs::path_expand()`](https://fs.r-lib.org/reference/path_expand.html).
    - What R usually regards as home on Windows: `C:\Users\username\Documents`
    - What fs (and Git, ssh, Python, and rust) regard as home on Windows: `C:\Users\username`
    - usethis now inherits the fs behaviour
  * For each instance of `normalizePath()`, had to ask why it's there and what is the correct fs substitution? I decided that `fs::path_real()` is the usethis policy for processing (i.e. storing) the path to the active project. And that `fs::path_expand()` should always be applied to user-specified locations, so that usethis is self-consistent re: what `~` means.
  * Paths can now be reported in a much more humane way in messages, typically relative to the active project.

More comments inline.


tests/testthat/test-create.R
```diff
@@ -65,7 +69,7 @@ test_that("rationalize_fork() won't attempt to fork w/o PAT", {
   )
   expect_error(
     rationalize_fork(fork = TRUE, repo_info = list(), pat_available = FALSE),
-    "No GitHub Personal Access Token available"
+    "No GitHub .+auth_token.+"
```
jennybc: A very low-tech solution to dealing with an error message that might contain ANSI escapes.

tests/testthat/test-edit.R
```diff
@@ -1,37 +1,49 @@
...
+  ## have them or delete them
+  skip_if_not_ci()
+
+  capture_output(expect_error_free(edit_r_profile()))
+  capture_output(expect_error_free(edit_r_environ()))
+  capture_output(expect_error_free(edit_r_makevars()))
+  capture_output(expect_error_free(edit_git_config()))
+  capture_output(expect_error_free(edit_git_ignore()))
+})
+
```
jennybc: These functions previously didn't work if called with no input 🙈 Needed to call `match.arg()` before passing `scope` onto `switch()`.

tests/testthat/test-helpers.R
```diff
@@ -1,5 +1,61 @@
 context("helpers")
 
+test_that("create_directory() requires base_path to be pre-existing directory", {
```
jennybc: These helpers were already hit by tests, but the coverage wasn't very direct.

tests/testthat/test-use-course.R
```diff
@@ -87,79 +87,7 @@ test_that("make_filename() uses fallback if no `content-disposition` header", {
   expect_match(make_filename(NULL), "^file[0-9a-z]+$")
 })
 
-## mostly inspired by
```
jennybc: Now using `fs::path_sanitize()` 🎉

tests/testthat/test-write.R
```diff
@@ -1,9 +1,46 @@
...
 test_that("same_contents() detects if contents are / are not same", {
-  tmp <- tempfile()
+  tmp <- file_temp()
   x <- letters[1:3]
   writeLines(x, con = tmp, sep = "\n")
   expect_true(same_contents(tmp, x))
   expect_false(same_contents(tmp, letters[4:6]))
 })
+
+test_that("write_union() writes a de novo file", {
```
jennybc: Again, there weren't many direct unit tests of these helpers.

jimhester:
Looks like the tests are failing because of a missing `)` in the glue expression at https://github.com/r-lib/usethis/blob/fs/R/write.R#L16

R/course.R
```diff
@@ -263,11 +263,12 @@ normalize_url <- function(url) {
 }
 
 conspicuous_place <- function() {
-  Filter(dir.exists, c(
-    file.path(Sys.getenv("HOME"), "Desktop"), # typical macOS = ~/Desktop
-    file.path(Sys.getenv("USERPROFILE"), "Desktop"), # typical Windows Desktop
-    "~",
-    getwd()
+  Filter(dir_exists, c(
+    path(Sys.getenv("HOME"), "Desktop"), # typical macOS = ~/Desktop
```
jennybc: This is probably redundant with next line.

R/create.R
```diff
@@ -184,41 +175,41 @@ create_from_github <- function(repo_spec,
     ssh = repo_info$ssh_url
   )
 
-  done("Cloning repo from ", value(origin_url), " into ", value(repo_path))
+  done(glue("Cloning repo from {value(origin_url)} into {value(repo_path)}"))
```
jimhester: I wonder if we should just redefine done, todo etc. to use glue directly rather than paste0?

https://github.com/r-lib/usethis/blob/ca21abeb936ff75d41a3f39e421a919d7ba1538c/R/style.R#L11-L13

e.g.

```r
done <- function(...) {
  bullet(glue(...), bullet = crayon::green(clisymbols::symbol$tick))
}

File: R/r.R [hunk 8]
```diff
@@ -15,38 +18,42 @@ use_r <- function(name = NULL) {
...
-find_r_name <- function(name = NULL) {
-  if (!is.null(name)) {
-    check_file_name(name)
-    return(slug(name, ".R"))
-  }
-
+get_active_r_file <- function(path = "R") {
   if (!rstudioapi::isAvailable()) {
-    stop("Argument ", code("name"), " is missing, with no default", call. = FALSE)
+    stop(glue("Argument {code('name')} must be specified."), call. = FALSE)
```
jimhester: It feels like we are doing a lot of `stop(glue(` etc, maybe worth creating a helper function for it, e.g. something like https://github.com/r-lib/fs/blob/6ee3322076bd2cb6b9cacd0b52ac1717b11996e4/R/utils.R#L113-L123.

Or we can wait on this until things are more sure in rlang.

R/utils.R
```diff
@@ -80,13 +78,18 @@ dots <- function(...) {
   eval(substitute(alist(...)))
 }
 
-slug <- function(x, ext) {
+asciify <- function(x) {
   stopifnot(is.character(x))
 
   x <- tolower(x)
-  x <- gsub("[^a-z0-9_]+", "-", x)
+  gsub("[^a-z0-9_]+", "-", x)
```
jimhester: I wonder if we should include `-` in the character class, right now we are substituting `-` with itself, e.g. change the regex to

```r
gsub("[^a-z0-9_-]+", "-", x)
```

tests/manual/manual-create-from-github.R
```diff
@@ -1,47 +1,68 @@
-load_all()
+devtools::load_all()
```
jimhester: Maybe it would useful to have a README.md in the tests/manual directory explaining how you typically run these.

e.g. do you use `source()`, do you just run them line by line in the console, is there setup you need to do prior to running etc.

It doesn't need to be long, a few sentences would probably be ok.

DESCRIPTION
```diff
@@ -20,7 +20,6 @@ BugReports: https://github.com/r-lib/usethis/issues
 Depends: 
     R (>= 3.1)
 Imports: 
-    backports (>= 1.1.0),
     clipr (>= 0.3.0),
     clisymbols,
     crayon,
```
hadley: Should we need to require a specific version of fs?

NEWS.md
```diff
@@ -16,7 +16,12 @@
 
 * `use_github()` fails earlier, with a more informative message, in the absence of a GitHub personal access token (PAT). Also looks for the PAT more proactively in the usual environment variables (i.e., GITHUB_PAT, GITHUB_TOKEN) (#320, #340, @cderv).
 
-* usethis newly depends on fs and, before the next release, will use it for all file and path operations. (#177)
+* All file system operations use the [fs](https://fs.r-lib.org) package (#177). Therefore backports is no longer a dependency. This should not change how usethis functions, but users may notice these features of fs-mediated paths:
+
+  - Paths are "tidy", meaning `/` is the path separator and there are never multiple or trailing `/`.
+  - Paths are UTF-8 encoded.
+  - A Windows user's home directory is interpreted as `USERPROFILE` or `HOMEDRIVE`/`HOMEPATH`, by default. Read more in [`fs::path_expand()`](https://fs.r-lib.org/reference/path_expand.html).
```
hadley: As well as the existing text, I think you also need a more using facing description here, something like: "On windows, `~` now refers your home directory, not your documents directory"

NEWS.md
```diff
@@ -16,7 +16,12 @@
...
+  - Paths are typically expressed relative to the active project.
```
hadley: In output?

R/code-of-conduct.R
```diff
@@ -16,7 +16,7 @@ use_code_of_conduct <- function(path = NULL) {
   if (!is.null(path)) {
     use_directory(path, ignore = is_package())
   }
-  save_as <- fs::path_join(c(path, "CODE_OF_CONDUCT.md"))
+  save_as <- path_join(c(path, "CODE_OF_CONDUCT.md"))
```
hadley: Should this be just `path(path, "CODE_OF_CONDUCT.md")`?

R/course.R
hadley: Yeah, I think you should remove that line now (and the comment on the next line)

R/course.R
```diff
@@ -263,11 +263,12 @@ normalize_url <- function(url) {
...
+    path_home("Desktop"), # typical Windows Desktop
+    path_home(),
+    path_home_r(),
```
hadley: Why both `path_home()` and `path_home_r()`?

R/create.R
```diff
@@ -51,7 +48,7 @@ create_package <- function(path,
     use_rstudio()
   }
   if (open) {
-    open_project(path, name, rstudio = rstudio)
+    open_project(proj_get())
```
hadley: The previous behaviour was deliberate, so I could launch a new RStudio project from the terminal.

R/git.R
```diff
@@ -54,22 +54,26 @@ use_git <- function(message = "Initial commit") {
...
 #'
 #' @param ignores Character vector of ignores, specified as file globs.
-#' @param directory Directory within current project to set ignores
+#' @param directory Directory relative to active project to set ignores
 #' @family git helpers
 #' @export
 use_git_ignore <- function(ignores, directory = ".") {
-  write_union(proj_get(), file.path(directory, ".gitignore"), ignores)
+  write_union(
+    ## path_norm() needed for this: './.gitignore' --> '.gitignore'
```
hadley: Should that happen in `proj_path()` then?

R/github.R
```diff
@@ -187,8 +180,11 @@ use_github_links <- function(auth_token = NULL,
   )
 
   use_description_field("URL", res$html_url, overwrite = overwrite)
-  use_description_field("BugReports", file.path(res$html_url, "issues"),
-    overwrite = overwrite)
+  use_description_field(
+    "BugReports",
+    paste(res$html_url, "issues", sep = "/"),
```
hadley: `paste0(res$html_url, "/issues")` ?

DESCRIPTION
jennybc: Yes, I should require `fs (>= 1.2.0)` to ensure everything we assume and say about home directory handling on Windows is true.

NEWS.md
jennybc: Done. Used the concrete path examples from the description of this PR.

R/code-of-conduct.R
jennybc: No, it has to be this way to gracefully handle the case where `path = NULL`.

``` r
library(fs)

path <- ".github"
path_join(c(path, "CODE_OF_CONDUCT.md"))
#> .github/CODE_OF_CONDUCT.md
path(path, "CODE_OF_CONDUCT.md")
#> .github/CODE_OF_CONDUCT.md

path <- NULL
path_join(c(path, "CODE_OF_CONDUCT.md"))
#> CODE_OF_CONDUCT.md
path(path, "CODE_OF_CONDUCT.md")
#> character(0)
```

Created on 2018-06-18 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

R/course.R
jennybc: IDK ... belt and suspenders? I agree it's hard to imagine `path_home_r()` existing and `path_home()` not existing, but I suppose it's possible if someone has fiddled with their env vars. In terms of people finding their downloaded stuff, `C:\Users\username\Documents` still seems more conspicuous than falling back to working directory.

R/create.R
jennybc: I don't think I meant to change functionality 😬. You think I did? 

I was somewhat confused, though, that the `rstudio` argument of `create_*()` functions controls the addition of an Rproj file, whereas the `rstudio` argument of `open_project()` is about attempting to open in RStudio vs. changing wd of current R session.

R/git.R
jennybc: Sure, that works too. Done.

R/r.R
jennybc: Yeah, I will use glue more, but in another PR.

R/create.R
hadley: If you don't leave `rstudio = rstudio` here, I'm pretty sure it won't do what I want anymore.

jennybc: If I go back to a version prior to this PR and call `create_package()` in the terminal, without specifying `rstudio` at all and with specifying `rstudio = TRUE`, I don't get the new RStudio project launched either way. The new package is definitely an RStudio project, but even if I attempt to call `rstudioapi::openProject(proj_get())` explicitly I see:

``` r
> rstudioapi::openProject(proj_get())
Error: RStudio not running
```

Sorry if I am just misunderstanding what you mean by "launch a new RStudio project from the terminal".

hadley: I often run `create_package(..., rstudio = TRUE)` in the terminal. I’m pretty sure that used to work for me. (But maybe it didn’t or it was a long time before this PR)

jennybc: Yes I think you are either misremembering or we lost that behaviour a while ago.

PR #388, opened 2018-06-17
Add codeowners w/ understudy

batpigandme:
* Add CODEOWNERS w/ @hadley as principal, @jennybc as understudy.


jennybc:
Thanks!

Does it make sense that the URL doesn't go anywhere yet?

batpigandme:
Yeah, the dummy url is per Hadley's request, but I'll try to get something in there this week.

jennybc:
I figured.

PR #389, opened 2018-06-19
Export file-writing helpers

### LINKED ISSUE #344: Export edit_file
(This is @hadley on @behrman's computer)

### LINKED ISSUE #366: any chance you'd consider exporting write_union?
Was just writing something to do this when I realized `usethis` must already be using such a function to update `.Rbuildignore` and `.gitignore` lists; and of course it does and looking at the source of `write_union` I can easily see your version is better engineered than mine.  Of course I could adapt my code, but since it would then be a modification / derivation from `usethis` I should be crediting `usethis` directly and switching my license to GPL-3, and possibly listing `usethis` authors as `ctb` in the DESCRIPTION (though I sometimes worry that is misleading without proper context).  All of which is obviously more cumbersome than if I could just call it as an exported function.  Of course I realize my weird case is hardly a strong general case for exporting this.

jennybc:
Closes #366 any chance you'd consider exporting write_union?
Closes #344 Export edit_file 

* Re-locate some functions. `helper.R` had gotten quite large and diffuse.
* Make `is_in_proj()` resilient to there being no active project
* Export `edit_file()`, `write_union()`, `write_over()`; mark with `@keywords internal`
* Process input `path` with `user_path_prep()` now that it's user-accessible


jennybc:
Might be nice to add "Export edit_file" #344 to this PR, yes?

R/write.R
```diff
@@ -1,5 +1,31 @@
...
+#'
+#' @examples
+#' tmp <- tempfile()
+#' write_union(tmp, letters[1:3])
+#' readLines(tmp)
+#' write_union(tmp, letters[1:5])
+#' readLines(tmp)
+#'
+#' ## clean up
+#' file.remove(tmp)
```
hadley: I wonder if there's a non-awkward way to give this a name starting with `edit_`? That would make it more clearly a member of an existing family.

R/edit.R
```diff
@@ -1,3 +1,39 @@
...
+#' @keywords internal
+#'
+#' @examples
+#' \dontrun{
+#' edit_file("DESCRIPTION")
+#' edit_file("~/.gitconfig")
+#' }
+edit_file <- function(path) {
+  path <- user_path_prep(path)
+  dir_create(path_dir(path), recursive = TRUE)
```
hadley: Should this call `use_directory()`? The additional messaging seems like it would be nice.

R/write.R
jennybc: Yeah. But perhaps `write_union()` and `write_utf()` should also be exported -- why export only `write_union()`? And then they make their own little family.

R/edit.R
jennybc: Hmm, I'm a bit boxed into a corner atm. `use_directory()` assumes that input `path` is relative to the active project. Whereas all the `write_*()` helpers and `edit_file()` assume the input path can be taken at face value.

I'll take this as evidence that the functions re: directory creation aren't quite right yet. Will work on that.

hadley: Oh good point; can definitely happen after this PR.

R/write.R
hadley: write_utf8 definitely belongs elsewhere but seems good to export write_over

R/edit.R
jennybc: Will sort this out here: https://github.com/r-lib/usethis/issues/393

jennybc:
Assuming CI goes well, I think this is done.

cboettig:
🎉

PR #395, opened 2018-06-21
Restore original project if new project launches in new RStudio instance

### LINKED ISSUE #390: When rstudio = TRUE, don't change active project in create_package()
Because it's confusing — you end up with two RStudio projects (the old and the new) with the active project set to the package that you just created.

jennybc:
Fixes #390 When rstudio = TRUE, don't change active project in create_package()

Reminder, because I have tripped on this a lot myself: the `rstudio` argument of `create_*()` controls whether an Rproj file is added to the new project. `open` is the argument that controls if/how the new project is launched, i.e. whether `open_project()` is called (which has its own `rstudio` argument, with a different meaning).

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -37,7 +37,7 @@ create_package <- function(path,
 
   create_directory(path_dir(path), name)
   cat_line(crayon::bold("Changing active project to", crayon::red(name)))
-  proj_set(path, force = TRUE)
+  op <- proj_set(path, force = TRUE)
 
   use_directory("R")
   use_directory("man")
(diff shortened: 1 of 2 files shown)


R/create.R
```diff
@@ -47,8 +47,16 @@ create_package <- function(path,
   if (rstudio) {
     use_rstudio()
   }
+
   if (open) {
-    open_project(proj_get())
+    fresh_rstudio <- open_project(proj_get())
```
hadley: Make this into a function?

jennybc:
I've incorporated this into a new branch and PR.

PR #397, opened 2018-06-27
Project activation

### LINKED ISSUE #390: When rstudio = TRUE, don't change active project in create_package()

jennybc:
Fixes #390 When rstudio = TRUE, don't change active project in create_package()

Generally tightens up project activation:

  * ~Quietly attempt to activate project on load (should this be on attach?)~
  * Announce project activation, by default
  * Rename `is_proj(path)` to `possibly_in_proj(path)` to better reflect what it actually reports
  * ~Store current project *and the previous*, so we can have `proj_revert()`~

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -36,7 +36,6 @@ create_package <- function(path,
   check_not_nested(path_dir(path), name)
 
   create_directory(path_dir(path), name)
-  cat_line(crayon::bold("Changing active project to", crayon::red(name)))
   proj_set(path, force = TRUE)
 
   use_directory("R")
@@ -64,7 +63,6 @@ create_project <- function(path,
   check_not_nested(path_dir(path), name)
 
   create_directory(path_dir(path), name)
-  cat_line(crayon::bold("Changing active project to", crayon::red(name)))
   proj_set(path, force = TRUE)
 
   use_directory("R")
--- a/R/create.R
+++ b/R/create.R
@@ -213,9 +213,7 @@ open_project <- function(path, rstudio = NA) {
 }
 
 check_not_nested <- function(path, name) {
-  path_is_proj <- is_proj(path)
-
-  if (!path_is_proj) {
+  if (!possibly_in_proj(path)) {
     return(invisible())
   }
 
--- a/R/create.R
+++ b/R/create.R
@@ -47,7 +47,10 @@ create_package <- function(path,
     use_rstudio()
   }
   if (open) {
-    open_project(proj_get())
+    fresh_rstudio <- open_project(proj_get())
+    if (fresh_rstudio) {
+      proj_revert()
+    }
   }
 
   invisible(TRUE)
@@ -76,7 +79,10 @@ create_project <- function(path,
     file_create(proj_path(".here"))
   }
   if (open) {
-    open_project(proj_get())
+    fresh_rstudio <- open_project(proj_get())
+    if (fresh_rstudio) {
+      proj_revert()
+    }
   }
 
   invisible(TRUE)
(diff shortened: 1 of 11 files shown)


R/zzz.R
```diff
@@ -0,0 +1,3 @@
+.onLoad <- function(libname, pkgname) {
+  try(proj_set(".", quiet = TRUE), silent = TRUE)
+}
```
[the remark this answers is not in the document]
hadley: Why is this needed now, but not previously? I don't really understand how project activation is supposed to work.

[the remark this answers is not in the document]
jennybc: Oh, there is no new need! In most usage scenarios, this doesn't matter. You do notice it when developing usethis itself. This activates a project after `load_all()`, which is just nice.

I would say project activation is supposed to work "on first demand".

Another consequence of auto-activate, more relevant to other people: All `use_*()` functions that operate on a project eventually make a `proj_get()` and that will activate a project. But if the first usage is of a helper like `write_union()`, that won't activate a project, but it *will* honour an already-activated project. This does not change what is written or where, but makes the path reported in the message nicer (i.e. relative to the project instead of absolute).

jennybc:
BTW here's what the new messaging/flow looks like. Project activation is announced, but is no longer in bold and red. If new project was launched in RStudio, host process reverts to original state.

screen shot 2018-06-27 at 7 26 17 am

R/zzz.R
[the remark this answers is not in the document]
jennybc: If auto-activation seems like a bad idea, I could also deal with the usethis development hangnail via something in the test helper file.

hadley:
Candidate function (name needs work):

```R
open_rstudio <- function(open, old_project) {
  if (!open)
    return()
  
  fresh_rstudio <- open_project(proj_get())
  if (!fresh_rstudio) 
    return()
  
  proj_set(old_project, force = TRUE)
}
```

(Assuming we revert `proj_revert()` and use the pattern in the previous PR)

jennybc:
Also record the unresolved dilemmas about the relationship between `write_*()`-type helpers and the active project.

R/create.R
```diff
@@ -194,30 +193,36 @@ create_from_github <- function(repo_spec,
...
   }
+  invisible(TRUE)
+
 }
 
-open_project <- function(path, rstudio = NA) {
-  rproj_path <- rproj_path(path)
+## `rstudio` arg here is about whether to attempt a launch in RStudio
+## `rstudio` arg of `create_*()` functions is about whether to add .Rproj file
+open_project <- function(path, restore = NA, rstudio = NA) {
```
jennybc: The default for `restore` must be `NA`, because `NULL` is a valid value to restore, i.e. we could need to restore the state of "no active project".

tests/testthat/helper.R
```diff
@@ -1,3 +1,6 @@
+## attempt to activate a project, which is nice during development
+try(proj_set("."))
+
```
jennybc: New solution to the "no active project" problem during usethis development.

jennybc:
@hadley I think is good to go.

  * Added principles.md to document design decisions and dilemmas.
  * Back to old way of restoring project state, i.e. reverted `proj_revert()`.
  * Refactored `open_project()` to do project restoration.

jennybc:
Spurious AppVeyor failure coming from fact that we depend on new styler release and Windows binaries aren't up yet.

R/create.R
```diff
@@ -194,30 +193,36 @@ create_from_github <- function(repo_spec,
...
-    setwd(path)
-    done("Changing working directory to {value(path)}")
+    done("Opening new project {value(path_file(path))} in RStudio")
+    rstudioapi::openProject(rproj_path(path), newSession = TRUE)
+    ## TODO: check this is correct on rstudio server / cloud
+    if (!is.na(restore)) {
+      proj_set(restore, force = TRUE)
+    }
+    return(invisible(TRUE))
   }
```
hadley: (Only commenting because we've been discussing early returns a lot)

This doesn't feel quite right as an early return to me. It would feel better as true and false branches of an if statement, because it's more symmetric: either open in RStudio or change the working directory.

principles.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Active project
+
+Many usethis functions act on the **active project**, the path to which is stored in the internal environment `proj`, specifically in `proj$cur`. We do this instead of constantly passing around a base path. It is implied that such functions create or modify files inside the active project. This is mostly true of `use_*()` functions, though there are exceptions. For example, `use_course()` makes no reference to the active project.
+
+The project is activated upon first need, i.e. eventually some function calls `proj_get()` and, if `proj$cur` is `NULL`, we attempt to activate a project at (or above) current working directory.
+
+### Activation upon load or attach? No.
+
+We've contemplated project activation in `.onLoad()` or `.onAttach()`, but it's not clear which is more appropriate. Which suggests that neither is appropriate. If we ever do this, `zzz.R` would include something like this:
```
hadley: Two more reasons not to activate on load/attach:

* If on attach, would mean that an package that imported usethis would also need to set on attach

* Current workflow allows you to open R, then `setwd()`, then usethis (not a workflow that we recommend, but one that is common in practice)

principles.md
```diff
@@ -0,0 +1,31 @@
+# usethis design principles
+
+*This is an experiment in making key package design principles explicit, versus only implicit in the code. The goal is to make maintenance easier, when spread out over time and across people.*
+
+## Active project
+
+Many usethis functions act on the **active project**, the path to which is stored in the internal environment `proj`, specifically in `proj$cur`. We do this instead of constantly passing around a base path. It is implied that such functions create or modify files inside the active project. This is mostly true of `use_*()` functions, though there are exceptions. For example, `use_course()` makes no reference to the active project.
```
hadley: We do this instead of constantly passing around a base path or relying on the working directory.

PR #399, opened 2018-06-27
Filter results for only issues created after the from_timestamp

jimhester:
This avoids erroneously including issues which were merely updated after
the from timestamp

### DIFF
--- a/R/tidyverse.R
+++ b/R/tidyverse.R
@@ -286,10 +286,14 @@ use_tidy_thanks <- function(repo_spec = github_repo_spec(),
     return(invisible())
   }
 
+  creation_time <- function(x) {
+    as.POSIXct(pluck_chr(x, "created_at"))
+  }
+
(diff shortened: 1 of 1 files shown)


R/tidyverse.R
```diff
@@ -286,9 +286,11 @@ use_tidy_thanks <- function(repo_spec = github_repo_spec(),
     return(invisible())
   }
 
+  res_POSIXct <- as.POSIXct(pluck_chr(res, "created_at"))
+  res <- res[res_POSIXct >= from_timestamp]
+
```
jennybc: We have to do something about the fact that `res_POSIXct` may no longer be synced with `res` before the next block, where it gets used again.

jimhester: You are right, made a change which should fix this and hopefully clean it up a little.

jennybc:
Thanks!

PR #409, opened 2018-07-05
Add "__MACOSX" and ".DS_Store" to ignores

vnijs:
The `keep_lgl` function is very handy to ensure certain files and folders are removed when using `use_course`. Two other candidates are the "__MACSOSX" directory and ".DS_Store". This PR adds to the default value for the `ignores` argument.


jennybc:
Yeah I noticed the `__MACOSX` thing myself this week. Good call.

PR #412, opened 2018-07-17
add encoding argument to readLines calls

dpprdan:
This is a small fix that adds the encoding argument to all `readLines()` calls in `R`. 

I noticed that the LICENSE.md file had some encoding issues when adding it via e.g. `use_gpl3_license()` (on Windows of course). Line 5 would for example read: 

`_Copyright Â© 2007 Free Software Foundation, Inc. &lt;<http://fsf.org/>&gt;_`

The problem is, is that incoming strings are not marked as "UTF-8". I tracked the problem to `use_template` which calls `readLines` without the `encoding` argument, which on Windows means that they are assumed to be in the "native" encoding. All template files are UTF-8 encoded, however. So I added `encoding = "UTF-8"`. 

I assume that all files generated and read by usethis are also supposed to be UTF-8 as well, so I looked up the other `readLines` calls in `R` as well and added the `encoding = "UTF-8"` argument there as well. 

I was not able to run the relevant tests because I get a lot of `could not find function "file_temp"` errors and unfortunately don't have time to sort those out at the moment.

### DIFF
--- a/R/badge.R
+++ b/R/badge.R
@@ -132,6 +132,6 @@ has_badge <- function(href) {
     return(FALSE)
   }
 
-  readme <- readLines(readme_path)
+  readme <- readLines(readme_path, encoding = "UTF-8")
   any(grepl(href, readme, fixed = TRUE))
 }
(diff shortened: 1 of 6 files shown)


jennybc:
Thanks!

PR #424, opened 2018-08-03
Put user-facing output under control of 'usethis.quiet' option

### LINKED ISSUE #416: Need way to turn off messaging globally for usethis functions
It would be useful to be able to turn off the status printing of usethis temporarily, perhaps by setting an option?

Particularly in the devtools case where I need to temporarily change the active project often the messages are not useful. You could also use this to disable them while running tests to avoid needing all of the `capture_output()` calls.

jennybc:
Fixes #416 Need way to turn off messaging globally for usethis functions

Well, this was pretty satisfying!

I'm asking for review just to make sure I implemented as discussed and to see if there's anything else I can do along these lines to make life easier for someone using usethis in another package.

I'm also continuing to record implicit design principles *explicitly* in `principles.md`.

All of usethis's user-facing output goes through the `cat_line()` helper. I've put it under the control of a undocumented option "usethis.quiet".

So to use usethis functions quietly in a function do:

``` r
withr::local_options(list(usethis.quiet = TRUE))
```

cc @jimhester @isteves


principles.md
```diff
@@ -42,3 +42,12 @@ Uncomfortable fact: `write_union()` uses the active project, if such exists, to
...
+
+## Communicating with the user
+
+User-facing messages are emitted via helpers in `style.R` and *everything* is routed through `cat_line()`. That is intentional and should be preserved. This function is under the control of an undocumented option: `usethis.quiet`. When `getOption("usethis.quiet", FALSE)` evaluates to `TRUE`, `cat_line()` does nothing. This eliminates the need for ubiquitous `capture_output()` calls in the usethis tests and makes it easier for other packages to call functions in usethis and suppress the usual messaging.
+
+You might also notice that usethis communicates with the user via `cat()` instead of `message()`. Why?
+
+  * Pragmatic explanation: default styling of `message()` (at least in RStudio) is red, which suggests that something is wrong. We prefer default styling to be more neutral and less alarmist.
+  * Principled explanation: if one diverts where various streams go, `cat()` follows printed output, whereas `message()` goes to standard error.
```
jimhester: FWIW I don't agree with this principal, I think these status messages _should_ be using `message()` (and stderr) and assuming RStudio / the Windows console did not re-color them I would push harder to make them use `message()`.

If usethis was a normal unix program I would expect these user facing messages to be on `stderr` as they would be with `message()`.

R/style.R
```diff
@@ -1,7 +1,13 @@
 # Helpers --------------------------------------------------------------------
 
 ## anticipates usage where the `...` bits make up one line
+##
+## 'usethis.quiet' is an undocumented option; anticipated usage:
+##   * eliminate `capture_output()` calls in usethis tests
+##   * other packages, e.g., devtools can call usethis functions quietly
 cat_line <- function(...) {
+  quiet <- getOption("usethis.quiet", default = FALSE)
```
jimhester: Wonder if it would be worth promoting this to an argument (with the same default value), so you could call `cat_line(quiet = FALSE)`. e.g.

```r
cat_line <- function(..., quiet = getOption("usethis.quiet", default = FALSE)) {
```

Maybe only useful for internal testing in usethis, but I think it is a little more natural.

principles.md
jennybc: Yeah, you are not wrong. I just wrote this down because Irene had asked (she's using usethis in the package for Tidies of March) and I decide to record the rationale, such as it is. This would not be hard to change, since everything is so centralized.

R/style.R
jennybc: For this to be productive internally, I'd need to add a `quiet` argument to `todo()` and `done()` and then this would happen in a few places:

Before:
```
if (!quiet) {
    done("Adding {collapse(value(new))} to {value(proj_rel_path(path))}")
}
```

After:
```
done("Adding {collapse(value(new))} to {value(proj_rel_path(path))}", quiet = quiet)
```

To be useful externally, I'd need to add `quiet` to user-facing `use_*()` functions.

But I see how that is a nicer definition of `cat_line()` itself, even if I don't do the other bits.

jennybc: And ... done.

jennybc: OK I did that, but one downside is that a future usethis developer could make a direct call to `cat_line()` and use `quiet = FALSE` argument and the "usethis.quiet = TRUE" option won't have the promised effect of silence. But presumably they'd either suffer immediately in tests or a downstream maintainer would raise a concern.

PR #427, opened 2018-08-07
Tidy default description

### LINKED ISSUE #396: Ensure default DESCRIPTION matches use_tidy_description()

jennybc:
Fixes #396 Ensure default DESCRIPTION matches use_tidy_description()

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -46,7 +46,11 @@ use_description <- function(fields = NULL) {
   fields[["Package"]] <- name
 
   desc <- build_description(fields)
-  write_over(proj_path("DESCRIPTION"), desc)
+  desc <- desc::description$new(text = desc)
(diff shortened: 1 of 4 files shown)


R/description.R
```diff
@@ -102,3 +106,22 @@ check_package_name <- function(name) {
 valid_name <- function(x) {
   grepl("^[[:alpha:]][[:alnum:].]+$", x) && !grepl("\\.$", x)
 }
+
+tidy_desc <- function(desc) {
+  stopifnot(inherits(desc, "description"), inherits(desc, "R6"))
```
jennybc: Is this how I check that I have an R6 `description` object?

hadley: Given that this is an internal function, I think you should assume `desc` is already the correct type.

PR #441, opened 2018-08-19
Add with_project() and local_project()

jennybc:
Add `with_project()` and `local_project()` (and make some improvements to `proj_get()` / `proj_set()` that also make this easier.)

Remove the `quiet` argument from `proj_get()` / `proj_set()`. All user-facing messages go through `cat_line()` and as of b0df5f1ca90d68fa72a9d337972d361792e94749, it consults the `usethis.quiet` option, defaulting to `FALSE`. Seems like progress for `proj_get()` / `proj_set()` to work the way everything else does.

I have written *many* versions of `proj_set()` at this point. How I ended up here:

  * Must support setting `path = NULL`. Comes up when restoring the state of "no active usethis project".
  * Must run `path` through `proj_path_prep()`, preferably just once, and prior to any user-facing messages.
  * Keep these concerns separate: `NULL`-ness of `path`, `force` (and `quiet`).


R/proj.R
```diff
@@ -1,5 +1,13 @@
 proj <- new.env(parent = emptyenv())
 
+proj_get_ <- function() proj$cur
+
+proj_set_ <- function(path) {
+  old <- proj$cur
+  proj$cur <- path
+  invisible(old)
+}
+
```
jennybc: These are now the only 2 functions that handle `proj$cur` directly.

R/proj.R
```diff
@@ -31,19 +39,26 @@ proj <- new.env(parent = emptyenv())
...
 NULL
 
-#' @describeIn proj_utils Retrieves the active project and, if necessary, attempts to set it in the first place.
-#' @param quiet Logical. Whether to announce project activation.
+#' @describeIn proj_utils Retrieves the active project and, if necessary,
+#'   attempts to set it in the first place.
+#' @param quiet Logical. Whether to suppress messages about project operations.
 #' @export
-proj_get <- function(quiet = FALSE) {
+proj_get <- function(quiet = getOption("usethis.quiet", default = FALSE)) {
```
jennybc: Now that this option is a thing, might as well default to it here.

NAMESPACE
```diff
@@ -18,6 +18,7 @@ export(edit_r_environ)
 export(edit_r_makevars)
 export(edit_r_profile)
 export(edit_rstudio_snippets)
+export(local_project)
```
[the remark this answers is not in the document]
jimhester: I think `with_project()` is good

R/proj.R
```diff
@@ -54,40 +68,66 @@ proj_get <- function(quiet = FALSE) {
...
+#'   temporary active project.
+#' @export
+with_project <- function(path = ".",
+                         code,
+                         force = FALSE,
+                         quiet = getOption("usethis.quiet", default = FALSE)) {
+  old_quiet <- options(usethis.quiet = quiet)
+  old_proj  <- proj_set(path = path, force = force)
+
+  on.exit(proj_set(path = old_proj, force = TRUE))
```
jimhester: I maybe would just combine these together into one `on.exit()` call.

```r
on.exit({
  proj_set(path = old_proj, force = TRUE)
  options(old_quiet)
})
```

R/proj.R
```diff
@@ -54,40 +68,66 @@ proj_get <- function(quiet = FALSE) {
...
+local_project <- function(path = ".",
+                          force = FALSE,
+                          quiet = getOption("usethis.quiet", default = FALSE)) {
+  old_quiet <- options(usethis.quiet = quiet)
+  old_proj  <- proj_set(path = path, force = force)
+
+  withr::defer({
+    proj_set(path = old_proj, force = TRUE)
+    options(old_quiet)
+  }, envir = parent.frame())
```
jimhester: Generally I think you need to expose the `envir` parameter for local functions, there are some cases where `parent.frame()` is not actually the environment you might want.

R/proj.R
jennybc: Yeah, that also makes the `with_` and `local_` code more parallel. Done.

R/proj.R
jennybc: OK I did this exactly as I see in all other `withr::local_*()` functions.

Issue #445, opened 2018-08-26
Improve the file created by `use_vignette()`


rorynolan:
Ordinarily, the output of `use_vignette("Cool thing")` would be a file with the following YAML header:

```
---
title: "Vignette Title"
author: "Vignette Author"
date: "`r Sys.Date()`"
output: rmarkdown::html_vignette
vignette: >
  %\VignetteIndexEntry{Vignette Title}
  %\VignetteEngine{knitr::rmarkdown}
  %\VignetteEncoding{UTF-8}
---
```

whereas it would be great if the vignette title fields and Author fields were filled in too such that it would be 

```
---
title: "Cool thing"
author: "Package authors (from DESCRIPTION file)"
date: "`r Sys.Date()`"
output: rmarkdown::html_vignette
vignette: >
  %\VignetteIndexEntry{Cool thing}
  %\VignetteEngine{knitr::rmarkdown}
  %\VignetteEncoding{UTF-8}
---
```

I also think it would be good to replace the space in the file name, such that it's `Cool-thing.Rmd`.

If you agree, I'm happy to implement some or all of this myself with a pull request.

Thanks, I use `usethis` everyday.
Rory

jennybc:
I like the idea re: populating title (and in the same way in both places 🙄). I sure wish the duplication wasn't necessary.

In theory, the tidyverse/r-lib "default house style" is to not have package vignette authors, at least if it just repeats package authors. But, I grant, it's easier to delete a field than add it! Pre-populating with something from DESCRIPTION sounds interesting, but I'm also suspicious the code necessary to do this, in full generality, is worth the maintenance. If you try it, I'd seriously consider using just maintainer, to guarantee it is ONE name. If that's the wrong name, it is easy enough to edit by hand.

I believe the vignette file name *is* already "asciified"?

I suspect this comes down to whether populating the title is worth departing from the very simple approach we currently have, which just delegates to `rmarkdown::draft()`.

It's great if you want to explore it, but I think it will be a hard sell if it can't stay very simple and easy to maintain.

rorynolan:
Alrighty. It seems that putting in the authors automatically isn't a great idea, wasn't completely sold on it myself anyway. Could maybe change it to `INSERT AUTHOR HERE OR DELETE THIS LINE`  to make it more obvious that that needs to be changed? No biggie either way.

[...]

PR #446, opened 2018-08-27
Fill vignette titles in YAML with `use_vignette()`.

rorynolan:
Hi Jenny,
Here's the PR from https://github.com/r-lib/usethis/issues/445.
I'm pretty sure it works, I've tested it manually. Before I write a unit test for it, I'd like you to advise me how to. My instinct is to put a file in `inst/for-testthat` (or something like that) that should be identical to an output of `use_vignette("Cool thing")`, but that's probably not the way you do things.
Rory


jennybc:
I'm working on something else now FYI, not ignoring you. usethis just had a release, so it will lie fallow for a bit.

rorynolan:
No worries, I realise this would be a low-priority issue even if you were working on `usethis`. I might tinker around a bit with this PR, but you can safely ignore me until you're ready to deal with this.

R/vignette.R
```diff
@@ -34,9 +34,13 @@ use_vignette <- function(name) {
   path <- proj_path("vignettes", asciify(name), ext = "Rmd")
 
   done("Creating {value(proj_rel_path(path))}")
-  rmarkdown::draft(
-    path, "html_vignette", "rmarkdown",
-    create_dir = FALSE, edit = FALSE
+  use_template("vignette.Rmd",
+    save_as = proj_rel_path(path),
+    data = list(
+      vignette_title = name,
```
hadley: I think it would be better to add a new argument, `title`.

inst/templates/vignette.Rmd
```diff
@@ -0,0 +1,65 @@
+---
+title: "{{{ vignette_title }}}"
+author: "{{{ vignette_author }}}"
```
hadley: I think we should just remove the `author` and `date` fields altogether. I don't think they're very useful.

hadley:
@rorynolan are you interested in finishing this off? No problems if not, just let me know so I can get it over the finish line.

rorynolan:
Sure @hadley,
Ill do this tomorrow. 
Rory

rorynolan:
OK @hadley I think this is good to go.

rorynolan:
Alrighty @hadley, I've done that now :-)

hadley:
Thanks @rorynolan! 🎉

PR #459, opened 2018-09-11
Improve alt text for badges

hadley:
A couple of minor tweaks that should make badges slightly more useful for visually impaired readers


jennybc:
Good idea.

PR #464, opened 2018-09-18
Exit gracefully when calling use_version() and then selecting "0" - fixes #463

### LINKED ISSUE #463: Calling use_version() and then selecting "0" leads to error
In the current `usethis` version, calling `use_version()` and then selecting "0" to exit without changing anything leads to an error instead than exiting gracefully: 

``` r
library(usethis)
use_version()
#> Current version is 1.4.0.9000.
Which part to increment? 

1: major --> 2.0.0
2: minor --> 1.5.0
3: patch --> 1.4.1
4: dev --> 1.4.0.9001

Selection: 0
Error in match.arg(which, c("major", "minor", "patch", "dev")) : 
  'arg' should be one of “major”, “minor”, “patch”, “dev”
```

Though this is a small issue, I made a PR which would fix this and exit "gracefully" instead, leading to this behaviour: 

```  r
> library(usethis)
> use_version()
✔ Setting active project to '/home/lb/Source/git/usethis'
Current version is 1.4.0.9000.
Which part to increment? (0 to Exit) 

1: major --> 2.0.0
2: minor --> 1.5.0
3: patch --> 1.4.1
4: dev --> 1.4.0.9001

Selection: 0
● You selected to quit! No changes will be made.
```

lbusett:
This would fix #463, leading to graceful exit if the user selects "0" after calling `use_version()` without arguments.

### DIFF
--- a/R/version.R
+++ b/R/version.R
@@ -37,7 +37,7 @@ use_version <- function(which = NULL) {
 
   ver <- desc::desc_get_version(proj_get())
 
-  if(is.null(which) && !interactive()) {
+  if (is.null(which) && !interactive()) {
     return(invisible(ver))
   }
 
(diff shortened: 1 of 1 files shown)


R/version.R
```diff
@@ -46,20 +46,26 @@ use_version <- function(which = NULL) {
   if (is.null(which)) {
     choice <- utils::menu(
       choices = glue("{names(versions)} --> {versions}"),
```
jennybc: Will you make this change while you're here?

```
format(names(versions), justify = 'right')
```

R/version.R
```diff
@@ -46,20 +46,26 @@ use_version <- function(which = NULL) {
...
-      title = glue("Current version is {ver}.\n", "Which part to increment?")
+      title = glue("Current version is {ver}.\n", "Which part to increment?",
+                   " (0 to Exit)")
```
jennybc: I think you should check for 0 here and simply return an invisible `FALSE` immediately in that case. Also please make sure to use the styler "Style selection" addin on any areas you edit.

lbusett: Of course, you are right... much simpler. Do not know what I was thinking....

Concerning the `return`, would you prefer to keep the message before returning such as in: 

``` r
todo("You selected to quit! No changes will be made.") # or something similar
return(invisible(FALSE))
```

or jest return directly without messaging?

jennybc: I think it's ok to have no message here.

lbusett: ok, done. Also corrected styling and introduced right-alignment in names. Hope now it's ok.

jennybc:
Thanks!

PR #467, opened 2018-09-26
Make use_pkgdown configurable

### LINKED ISSUE #392: Make use_pkgdown configurable?
Would you take a PR that implements configurable arguments to `use_pkgdown()`? I'd like to use usethis in pkgdown but these values change across different pkgdown usage scenarios.

Something like this:

```r
#' Use pkgdown
#'
#' [pkgdown](https://github.com/hadley/pkgdown) makes it easy to turn your
#' package into a beautiful website. This helper creates `_pkgdown.yml`
#' and `docs/` for you, and adds to `.Rbuildignore`
#'
#' @export
use_pkgdown <- function(pkg_file = "_pkgdown.yml", pkg_dir = "docs") {
  check_is_package("use_pkgdown()")
  edit_file(proj_path(pkg_file))
  use_build_ignore(pkg_file)

  use_directory(pkg_dir, ignore = TRUE)
}
```

jayhesselberth:
Closes #392

### DIFF
--- a/R/pkgdown.R
+++ b/R/pkgdown.R
@@ -5,13 +5,13 @@
 #' and `docs/` for you, and adds them to `.Rbuildignore`
 #'
 #' @param config_file pkgdown yaml config file
-#' @param build_dir target directory for pkgdown docs
+#' @param dst_path target directory for pkgdown docs
 #'
 #' @export
-use_pkgdown <- function(config_file = "_pkgdown.yml", build_dir = "docs") {
+use_pkgdown <- function(config_file = "_pkgdown.yml", dst_path = "docs") {
   check_is_package("use_pkgdown()")
   edit_file(proj_path(config_file))
   use_build_ignore(config_file)
 
-  use_directory(build_dir, ignore = TRUE)
+  use_directory(dst_path, ignore = TRUE)
 }
--- a/R/pkgdown.R
+++ b/R/pkgdown.R
@@ -2,7 +2,7 @@
 #'
 #' [pkgdown](https://github.com/r-lib/pkgdown) makes it easy to turn your
 #' package into a beautiful website. This helper creates a pkgdown config file
-#' and a directory for pkgdown docs, and adds them to `.Rbuildignore`
+#' and adds the file and destination diretory to `.Rbuildignore`
 #'
 #' @seealso http://pkgdown.r-lib.org/articles/pkgdown.html#configuration
 #'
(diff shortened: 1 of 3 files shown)


jayhesselberth:
`build_dir` could also be `dst_path`, which is what pkgdown uses.

R/pkgdown.R
```diff
@@ -1,14 +1,19 @@
...
+#' and a directory for pkgdown docs, and adds them to `.Rbuildignore`
+#'
+#' @seealso http://pkgdown.r-lib.org/articles/pkgdown.html#configuration
+#'
+#' @param config_file pkgdown yaml config file
+#' @param dst_path target directory for pkgdown docs
 #'
 #' @export
-use_pkgdown <- function() {
+use_pkgdown <- function(config_file = "_pkgdown.yml", dst_path = "docs") {
```
hadley: If the user sets `dst_path` should it be reflected in the generated `pkgdown.yml` file?

R/pkgdown.R
```diff
@@ -1,14 +1,19 @@
...
   check_is_package("use_pkgdown()")
-  edit_file(proj_path("_pkgdown.yml"))
-  use_build_ignore("_pkgdown.yml")
+  edit_file(proj_path(config_file))
+  use_build_ignore(config_file)
 
-  use_directory("docs", ignore = TRUE)
+  use_directory(dst_path, ignore = TRUE)
```
hadley: I think we should probably build ignore the docs directory, but not create it. That would also fix #495

R/pkgdown.R
jayhesselberth: Yeah pkgdown takes:

```yaml
destination: docs
```

So we'd need to write that line into the created `_pkgdown.yml` file.

jayhesselberth: What's the recommended way to modify this file? Something like:

```r
cat(paste("destination:", dst_dir), file = config_file)
```

jennybc: Although it means something different, other functions in usethis this have a `destdir` argument FWIW (`create_*()`, `use_course()`).

hadley: @jayhesselberth  I think we can assume that this function creates the file, so you can use `write_over()`

And make the argument either `destination` (to match pkgdown) or `destdir` (to match usethis). (I don't have strong feelings which)

hadley:
Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

(And looks like the docs need an update)

hadley:
Thanks!

PR #411, opened 2018-07-15
Integration of 'travis' and `tic` functionality

pat-s:
https://github.com/ropenscilabs/travis/issues/73

Travis: https://github.com/ropenscilabs/travis
tic: https://github.com/ropenscilabs/tic

Travis builds: https://travis-ci.org/pat-s/usethis

### Open tasks

- [ ] Look in sysreqdb for additional dependencies, suggest "addons: apt: ..." entry

- [ ] Write test: How to handle the interactive prompts in tests?
- [x] Export all unexported needed functions from Travis to be able to import them here

### Implementation

New exported functions: `use_ci()`, `use_travis_deploy()`

Main function: `use_ci`:
- Creates YAML templates for `travis` and `appveyor`
- Uses `use_travis_deploy()` to setup deploy access on Travis via ssh
- Adds a Github PAT to the Github repo
- Works interactively so the user can also deny some steps if only partial functionality is needed

Added tic.R templates for different kind of repos under `inst/templates/.

PR #476, opened 2018-10-03
New use_travis_deploy()

krlmlr:
Extracted from #411.


jennybc:
I'm readying for release and am trying to get our PRs / issues under control for the first time in a while.

I'm going to close this since the travis and tic packages are not on CRAN and recent events at Travis-CI are cause for some concern about its future.

We can always re-open this when either of those pieces of context change.

PR #486, opened 2018-10-29
Allow symbolic links to data directories

nijibabulu:
The situation occurs with large data sets that there may be multiple versions of an analysis in multiple data directories that the user may want to maintain. One solution is to use symbolic links from `data` to the current analysis The current version of usethis prevents this, but the attached patch would allow it.

### DIFF
--- a/R/directory.R
+++ b/R/directory.R
@@ -26,12 +26,16 @@ use_directory <- function(path,
   invisible(TRUE)
 }
 
+is_traversable <- function(path) {
+  is_dir(path) || (is_link(path) && is_dir(Sys.readlink(path)))
+}
+
 create_directory <- function(base_path, path) {
   if (!file_exists(base_path)) {
     stop_glue("{value(base_path)} does not exist.")
   }
 
-  if (!is_dir(base_path)) {
+  if (!is_traversable(base_path)) {
     stop_glue("{value(base_path)} is not a directory.")
   }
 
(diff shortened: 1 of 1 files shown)


hadley:
@jimhester is there a better way to do this with fs? It surprises me that `is_dir()` is not true for a link to a directory.

jennybc:
@nijibabulu Can you elaborate on how you use usethis where this comes up? Specifically, which function are you calling that then does not work because of a symbolic link?

jennybc:
@nijibabulu Here's the motivation for my question: I don't think we would want symbolic links to be tolerated by functions such as `use_data()`. So I'm interested to now what function you are using where the symbolic links come up **and** we would want to support that.

nijibabulu:
@jennybc I am using this indeed with `use_data()`, where it would fail. The motivation behind this is that for some project, I would use the `data` directory for storing my large data sets, or versions of the large data sets that change over the course of tinkering (e.g. `data.v1`, `data.v2` `data -> data.v2`). Having the ability to unlink and relink to different versions of the data are convenient for me.

R/directory.R
```diff
@@ -26,12 +26,16 @@ use_directory <- function(path,
   invisible(TRUE)
 }
 
+is_traversable <- function(path) {
+  is_dir(path) || (is_link(path) && is_dir(Sys.readlink(path)))
```
jimhester:
```suggestion
  is_dir(path) || (is_link(path) && is_dir(link_path(path))
```

hadley:
Fixed upstream in fs

PR #493, opened 2018-11-13
Broken with \n in title

muschellij2:
It looks as though https://github.com/r-lib/usethis/commit/65764fa364e5cc3b1ac113eedcb52312bf00b0d4#diff-4ac879304421b0c163d57d6d4be6b43dL89 changed the `gsub` part of the `pkg$Title` for `repo_desc` in `use_github`, which causes `gh::gh` to fail if there is a newline `\n`.  

This should revert that behavior.


R/git.R
```diff
@@ -28,6 +28,9 @@ use_git <- function(message = "Initial commit") {
     if (yep(commit_consent_msg)) {
       done("Adding files and committing")
       git2r::add(r, paths)
+      if (length(paths) > 0) {
```
jennybc: I feel `git_uncommitted()` (in the condition for this block) implies that `paths` has non-zero length. Have you seen something to the contrary?

muschellij2: This PR part was for https://github.com/r-lib/usethis/issues/128

jennybc: If there are two different issues, let's split into two PRs then.

muschellij2: LIttle swamped now, will within next week or 2.

jennybc: OK thanks!

hadley:
Closing in favour of two smaller PRs

PR #494, opened 2018-11-13
Add option usethis.protocol to change default globally 

### LINKED ISSUE #492: FR: Add option to change default protocol for git
I use mostly `https` and not `ssh` for a github stuff. I would find it useful to be able to change the default `protocol` argument in function like `usethis::create_from_github()`. An option to add to my .Rprofile would be awesome, like `pkgdown.protocol`. 

Currently, I must add `protocol = "https"` to each function that use `ssh` by default, and most often... i forget it so I have an error, then I put the argument. 

What do you think ? 

I am willing to do a PR if you find this interesting to have. 

Thanks for considering.

cderv:
This is an implementation that closes #492 feature request. 

* Keep the "ssh" the default behaviour
* Continue to match `protocol` to `"https"` or `"ssh"`

There are no tests about protocol in https, only manual test with ssh protocol. I could add some for protocol https but I won't be running the manual tests so let me know what you wish on this. 

About `match.arg`, I wanted to use `rlang::arg_match` but it is not use yet so I comply with the current standard for _usethis_. I can do another PR about migrating to `rlang::arg_match`. Error message would be clearer and there would be no partial matching, which is interesting here for `protocol` argument.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis *development version*
 
+* `protocol` argument is now an option in `use_github()` and `create_from_github()`. The default is still to use `"ssh"` protocol but it can be changed globally to `"https"` with `options(usethis.protocol = "https")`. (#494, @cderv)
+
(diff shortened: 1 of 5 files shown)


jennybc:
LGMT, thanks!

PR #501, opened 2018-11-24
Add use_makefile() and use_jenkinsfile().

ryapric:
Added templates & calling functions to place Makefiles and Jenkinsfiles at project root.

R CMD check results, which I ran using the `--as-cran` flag:

```
── R CMD check results ───────────────────────────────── usethis 1.4.0.9000 ────
Duration: 33.6s

0 errors ✔ | 0 warnings ✔ | 0 notes ✔

R CMD check succeeded
```

Thanks!


ryapric:
I've updated with your suggestions, and their implied changes.

- Functions are now `use_<make:jenkins>()`

- Makefile uses `devtools::install()` as part of the default targets (first checking if `devtools` itself is installed), which handles dependencies, etc. gracefully. A second `make` install target, `install_src`, instead calls `R CMD INSTALL` for those that would rather do it that way.

Note that the AppVeyor build on Windows seems to have failed because of comparing string literal paths (failed because of path separator discrepancy), instead of resolving the path separators. I am not certain if this is because of the new tests I added, because the failures are thrown via calls to the helper functions:

```
  1. Failure: edit_r_XXX('user') ensures the file exists (@test-edit.R#40) 
  2. Failure: edit_r_profile() respects R_PROFILE_USER (@test-edit.R#56)
```

All tests, build steps, etc. pass on my end (Linux), as do the Travis builds.

hadley:
Can you please merge the latest from master? That should fix the appveyor issues.

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
+R = R --vanilla CMD
+
+
+all: build check install clean
+
+build:
+	$(R) build .
+
+check: build
+	$(R) check --no-manual *.tar.gz
```
jimhester: Rather than relying on local directory and `*.tar.gz` (which could match non-package tarballs) maybe take a page from the knitr makefiles

https://github.com/yihui/knitr/blob/dc5ead7bcfc0ebd2789fe99c527c7d91afb3de4a/Makefile#L1-L4

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
+R = R --vanilla CMD
```
jimhester: I don't really think this is necessary, I would drop this and just use `R CMD` directly in the build steps.

Also are we trying to make this work on Windows? or not?

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+
+all: build check install clean
```
jimhester: I don't really think install should be in `all`, maybe just `check` (which implies `build`)?

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
...
+		tryCatch( \
+			invisible(packageVersion('devtools')), \
+			error = function(e) { \
+				install.packages('devtools', repos = 'https://cloud.r-project.org') \
+			} \
+		)"
+endef
+install:
+	@$(devtools_check); \
+	Rscript -e 'devtools::install()'
```
jimhester: Rather than using devtools maybe use remotes, break this into a separate rule and install from the built tarball.

```make
install_deps:
	Rscript -e 'if (!requireNamespace("remotes") install.packages("remotes")' -e 'remotes::install_deps(dependencies = TRUE)'

install: install_deps build
	R CMD INSTALL $(PKGNAME)_$(PKGVERS).tar.gz
```

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
...
+
+install_src:
+	$(R) INSTALL .
+
+clean:
+	rm -rf *.tar.gz *.Rcheck ..Rcheck
```
jimhester:
```suggestion
	rm -r $(PKGNAME).tar.gz $(PKGNAME.Rcheck)
```

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
...
+			error = function(e) { \
+				install.packages('devtools', repos = 'https://cloud.r-project.org') \
+			} \
+		)"
+endef
+install:
+	@$(devtools_check); \
+	Rscript -e 'devtools::install()'
+
+install_src:
```
jimhester: I think this rule can just be dropped entirely.

inst/templates/Makefile
```diff
@@ -0,0 +1,29 @@
+R = R --vanilla CMD
+
+
+all: build check install clean
+
+build:
+	$(R) build .
```
jimhester: I guess we should think about if it would be better to build the tarball in the current working directory or the parent... I don't have a strong preference either way...

jennybc:
@ryapric Can you give me a high-level overview of the objective here? I get that this creates a Makefile but to what purpose? What is the workflow facilitated by the added files?

Context: I'm readying for release and am trying to get our PRs / issues under control for the first time in a while. I'm triaging into "for current release", "backlog", "will not pursue at this time".

ryapric:
Hi @jennybc, do I win the award for longest-stale PR yet?

To address your question, adding a well-written Makefile to someone's R package workflow makes their build/test/etc. steps move entirely with their code. RStudio can't always be around to facilitate those steps, and not everyone uses Travis, etc. where the steps are already configured. Exposing targets like `make test` to any build system of the developer's choosing (e.g. Jenkins, etc.) facilitates portable, readable CI/CD.

I'd like to take an hour or so to address the great feedback that @jimhester and @hadley offered above, but if you want to keep this under the Backlog milestone, that's fine.

hadley:
@ryapric you PR has to be at least 3 years old to even be in the running for that award 🤣

inst/templates/Makefile
ryapric: I rarely write anything on Windows these days, personally, and I've never written `make` recipes that don't specifically rely on `sh` or `bash`. If someone has a quick `ifeq()` block with support, I'm all ears, but I'm fine stuffing Windows-specific support for this particular PR into the backlog.

ryapric: And regarding your first comment: not passing `--vanilla` will fail if it tries to run and some tools aren't installed, right? Like `pandoc` for building the manual, etc.? Or did you just mean don't add a Makevar for the whole command `R --vanilla CMD`?

jennybc:
I'm neutral on this and defer to @jimhester and @Hadley.

If they're on board, I'm happy to rescue from the backlog if you want to finish this off.

inst/templates/Makefile
ryapric: Hmmm, good point. I feel that if it's built in the current directory (hopefully package top-level), then that would be easier to control for by the end-user (via .gitignore etc), and more portable besides. If a user stores all their package code in folders where the parent of each is write-restricted or something else weird, then the build step would fail. Enforcing focus on the current-and-below path should be much more manageable, by everyone. I'm happy to hear thoughts, though.

inst/templates/Makefile
ryapric: Edit: I've ran some ad-hoc stuff, I might have gotten ahead of myself. Updated.

ryapric:
Cool, entirely up to you all, then. I've pushed all the suggested changes, CI builds are running at the time of this posting. Waiting for a comment reply above from @jimhester, but he said he doesn't have strong opinions on the matter (which directory the build step is in).

jimhester:
I think the Makefile part of this looks good now

R/jenkins.R
```diff
@@ -0,0 +1,19 @@
+#' Create Jenkinsfile for Jenkins CI Pipelines
+#'
+#' `use_jenkins()` adds a basic Jenkinsfile for R packages to the project root
+#' directory. The Jenkinsfile stages take advantage of calls to `make`, and so
+#' calling this function will also run `use_make()` if a Makefile does not
+#' already exist at the project root.
+#'
+#' @seealso The [documentation on Jekins
```
jennybc:
```suggestion
#' @seealso The [documentation on Jenkins
```

R/jenkins.R
```diff
@@ -0,0 +1,19 @@
...
+#'   Pipelines](https://jenkins.io/doc/book/pipeline/jenkinsfile/).
```
jennybc:
```suggestion
#'   Pipelines](https://jenkins.io/doc/book/pipeline/jenkinsfile).
```

ryapric:
Great catch, thank you!

jennybc:
You're going to kill me but I just realized we still need a NEWS bullet, since this adds 2 new exported functions.

Can you please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

ryapric:
Sure, do you want it under the current 1.5.0 heading, or do you want me to bump it to a dev version since 1.5.0 already went out?

jennybc:
There should be a dev heading but I guess there's not. Yes please add such a heading and I'll sort it out from there, after merge.

jennybc:
Nevermind. There is such a heading and you need to merge in master.

ryapric:
I see it now, sorry! Pushed.

PR #503, opened 2018-11-26
Constant title for dev news

### LINKED ISSUE #440: Implement "Devel news should have constant title"
Implement https://github.com/tidyverse/style/issues/100 "Devel news should have constant title"

hadley:
Fixes #440


jennybc:
Looks good.

FWIW this is another case of an autocommit happening (I saw another discussion of that). Part of me thinks this should be left undone or the user should be asked to confirm, the way we do now in `use_git()`. That is a larger, policy-type discussion.

hadley:
Yeah, agreed.

PR #508, opened 2018-11-26
Initial implementation of pr_use

hadley:


R/pr.R
```diff
@@ -0,0 +1,137 @@
+pr_create <- function(branch) {
```
jennybc: At first I though `pr_create()` would create a PR. But it really makes sure you are working on a branch, probably off of `master`. Maybe `pr_init()` or `pr_start()`?

R/pr.R
```diff
@@ -0,0 +1,137 @@
...
+  done("Pulling changes from GitHub PR")
+  git2r::pull(repo)
+}
+
+pr_update <- function() {
+  done("Fetching master branch from GitHub")
+  git2r::fetch(repo, "origin", refspec = "refs/head/master")
+
+  done("Merging master branch")
+  out <- merge(repo, "origin/master")
```
jennybc: I think I rebase as or more often as I merge. I guess it doesn't matter that much given our "squash and merge" mentality.

R/pr.R
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  pr_push()
+
+  invisible(TRUE)
+}
+
+pr_view <- function() {
+  view_url(pr_url())
+}
```
jennybc: 🎉🙌

R/pr.R
```diff
@@ -0,0 +1,137 @@
...
+    url
+  }
+}
+
+# Helpers -----------------------------------------------------------------
+
+check_on_branch <- function() {
+  cur_branch <- proj_git_branch_get()
+  if (cur_branch$name == "master") {
+    stop("Currently on master branch. Do you need to call `pr_use()` first?", call. = FALSE)
```
jennybc: I don't see a `pr_use()` function; I suspect this is meant to be `pr_create()` or whatever that becomes.

R/pr.R
```diff
@@ -0,0 +1,137 @@
...
+  }
+}
+
+proj_git_repo <- function() {
+  check_uses_git()
+  git2r::repository(proj_path())
+}
```
jennybc: If we make this, there are several other places we should use it.

R/pr.R
```diff
@@ -0,0 +1,137 @@
...
+
+  cur_branch <- proj_git_branch_get()
+  if (cur_branch$name == branch) {
+    return(invisible(TRUE))
+  }
+
+  pr_branch <- proj_git_branch_find(branch)
+  if (is.null(pr_branch)) {
+    if (cur_branch$name != "master") {
+      if (nope("Start pull-request from non-master branch?")) {
```
jennybc: Wording quibble: I feel like you do usually want to PR from a non-master branch. This is really about whether you want the PR branch to have a non-`master` branch as parent.

R/pr.R
hadley: I also dabbled with `pr_use()` because you can also use it to switch to an existing PR branch.  I like `pr_init()` too.

R/pr.R
hadley: Reasons why I merge here:

* I couldn't figure out how to rebase with git2r

* Rebase will mean that you have to force push, which makes me a little worried now that others might contributing to the PR branch (esp. via suggestions)

hadley:
@jimhester do you mind taking a look? If the basic strategy looks ok, I'll finish up the documentation and merge. I expect it'll need quite a bit of use and iteration before release.

R/pr.R
```diff
@@ -0,0 +1,271 @@
...
+  diff <- git_branch_compare(branch)
+
+  if (diff[[2]] == 0) {
+    return()
+  }
+
+  stop(
+    glue::glue(
+      "{branch} branch is out of date.
+      Please resolve (somehow) before continuing.
```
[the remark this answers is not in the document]
hadley: Good catch - there's also a `stop_glue()` function I can use

Issue #479, opened 2018-10-12
Have `usethis::proj_path_prep()` use `fs::path_abs(fs::path_expand())` instead of `fs::path_real()`


jimhester:
usethis currently uses `fs::path_real()` to determine the project path, however this poses problems when the path is a mapped network drive on Windows, because the root no longer can be calculated relative to the input path.

I map my R projects folder to my windows VM, so I run into this situation often.

I propose to instead use `fs::path_abs(fs::path_expand())`.

``` r
usethis::proj_set("Z:/projects/remotes")
#>  Setting active project to '//vmware-host/Shared Folders/projects/remotes'
usethis::use_test()
#> Error: Open file must be in the 'R/' directory of the active package.
#> * Actual path: 'Z:/projects/remotes/R/submodule.R'
fs::path_real("Z:/projects/remotes/R/submodule.R")
#> //vmware-host/Shared Folders/projects/remotes/R/submodule.R
fs::path_abs(fs::path_expand("Z:/projects/R/submodule.R"))
#> Z:/projects/R/submodule.R

fs::path_real("~/Links")
#> C:/Users/IEUser/Links
fs::path_abs(fs::path_expand("~/Links"))
#> C:/Users/IEUser/Links
```

Created on 2018-10-12 by the [reprex package](http://reprex.tidyverse.org) (v0.2.0).

jennybc:
When I first did this, I did that survey of the fs path expanding/normalizing/realizing functions and how they overlap with each other (or do not).

I concluded that `fs::path_real()` was the best plan, but my confidence was far from 100%. I will look back at the commits to see if I recorded anything interesting.

Barring some discovery that we must discuss ... sure, I'm happy to make this switch.

jennybc:
Seems to be the most important commit/comment: https://github.com/r-lib/usethis/pull/385/commits/29397a677d01bfa93126979b93fb83eef5390b26#r195969953

I see no specific reason to prefer `fs::path_real()` (current) over `fs::path_abs(fs::path_expand())` (your proposal).

jennybc:
See #485 why this isn't quite simple. I am caught between path handling in rprojrooot (definitely uses `normalizePath()`) and rstudioapi (not sure what I can count on).

What if we leave `proj_path_prep()` as is and, instead, use it instead of `path_expand_r()` here?

https://github.com/r-lib/usethis/blob/44dd72c9917bcc7c7189ae25db53ef6dd377a7f9/R/r.R#L42

[...]

PR #511, opened 2018-11-27
New policy for create functions

### LINKED ISSUE #453: create_from_github() doesn't reset active project properly

hadley:
Create functions now preserve the active project. This shouldn't affect people using RStudio, since a new session will be started automatically, but I think this is easier to reason about

Fixes #453 — this ended up being a bigger change than I expected, but I think it's the right thing to do. I know the previous behaviour often ended up confusing me when teaching.


jennybc:
I took this out for a spin, hence the subsequent commits.

I'm not totally convinced re: not activating the new project, but I can live with it. It raises two questions for me:

  * Should we return the path to the newly created thing from `create_()` functions? This captures the path *after* applying our path-handling policies and would make various things easier, both in terms of usage and tests.
  * If we aren't going to open the newly created thing in a new RStudio session and it's interactive, should we *ask* if they user would like to switch the active project and working directory? Right now people have to remember to switch *both* of these manually.

hadley:
Yes, we should definitely return the path.

I'm not sure about asking. Changing the working directory just seems like such an odd thing to do. Shall we discuss on Wednesday?

jennybc:
Maybe there should be a `proj_switch(path)` function or similar that activates project at `path` and sets working directory to that project. Right now, in a non-RStudio workflow, right after `create_()`, executing `project_set()` alone leaves things in a confusing place: old working directory but usethis is targeting the new project. This awkward situation is what we're showing in the README of this PR.

R/create.R
```diff
@@ -39,21 +40,18 @@ create_package <- function(path,
...
-    use_rstudio()
-  }
-
-  if (open) {
-    open_project(proj_get(), restore = old_project)
+    use_rstudio(open = open)
   }
 
-  invisible(TRUE)
+  invisible(proj_get())
```
hadley:
```suggestion
  invisible(path)
```

I think that's easier to reason about

jennybc: If we do this, we'll also need to re-process with `proj_path_prep()`, i.e. return `invisible(proj_path_prep(path))`. The reason for using what `proj_get()` returns is that it guarantees consistent path handling (realization, expansion, normalization, etc.). This, in turn, is necessary for other path comparison things to work, e.g. try to prevent more headaches like #479.

jennybc:
We could also allow `open = TRUE` to mean "switch to new project", where that means launch in new RStudio session, if in RStudio, or change active project & working directory in current session. 

But then we've almost gone around in a giant circle and come back to where we started 😕

PR #515, opened 2018-11-28
Allow user to declare badge block

hadley:
Looks for  and , which  been added to the default templates. 

@maelle could you also please take a look? I decided to mark the start and the end of the badge block in order to make parsing more robust.


R/badge.R
```diff
@@ -2,7 +2,10 @@
 #'
 #' These helpers produce the markdown text you need in your README to include
 #' badges that report information, such as the CRAN version or test coverage,
-#' and link out to relevant external resources.
+#' and link out to relevant external resources. To add badges auomatically
```
jennybc:
```suggestion
#' and link out to relevant external resources. To add badges automatically
```

R/badge.R
```diff
@@ -2,7 +2,10 @@
...
+#' ensure your badge block is starts with a line containing only
```
jennybc:
```suggestion
#' ensure your badge block starts with a line containing only
```

R/badge.R
```diff
@@ -41,18 +44,49 @@ NULL
...
+    }
+
+    todo(
+      "Add a {field(badge_name)} badge by adding the following line ",
+      "to your README:"
+    )
+    code_block(link)
+    todo(
+      "Allow badges to be added automatically by adding ",
+      "{code('<!-- badges -->')} to the first line of the badges paragraph"
```
jennybc: I think this fell behind the help & code.

tests/testthat/test-use-badge.R
```diff
@@ -21,8 +21,17 @@ test_that("use_binder_badge() the github repository works", {
...
 test_that("use_badge() does nothing if badge seems to pre-exist", {
-  pkg <- scoped_temporary_package()
+  scoped_temporary_package()
   href <- "https://cran.r-project.org/package=foo"
   writeLines(href, proj_path("README.md"))
   expect_false(use_badge("foo", href, "SRC"))
 })
+
+test_that("default readme has placeholders / can add to empty badge block", {
+  withr::local_options(list(usethis.quiet = FALSE))
```
jennybc: This is option is already set globally for tests in `tests/testthat/setup.R`.

R/badge.R
```diff
@@ -41,18 +44,49 @@ NULL
 #' @rdname badges
 #' @export
 use_badge <- function(badge_name, href, src) {
-  if (has_badge(href)) {
+  link <- glue("[![{badge_name}]({src})]({href})")
+  badge_lines <- badge_lines()
```
jennybc: It feels odd to get the line numbers back separate from the README file path and to then re-find the README. But I'm not sure what to do about that.

hadley: I couldn't figure out how to factor it out without making the fallback code (when there's no README, or no badge comments) more complicated. Someone can take another stab at it in #517

PR #516, opened 2018-11-28
Harden use_github() re: partial setup

### LINKED ISSUE #221: use_github(): be more resilient to partial setup
Right now it is possible for `use_github()` to successfully create a GitHub repo and set a remote. But then fail to push, e.g. due to SSH key/passphrase problems.

Then, when you call it again, it exits early, because it thinks the repo already uses GitHub, because the remote is set. If you unset that, you get a new (cryptic) error because the remote repo already exists.

Could it handle "partial setup" better?

hadley:
Now checks individual for origin remote (i.e. is there a GitHub repo), description fields, and push-ability for master.

Fixes #221


R/github-labels.R
```diff
@@ -22,8 +22,8 @@
 #' * `good first issue` indicates a good issue for first-time contributors.
 #' * `help wanted` indicates that a maintainer wants help on an issue.
 #'
-#' @param repo Optional repository specification (`owner/repo`) if you
-#'   don't want to use the current project.
+#' @param repo_spec Optional repository specification (`owner/repo`) if you
```
jennybc: For consistency with `create_from_github()`, `use_tidy_thanks()`, and lots of internal usage.

R/github-utils.R
```diff
@@ -1,24 +1,30 @@
...
+  m <- vapply(remotes, function(x) grepl("github", x), logical(1))
+  if (length(m) == 0) return(NULL)
+  remotes[m]
+}
+
+github_origin <- function(path = proj_get()) {
+  r <- github_remotes(path)
+  if (length(r) == 0) return(NULL)
+  parse_github_remotes(r)[["origin"]]
 }
```
jennybc: We started using `gh::gh_tree_remote()` before usethis made so much direct use of git2r. gh actually parses the config file. Now that we use git2r so much, we might as get remote names and URLs this ourselves, through more conventional means.

R/github.R
```diff
@@ -147,14 +148,29 @@ use_github <- function(organisation = NULL,
...
+  }
+  if (pushed) {
+    git2r::branch_set_upstream(git2r::repository_head(r), "origin/master")
+  } else {
+    ui_todo(c(
+      "Failed to push and set tracking branch.",
+      "This often indicates a problem with git2r and the credentials.",
+      "Try this in the shell, to complete the set up:",
+      "{ui_code('git push --set-upstream origin master')}"
+    ))
```
jennybc: Here's where we are most likely to fail, esp. on Windows, due to git2r credential-finding and using problems. I think it makes sense to catch that and provide the git command that will probably finish the job.

jennybc: This a version of @maelle's suggestion in the issue thread https://github.com/r-lib/usethis/issues/221#issuecomment-432581263

jennybc:
@maelle and @ijlyttle 

Would you two be willing to test drive this branch and bring your special flavour of soul-crushing Windows realism 😉? If you need inspiration for how to explore and break it, look at my manual test script `tests/manual/manual-use-github.R`.

R/git-utils.R
```diff
@@ -23,6 +23,20 @@ uses_git <- function(path = proj_get()) {
   !is.null(git2r::discover_repository(path))
 }
 
+# Remotes ------------------------------------------------------------------
+git_remotes <- function(path = proj_get()) {
```
hadley: These definitely shouldn't take a path, but should instead use `git_repo()`

R/github-utils.R
```diff
@@ -1,24 +1,30 @@
-## suppress the warning about multiple remotes, which is triggered by
-## the common "origin" and "upstream" situation, e.g., fork and clone
-gh_tree_remote <- function(path) {
-  suppressWarnings(gh::gh_tree_remote(path))
+github_remotes <- function(path = proj_get()) {
+  remotes <- git_remotes(path)
+  if (length(remotes) == 0) return(NULL)
```
hadley: What happened to official Jim style? 😛

R/github.R
```diff
@@ -66,26 +66,29 @@ use_github <- function(organisation = NULL,
...
+  repo_desc <- project_data()$Title %||% ""
 
   if (interactive()) {
     ui_todo("Check title and description")
     ui_code_block(
       "
-      Name:        {repo_name},
-      Description: {repo_desc}
-      ",
+        Name:        {repo_name},
```
hadley: Are these deliberately getting extra indentation?

R/github.R
```diff
@@ -143,18 +144,33 @@ use_github <- function(organisation = NULL,
     }
   }
 
-  ui_done("Pushing to GitHub and setting remote tracking branch")
+  ui_done("Pushing {ui_value('master')} to GitHub and setting remote tracking branch")
```
hadley: Maybe add "branch" here, just to be extra clear?

R/github.R
hadley: Shouldn't we encourage to re-run `use_github()`? That will cause more pain for us in the short-term, but will force us to resolve the problems in the loing-term.

R/github.R
```diff
@@ -66,26 +66,29 @@ use_github <- function(organisation = NULL,
                        auth_token = NULL,
                        host = NULL) {
   check_uses_git()
+  check_branch("master")
+  check_uncommitted_changes()
+  check_no_origin()
```
hadley: I think aborting if origin is specified is too strong - we should be able to complete the branch set up if they got the remote set up last time.

R/github.R
```diff
@@ -243,29 +259,62 @@ browse_github_pat <- function(scopes = c("repo", "gist"),
...
+    This repo already has an {ui_value('origin')} remote, with value {ui_value(origin)}.
+    How to remove this setting in the shell::
+    {ui_code('git remote rm origin')}
+    "
+  )
+}
 
-check_uses_github <- function(base_path = proj_get()) {
-  if (uses_github(base_path)) {
+check_uses_github <- function(path = proj_get()) {
```
hadley: I think if you're going to touch `path()` you should remove it altogether.

hadley: (If you touch it, you have to bring it up to code)

R/github.R
jennybc: Well, no, I think.

I played with various options to dealing with partial setup, but it's hard to account for the various weird scenarios people could show up with. If we fail here, the first push is the only thing that hasn't happened. The GitHub repo exists and 'origin' is configured with that URL. I think it's ok to leave people here. Literally the only thing they need to to is `git push --set-upstream origin master`.

jennybc: If they fail due to ... not a git repo, not on master, working tree dirty, GitHub repo pre-exists, no APT ... we can (and do) tell them what to fix. They fix and call `use_github()` again.

If they fail due to because `gitr2::push()` fails, I don't even know how to help them. Do they need to build git2r from source to link against the correct system libraries? Do they need to setup an ssh agent? Do they need to allow the agent to store their passphrase? Who knows?!?

R/git-utils.R
jennybc: OK done, here and many other places.

R/github.R
jennybc: No. Fixed.

R/github-utils.R
jennybc: I thought these were OK:

```
if (length(remotes) == 0) return(NULL)

if (length(remotes) == 0) {
  return(NULL)
}
```

but this was not:

```
if (length(remotes) == 0)
  return(NULL)
```

@jimhester what is your ruling?

R/github.R
jennybc: I really think we shouldn't touch a repo that has a preconfigured `origin` remote. We do say exactly how to remove this, so they can call `use_github()` again.

If we accept a pre-existing `origin`, now we have to see where it points. Is it a GitHub URL? Does that repo already exist? Does it have the owner and repo name we expect? Etc etc.

R/github-utils.R
jimhester: @jennybc has it right, all on one line is ok since it is then not possible to add additional statements.

hadley: Oh I really dislike that formatting 😞

jimhester: I never really do that myself, I always use braces and put it on a new line

jennybc: Braces and new lines, it is then!

PR #518, opened 2018-11-29
Enhancements to use_tidy_labels

hadley:
* Can provide a repo spec
* Can rename labels
* Can add descriptions to labels
* Can selectively colour labels


DESCRIPTION
```diff
@@ -29,6 +29,7 @@ Imports:
     gh,
     git2r (>= 0.23),
     glue,
+    purrr,
```
jennybc: OK then!

hadley: It's now light weight, and makes life so much eaiser.

PR #525, opened 2018-12-01
Prepare externally-created paths under our policy

jennybc:
This was causing test failure for me locally, in the test I've edited:

```
>   expect_equal(path1, as.character(path2))
Error: `path1` not equal to as.character(path2).
1/1 mismatches
x[1]: "/var/folders/yx/3p5dt4jj1019st0x90vhm9rr0000gn/T//RtmpMtS78Z/file167c347ea5
x[1]: f6b"
y[1]: "/var/folders/yx/3p5dt4jj1019st0x90vhm9rr0000gn/T/RtmpMtS78Z/file167c347ea5f
y[1]: 6b"
```

It makes sense to treat a path stored in `R_PROFILE_USER` as a "user provided path" anyway.

### DIFF
--- a/R/edit.R
+++ b/R/edit.R
@@ -105,7 +105,7 @@ scoped_path_r  <- function(scope = c("user", "project"), ..., envvar = NULL) {
   if (scope == "user" && !is.null(envvar)) {
     env <- Sys.getenv(envvar, unset = "")
     if (!identical(env, "")) {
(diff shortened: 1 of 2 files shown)


hadley:
I think I fixed `Failure: write_ut8 always produces a trailing newline (@test-write.R#65)` on master

Issue #108, opened 2017-11-06
Consider encouraging the use of ORCID IDs in DESCRIPTION template


cboettig:
`usethis` promotes the use of machine-readable `Authors@R` format already, by providing an editable template:

```
Authors@R: person("First", "Last", email = "first.last@example.com", role = c("aut", "cre"))
```

With CRAN supporting ORCID ids via the comment field,`usethis` could consider promoting this practice as well, e.g.:

```
Authors@R: person("First", "Last",
  email = "first.last@example.com",
  role = c("aut", "cre"), comment = c(ORCID = "https://add_your_orcid_id"))
```

hadley:
I wonder if we generally need some better tooling for saying "this is my name, email address, and orc id"

cboettig:
Yes, it would be great if `utils::person` had a field for `id`.  Maybe `orcid` specifically, though maybe simply supporting URL would be a good start. (Generally I think schema.org is a good and popular standard for thinking about additional metadata terms).

For the moment, drafting the `comment` field as above seems to have some momentum; e.g. https://github.com/hadley/pkgdown/issues/421 and

![](https://pbs.twimg.com/media/DLr14xAW0AIZMqr.jpg)

hadley:
Maybe we should have `use_author()` ?

hadley:
Documentation at https://github.com/wch/r-source/commit/d140c744dbe383a83b6aece7f8fbeb5d5a6b7cfd#diff-1b797da2cb81237e2f828c280f718dbb

hadley:
Quick exploration:

```R
#' Add an author to `Authors@R`
#'
#' @param given,family Given and family names. Can be vectors
#' @param email Email address. Mandatory for maintainers; optional otherwise.
#' @param role Character vector of roles.
#' @param comment Any additional comments
#' @param orcid ORCID (<http://orcid.org/>) identifier
#' @noRd
use_author <- function(given, family, email = NULL, role = "ctb",
                       comment = NULL, orcid = NULL) {

  comment <- c(comment, ORCID = orcid)

  desc::desc_add_author(
    file = proj_get(),
    given = given,
    family = family,
    email = email,
    role = role,
    comment = comment
  )

  invisible()
}
```

Main problem is that it mangles all existing formatting.

cboettig:
Wow, there's a `desc` package now? that's awesome, wish I'd seen that sooner.  

The `use_author()` looks very promising to me!  

[...]

PR #530, opened 2018-12-04
Encourage ORCID use

maelle:
cf #108 in particular https://github.com/r-lib/usethis/issues/108#issuecomment-443096619

# About the changes made
I'm not sure whether (not) to introduce a line break in `person()`.

# About the changes not made
I had been considering adding links to `desc` (convenience) functions but I'm not sure how to and it's actually maybe a topic for another issue.
* Making sure `whoami` finds all your info could be part of the recommended setup (in the setup article) but `usethis` doesn't use `whoami` at all, only `desc::desc_add_me()` does.
* `desc` convenience functions [support ORCID](https://github.com/r-lib/desc/pull/70) but their use is actually beyond setup, it's more about managing authors later in a package life?
End of long comment. 😸

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -19,7 +19,8 @@
 #' ```
 #' options(
 #'   usethis.description = list(
-#'     `Authors@R` = 'person("Jane", "Doe", email = "jane@example.com", role = c("aut", "cre"))',
+#'     `Authors@R` = 'person("Jane", "Doe", email = "jane@example.com", role = c("aut", "cre"),
+#'                           comment = c(ORCID = "YOUR-ORCID-ID"))',
 #'     License = "MIT + file LICENSE",
 #'     Language =  "es"
 #'   )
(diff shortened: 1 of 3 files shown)


R/description.R
```diff
@@ -19,7 +19,8 @@
 #' ```
 #' options(
 #'   usethis.description = list(
-#'     `Authors@R` = 'person("Jane", "Doe", email = "jane@example.com", role = c("aut", "cre"))',
+#'     `Authors@R` = 'person("Jane", "Doe", email = "jane@example.com", role = c("aut", "cre"),
+#'                           comment = c(ORCID = "YOUR-ORCID-ID"))',
```
katrinleinweber: A bit nit-picky, sorry, but is the duplication of `ID` intended here & in the changes below?

maelle: You mean ORCID ID instead of ORCID? That's what ORCID writes https://orcid.org/register although I see it should be iD

jennybc:
I think this is a great start and next steps should probably unfold in desc as @maelle describes.

PR #540, opened 2018-12-06
Muffle setup and teardown of scoped_temporary_things

jennybc:
This is how the helper should have been in the first place. Now it's symmetric: setup and teardown are both quiet.

This makes it easier to manipulate `usethis.quiet` in a test, i.e. you can use `withr::local_options()` instead of `withr::with_options()`.

### DIFF
--- a/tests/testthat/test-use-rcpp.R
+++ b/tests/testthat/test-use-rcpp.R
@@ -25,5 +25,6 @@ test_that("use_src() doesn't message if not needed", {
   use_src()
 
   withr::local_options(list(usethis.quiet = FALSE))
+
   expect_silent(use_src())
 })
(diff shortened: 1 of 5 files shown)


tests/testthat/test-use-rcpp.R
```diff
@@ -24,8 +24,7 @@ test_that("use_src() doesn't message if not needed", {
   use_package_doc()
   use_src()
 
-  withr::with_options(
-    list(usethis.quiet = FALSE),
-    expect_silent(use_src())
-  )
+  withr::local_options(list(usethis.quiet = FALSE))
```
jennybc: I can't put this `local_options()` call next to `scoped_temporary_package()` w/o the intervening commands generating unwanted output.

cc @hadley

PR #544, opened 2018-12-10
Fix a few minor things in release_checklist

jimhester:
`rhub::check_with_sanitizers()` was missing a closing backtick and
the argument to `use_version()` should be a string of c("major", "minor",
"patch", "dev") or NULL, not a numeric version.


jimhester:
Also `usethis::use_release_tag()` doesn't currently exist

hadley:
Thanks - I'll try and take a look at the other small problems in the next couple of days.

PR #557, opened 2018-12-20
use_readme_rmd(): make git hooks optional -- closes #556

### LINKED ISSUE #556: use_readme_rmd ingests git hook
Currently `use_readme_rmd()` does ingest a pre commit Git hook into the package's local git repository. There are two problems resulting from this: (1) This hook also prevents updates of the repo when the Rmd-file is simply re-rendered (e.g. automatically obtained information is updated). (2) User are neither informed about the ingestion happening and have no possibility to prevent it. 

This is related to the discussion found here https://github.com/r-lib/usethis/issues/312 and to that found here https://twitter.com/jeroenhjanssens/status/1075010648587886592.

I have forked this repo and will file a pull request with a possible solution that will ...

1. allow users to opt in or opt out per function argument, 
2. will default to opt out in non-interactive environments, 
3. will ask user what to do in case of an interactive session, 
4. and will have tests covering the new adopted behaviour.

petermeissner:
See issue #556 for further information.

### DIFF
--- a/R/readme.R
+++ b/R/readme.R
@@ -12,6 +12,10 @@
 #' YAML frontmatter and R fenced code blocks (`md`) or chunks (`Rmd`).
 #'
 #' @inheritParams use_template
+#' @param use_hook Use pre commit Git hook to enforce README.Rmd and README.md
+#'   to be in sync or not. Defaults to NULL which will result in asking user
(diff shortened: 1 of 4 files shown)


hadley:
This seems too complicated to me. Let's brainstorm other solutions in the issue before committing to something.

PR #560, opened 2019-01-04
Clarify this help message with a bit of verbosity. 

statwonk:
The `fun` reference caused me a bit of "foo confusion." I thought, "is there some magic where I'm supposed to refer to `mutate` like `dplyr::fun(mutate)(...)`?" This update hopes to close that line of confusion by focusing the user on direct reference like `dplyr::mutate`.


hadley:
Thanks!

Issue #293, opened 2018-02-26
use_bitbucket() (et al.)


chris-billingham:
Having the good fortune to use bitbucket here in the office it would be tremendous if there were analogue functions to connect to bitbucket rather than github.

Likewise: create_from_bitbucket(), use_tidy_bitbucket(), and so on through all the various github helper functions.

jennybc:
FYI I'd be open to PRs here. Probably *not* a good first issue, though ... hence the label / lack of label.

It would definitely merit some advance discussion/planning. Might make sense to couple with a conversation about gitlab, because that's always the next question :)

chris-billingham:
Whilst I'd love to step up and help here being the original requestor I probably don't have the programming chops to cover this, however happy to help anyway I can.

LDalby:
use_gitlab() etc. would be super useful too.

overmar:
I was actually thinking the addition of a gitlab-ci.yml template would be nice for those of us that use gitlab (ie the same process as use_travis()).

jennybc:
@overmar I'm going to delete your comment re: gitlab ci (a la travis) because that should be opened as a separate issue. I am about to label this issue as fair fame for tidy verse developer day, hence my heightened concern re: scope.

overmar:
Thanks @jennybc I created a new issue for the creation of only the ci file in gitlab. https://github.com/r-lib/usethis/issues/565

jennybc:
A recent r-hub blog post [About URLs in DESCRIPTION](https://blog.r-hub.io/2019/12/10/urls/) contains a table that is relevant to this issue:

```
## # A tibble: 1,855 x 2
##    domain                    n
##    <chr>                 <int>
##  1 github.com             4660
##  2 www.r-project.org       164
##  3 cran.r-project.org      143
##  4 r-forge.r-project.org    82
##  5 bitbucket.org            67
##  6 sites.google.com         54
##  7 arxiv.org                52
##  8 gitlab.com               44
##  9 docs.ropensci.org        38
## 10 www.github.com           32
```

Taking a very naive summary of this suggests that ~98% of the URLs relating to remote hosts are GitHub vs 2% for bitbucket + gitlab.

So this is why we're pretty reluctant to take on (much) code relating to these other hosts, which also aren't used by the usethis maintainers.

antoine-sachet:
There is no need for separate functions. `use_remote` should just pick up the right remote type. That's actually a straightforward change to a single function: `package_remote`. 

[...]

PR #562, opened 2019-01-08
use_gitlab_ci function, ala use_travis

overmar:
Based on https://github.com/r-lib/usethis/issues/293. These are functions to create a basic .gitlab-ci.yml at the top level of a package, in the same fashion as the use_travis functions.


R/ci.R
```diff
@@ -111,3 +111,27 @@ appveyor_info <- function(base_path = proj_get()) {
...
   list(url = url, img = img)
 }
+
+#' @section `use_gitlab_ci()`:
+#' Adds a basic `.gitlab-ci.yml` to the top-level directory of a package. This is
+#' a configuration file for the [Gitlab CI/CD](https://docs.gitlab.com/ee/ci/) continuous
+#' integration service for Gitlab.
+#' @export
+#' @rdname ci
+use_gitlab_ci <- function(browse = interactive()) {
```
jennybc: I also don't see any use of `browse`.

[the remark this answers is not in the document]
overmar: As I had in another issue, I am working on a use_gitlab() function that should have the same functionality as use_github(). But in gitlab the actual file is called gitlab-ci.yml which is why I went with this. However let me know what you would like it to be called.

R/ci.R
```diff
@@ -111,3 +111,27 @@ appveyor_info <- function(base_path = proj_get()) {
...
+  check_uses_git()
+  ext <- rlang::arg_match(ext)
```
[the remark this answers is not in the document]
overmar: Hold over from use_travis(), I'm removing it.

R/ci.R
```diff
@@ -111,3 +111,27 @@ appveyor_info <- function(base_path = proj_get()) {
...
+    ".gitlab-ci.yml",
+    ignore = TRUE
+  )
+  if (!new) return(invisible(FALSE))
+
+  invisible(TRUE)
+}
+
+uses_gitlab_ci <- function(base_path = proj_get()) {
+  path <- glue("{base_path}/.gitlab-ci.yml")
```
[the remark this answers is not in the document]
overmar: I'll make that update

R/ci.R
overmar: Also holdover from use_travis(). With gitlab, there is no registration for the ci process. As long as .gitlab-ci.yml is present in the root directory of the package it runs by default.

R/ci.R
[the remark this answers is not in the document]
jennybc: Do `path <- path(base_path, ".gitlab-ci.yml")`, assuming we keep the `base_path` argument.

If we got rid of the `base_path` argument and made the active project the implied location, we'd do `proj_path(".gitlab-ci.yml")`.

R/ci.R
jennybc: Should the `browse` argument go away then?

overmar: I think for this one it would make sense.

overmar:
I made the changes requested by @jennybc, I will put in a pull request later to change the other use_* functions to use fs::path instead of glue. Thank you for explaining the change over to fs::path from glue.

R/ci.R
```diff
@@ -111,3 +111,39 @@ appveyor_info <- function(base_path = proj_get()) {
 
   list(url = url, img = img)
 }
+
+#' @section `use_gitlab_ci()`:
+#' Adds a basic `.gitlab-ci.yml` to the top-level directory of a package. This is
+#' a configuration file for the [Gitlab CI/CD](https://docs.gitlab.com/ee/ci/) continuous
+#' integration service for Gitlab.
```
jennybc:
```suggestion
#' integration service for GitLab.
```

R/ci.R
```diff
@@ -111,3 +111,39 @@ appveyor_info <- function(base_path = proj_get()) {
...
+use_gitlab_ci <- function() {
+  check_uses_git()
+  new <- use_template(
+    "gitlab-ci.yml",
+    ".gitlab-ci.yml",
+    ignore = TRUE
+  )
+  if (!new) return(invisible(FALSE))
+
+  invisible(TRUE)
```
jennybc: I think all the return bits can be simplified to `invisible(new)`.

R/ci.R
```diff
@@ -111,3 +111,39 @@ appveyor_info <- function(base_path = proj_get()) {
...
+}
+
+check_uses_gitlab_ci <- function(base_path = proj_get()) {
+  if (uses_gitlab_ci(base_path)) {
+    return(invisible())
+  }
+
+  ui_stop(
+    "
+    Cannot detect that package {ui_(project_name(base_path))} already uses Gitlab CI.
```
jennybc:
```suggestion
    Cannot detect that package {ui_(project_name(base_path))} already uses GitLab CI.
```

overmar:
Ok the checks have passed, and I had a coworker look it over. We use this variation internally, I just figured it would be nice to get it from the main package rather than having an add-on.

overmar:
@jennybc anything else that needs to be done?

.Rbuildignore
```diff
@@ -1,3 +1,4 @@
+^\.gitlab-ci\.yml$
```
jennybc: I don't think the usethis `.gitignore` file needs this. Is this perhaps leftover from some experimentation? If so, can you remove this change?

NEWS.md
```diff
@@ -36,6 +36,9 @@
 * `use_github_release()` creates a draft GitHub release using the entries
   in  `NEWS.md` (#137).
 
+* `use_gitlab_ci()` creates a draft gitlab-ci.yaml for use with Gitlab
```
```suggestion
* `use_gitlab_ci()` creates a draft gitlab-ci.yaml for use with GitLab 
```

jennybc:
I have a few pending comments and proposed changes that I don't think you've addressed. I just added two, but some were pre-existing. Are you seeing those?

overmar:
I see them now, I will get the rest of those cleaned up and pushed and will let you know.

.Rbuildignore
overmar: Removed it in https://github.com/r-lib/usethis/pull/562/commits/4fe65038e08dab007556adb5602e5760926ad925

R/ci.R
overmar: Updated in https://github.com/r-lib/usethis/pull/562/commits/4fe65038e08dab007556adb5602e5760926ad925

overmar:
@jennybc I think all comments and changes have now been addressed. Let me know if I missed anything. Thanks

jennybc:
LGTM, thanks!

inst/templates/gitlab-ci.yml
```diff
@@ -0,0 +1,46 @@
...
+
+testing:
+    stage: test
+    allow_failure: true
+    when: on_success
+    only:
+        - master
+    script:
+        - Rscript -e 'install.pacakges("DT")'
+        - Rscript -e 'cc <- covr::package_coverage; covr::gitlab(coverage = cc); cat("Code coverage:", paste0(format(covr::percent_coverage(cc), nsmall = 1),"%"))'
```
jimhester: This won't work at all as-as, you are not actually calling `package_coverage`, just assigning the function to a different name.

I think this could be simplified to just

```r
covr::gitlab(quiet = FALSE)
```

With the mentioned regex above if you want the automatic coverage badge.

inst/templates/gitlab-ci.yml
```diff
@@ -0,0 +1,46 @@
...
+  - test
+  - deploy
+
+building:
+  stage: build
+  script:
+    - R -e "remotes::install_deps(dependencies = T)"
+    - R -e 'devtools::check()'
+
+# To have the converage percentge appear as a gitlab badge follow these instructions
```
jimhester: As mentioned below I think a better solution is to just use `covr::gitlab(quiet = FALSE)`. In that case you would modify your coverage string to be `Coverage: \d+\.\d+`

jennybc:
@overmar or @jimhestee will one of you open a new issue about the things that need attention?

inst/templates/gitlab-ci.yml
overmar: Thanks for the advice on that, I think when I originally wrote this, I had a reason for doing it that way, but in the end, it doesn't make much sense. I'll move the resolution to https://github.com/r-lib/usethis/issues/738

PR #574, opened 2019-01-16
Auto-detect protocol in use_github

jimhester:
This method is from https://help.github.com/articles/testing-your-ssh-connection/

It certainly seem to work fine if ssh is already setup, but I am unsure what would exactly would happen if github is not in the allowed hosts and you get

> The authenticity of host 'github.com (IP ADDRESS)' can't be established.

The motivation was mass chaos when people tried to use `use_github()` with the default protocol of `ssh` while not having ssh setup during our workshop.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -61,7 +61,7 @@
 #' }
 use_github <- function(organisation = NULL,
                        private = FALSE,
-                       protocol = getOption("usethis.protocol", default = "ssh"),
+                       protocol = getOption("usethis.protocol", default = NULL),
                        credentials = NULL,
                        auth_token = NULL,
                        host = NULL) {
@@ -127,7 +127,13 @@ use_github <- function(organisation = NULL,
 
   ui_done("Adding GitHub remote")
   r <- git2r::repository(proj_get())
-  protocol <- match.arg(protocol, c("ssh", "https"))
+
+  if (is.null(protocol)) {
+    protocol <- detect_protocol(credentials)
+  } else {
+    protocol <- match.arg(protocol, c("ssh", "https"))
+  }
+
   origin_url <- switch(protocol,
     https = create$clone_url,
     ssh = create$ssh_url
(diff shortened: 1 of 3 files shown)


hadley:
I'm in favour of merging this since it seems unlikely to make anything worse; do you have any strong feelings @jennybc?

jimhester:
I guess one thing we might want to do is set a timeout on the `processx::run()` call. I am slightly worried that if ssh tries to prompt for a password it would just hang otherwise. Not really sure how long the timeout should be though, maybe 2 seconds?

R/github.R
```diff
@@ -279,3 +285,21 @@ check_gh_token <- function(auth_token) {
   }
   auth_token
 }
+
+detect_protocol <- function(credentials) {
+  # If a non-null credentials just return ssh
+  if (!is.null(credentials)) {
+    return("ssh")
```
jennybc: It is possible to pass credentials to git2r other than ssh. There's a whole family of functions: `cred_env()`, `cred_ssh_key()`, `cred_token()`, and `cred_user_pass()`. I do use them when git2r is being particularly obtuse about finding credentials. I wonder if this is a bit too simple? I.e. will still end up with lots of https people being treated as ssh people.

R/github.R
```diff
@@ -279,3 +285,21 @@ check_gh_token <- function(auth_token) {
...
+  res <- processx::run("ssh", c("-T", "git@github.com"), error_on_status = FALSE)
+
+  has_ssh <- identical(res$status, 1L) &&
+    grepl("successfully authenticated", res$stderr)[[1]]
+
+  if (has_ssh) {
+    "ssh"
+  } else {
+    "https"
+  }
```
jennybc: I propose a newsy `ui_done()` to relate our findings.

R/github.R
jennybc: We could cache in global git settings or strongly encourage people to set their `usethis.protocol` option. Maybe we should double down on the latter in this case, since we already support it.

jennybc:
I decided to attack this in a different way. See #648 Improve handling of git protocol.

`use_git_protocol()` is a new function to establish the git protocol. On demand or upon first need.

PR #577, opened 2019-01-19
Preface usethis function call with package name.

trestletech:
I didn't have `usethis` library()d, so when I got the instruction to `Call edit_r_environ() to open...` I got a "could not find function" error.

With this change, I would have gotten it right on the first try.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -237,7 +237,7 @@ browse_github_pat <- function(scopes = c("repo", "gist"),
   )
   view_url(url)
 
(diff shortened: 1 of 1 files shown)


jennybc:
How do you end up getting this message w/o having usethis attached? Just want to understand the path to getting here. But yeah if other packages trigger this, we can't assume the user knows we're talking referring to a function in usethis.

hadley:
This seems like a reasonable fix to me, although if this comes up a lot we might consider a specialised `ui_usethis_call()` or similar.

PR #580, opened 2019-01-19
use_lintr() proposal

### LINKED ISSUE #579: FR: use_lintr function to help set up code linting for package
The package does not have a `use_lintr` helper, but I use the `lintr` package with all of my packages and it is helpful to have a function that can automate the set up of this piece of package scaffolding. Others may find it similarly useful.

I see three typical ways to use `lintr` with a package:
 
* local interactive use
* automatic use on Travis CI post-build (non-failing)
* automatic local and Travis CI as a unit test (failing)

I think the first option should be the default. It is the simplest. The second option is the official recommended approach, which involves adding to .travis.yml:

```
after_success:
  - R CMD INSTALL $PKG_TARBALL
  - Rscript -e 'lintr::lint_package()'
```

but this option is specifically for cases where someone does not wish to have a unit test, or a build failure when the linter finds style issues.

The third option is the most rigorous and is intended to fail a build, locally and on Travis, if there are style issues present. Personally, I use this option almost exclusively in my packages, but it should not be the default because it is the most complex and also may not be desirable for long-running test suites.

I have an initial version of a function and will submit a pull request.

leonawicz:
This is an initial version of a `use_lintr` function that provides three typical methods for how to integrate code linting into package development. It is not ready to be merged yet but this can start the discussion about how it can be improved. It is also lacking unit tests currently.

Fixes #579


R/linting.R
```diff
@@ -0,0 +1,48 @@
...
+    .lintr <- "inst/.lintr"
+  } else {
+    .lintr <- ".lintr"
+  }
+
+  sink(.lintr)
+  cat(paste0("linters: with_defaults(\n  ", "line_length_linter(80))\n"))
+  sink()
+
+  if(type == "test" && !file.exists(".lintr")) file.symlink("inst/.lintr", ".lintr")
```
leonawicz: This symbolic link approach is problematic. This is the "old way" of making lintr used as a `testthat` unit test work with Travis CI. I think there is a newer, better way to do this that does not involve a symbolic link to a `.lintr` file located inside `inst` but I do not know how this should be altered to get rid of the linking approach for the unit test setup.

R/linting.R
```diff
@@ -0,0 +1,48 @@
...
+#' `use_lintr()` Adds support for code linting using the `lintr` package in one of three ways:
+#' local interactive linting only, non-failing (post-build) Travis CI linting, or failing local and Travis CI `testthat` unit test.
+#'
+#' @param type Options are \code{local}, \code{travis} or \code{test}.
+#' @export
+use_lintr <- function(type = c("local", "travis", "test")){
+  type <- match.arg(type)
+
+  if(type == "test"){
+    dir.create("inst", showWarnings = FALSE)
```
jennybc: This should probably use the internal `create_directory()` or exported `use_directory()` for consistency with the rest of usethis.

R/linting.R
```diff
@@ -0,0 +1,48 @@
...
+    .lintr <- "inst/.lintr"
+  } else {
+    .lintr <- ".lintr"
+  }
+
+  sink(.lintr)
+  cat(paste0("linters: with_defaults(\n  ", "line_length_linter(80))\n"))
+  sink()
```
jennybc: We have existing helpers for writing (into) files. See the functions in `write.R`.

R/linting.R
jennybc: usethis uses the fs package for all file system and file path operations. So that should be on the radar as you resolve this.

R/linting.R
leonawicz: Thank you. I'll look into all these changes I need to make to resolve these inconsistencies vis a vis `usethis` and `fs` functions.

tests/testthat/test-use-lintr.R
```diff
@@ -0,0 +1,24 @@
...
+})
+
+test_that("use_lintr('travis') creates build-ignored, promised file", {
+  scoped_temporary_package()
+  use_lintr("travis")
+  expect_true(is_build_ignored("^\\.lintr$"))
+  expect_proj_file(".lintr")
+})
+
+# test_that("use_lintr('test') creates build-ignored, promised files", {
```
leonawicz: This test cannot be run because the `testthat` unit test setup option for `use_lintr` still depends on the old, unacceptable symlink approach. I am hoping someone knows of an alternative approach. Obviously the symbolic link cannot be created without elevated privileges.

Separately, I am unsure how to test the writing out of a test file. If I tentatively uncomment this test and elevate my privileges to enable writing of the symbolic link, the test subsequently fails because of an inability to write out the test file during testing.

tests/testthat/test-use-lintr.R
```diff
@@ -0,0 +1,24 @@
+context("use_lintr")
+
+test_that("use_lintr('local') creates build-ignored, promised file", {
```
leonawicz: Should I add `lintr` to Suggests in DESCRIPTION?

jennybc: No, in general we don't necessarily add a package to Suggests here in usethis, just because usethis offers some support for it. It would not be very sustainable. They way we test related functionality is to use a test mock to get around the presence test. See, for example, `use_tutorial()` is tested vis-a-vis learnr.

R/linting.R
```diff
@@ -0,0 +1,46 @@
...
+      after_success:
+      - R CMD INSTALL $PKG_TARBALL
+      - Rscript -e 'lintr::lint_package()'
+      "
+    )
+    ui_todo("Code linting set up for Travis CI. Edit {ui_path('inst/.lintr')} as needed.")
+  } else {
+    ui_todo("Code linting set up for interactive use. Edit {ui_path('.lintr')} as needed.")
+  }
+  use_package("lintr", "suggests")
```
jennybc: This should near the top of the body, since we would exit early lintr is not installed.

hadley:
@jennybc I think we need to agree on one way of using lintr before adding the support for 3 different styles.

R/linting.R
leonawicz: Thanks, I'll make the suggested changes.

leonawicz:
Even though I use lintr as a unit test in packages, that mode is excessive for most users. Maybe the best option for one agreed upon mode is the simplest one: lintr for interactive use only, where all that is created is a default config file?

Also, since users may not initially be familiar with how to edit this file to customize their linting preferences, is better to create one that includes some linting defaults rather than an empty config file? Placing entries in the config file that are already lintr defaults is technically redundant, but would help show users how the file is structured. Thoughts?

R/linting.R
leonawicz: I've made the changes. I also had to tentatively comment out all three unit tests. I'm still having difficulty understanding how to make those work in this context.

jennybc:
> I think we need to agree on one way of using lintr before adding the support for 3 different styles.

Yeah and the challenge is that I, at least, don't really use lintr with any regularity. Which is one reason to give this a bit more time to simmer.

ThierryO:
May I suggest my workflow on this? It might be useful in the discussion on the potential workflows.

- A unit test in `tests/testthat/test_zzz_coding_style.R`, this file is in `.gitignore` and runs only locally, integrated in the unit tests.
- A separate step in `.travis.yml`. As `tests/testthat/test_zzz_coding_style.R` is never commited, lintr is only run as a separate step and not during the unit tests.
- If I need it interactively I use `lintr::lint_package()` or the RStudio addin.

hadley:
Over a year later and I think it's more obvious how this should now work: as a github action. So I'm going to close this issue — thanks @leonawicz for all your hard work, and we're sorry it didn't end up making it into usethis for before other technology moved on.

PR #582, opened 2019-01-19
Tidy dev day issue 526

### LINKED ISSUE #526: `git_vaccinate()` drops blank lines and repeated comments
`git_vaccinate()` calls `write_union()` to update the global gitignore file and `write_union()` keeps only unique lines.

https://github.com/r-lib/usethis/blob/e9a3c8299c132f9fc8a630886e26c36668f427c7/R/write.R#L61

Unfortunately this deletes all empty lines (but the first) and repeated comments, such as those often used around section headings, e.g.

```
value = 123

########################
# A new section        #
########################

value2 = 456
```

Becomes

```
value = 123

########################
# A new section        #
value2 = 456
```

Since you are already doing a set diff of the new lines and the existing lines, couldn't you just prepend the new lines rather than taking the union to avoid this? e.g. change that line to

```r
all <- c(new, existing_lines)
```

khailper:
Closes #526

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -62,6 +62,34 @@ write_union <- function(path, lines, quiet = FALSE) {
   write_utf8(path, all)
 }
 
+#' @describeIn write-this writes lines to a file, taking what's
+#'   already there, if anything, and appending some new lines. Note, there is no
+#'   explicit promise about the line order. Designed to modify simple config
+#'   files like `.Rbuildignore` and `.gitignore`.
+#' @export
+write_append <- function(path, lines, quiet = FALSE) {
+  stopifnot(is.character(lines))
+  path <- user_path_prep(path)
+
+  if (file_exists(path)) {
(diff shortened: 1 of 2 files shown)


R/write.R
```diff
@@ -62,6 +62,34 @@ write_union <- function(path, lines, quiet = FALSE) {
...
+  new <- setdiff(lines, existing_lines)
+  if (length(new) == 0) {
+    return(invisible(FALSE))
+  }
+
+  if (!quiet) {
+    ui_done("Adding {ui_value(new)} to {ui_path(path)}")
+  }
+
+  all <- c(lines, existing_lines)
```
hadley: I think you can simply change this line in `write_union()`; there's no need to create a function.

khailper: Will do; assumed without checking it was used elsewhere.

PR #583, opened 2019-01-19
Be more respectful of existing lines in write_union()

### LINKED ISSUE #526: `git_vaccinate()` drops blank lines and repeated comments

khailper:
Closes #526 `git_vaccinate()` drops blank lines and repeated comments


jimhester:
The code LGTM. @khailper can you please add a bullet to NEWS? It should briefly [describe the change](https://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Issue #534, opened 2018-12-05
Making pr_*() functions work for anyone besides Hadley 😇


jennybc:
I'm trying out the workflow with a PR in readr.

My local readr repo is the active usethis project.

My remote situation:

```
jenny@2015-mbp readr $ git remote -v
origin  https://github.com/jennybc/readr.git (fetch)
origin  https://github.com/jennybc/readr.git (push)
upstream        https://github.com/tidyverse/readr.git (fetch)
upstream        https://github.com/tidyverse/readr.git (push)
```

That is: `origin` is my fork and `upstream` is the main repo.

`pr_fetch(933)` fails here for me:

```
>   pr <- gh::gh("GET /repos/:owner/:repo/pulls/:number",
+     owner = github_owner(),
+     repo = github_repo(),
+     number = number
+   )
Error in gh_process_response(raw) : 
GitHub API error (404): 404 Not Found
Message: Not Found
Read more at https://developer.github.com/v3/pulls/#get-a-single-pull-request

URL not found: https://api.github.com/repos/jennybc/readr/pulls/933
```

because it tries to get the PR from my fork instead of the main repo. We have to be fork-savvy re: `owner`. Shortly thereafter, this also influences the `user` object that is the name of the remote where the PR lives.

jennybc:
To get the ball rolling

``` r
library(gh)

x <- gh("/repos/jennybc/readr")

x[["fork"]]
#> [1] TRUE

x[["parent"]][["full_name"]]
#> [1] "tidyverse/readr"

(upstream_urls <- c(
  x[["parent"]][["ssh_url"]],
  x[["parent"]][["clone_url"]]
))
#> [1] "git@github.com:tidyverse/readr.git"    
#> [2] "https://github.com/tidyverse/readr.git"
  
r <- git2r::repository("~/rrr/readr")

rnames <- git2r::remotes(r)

(remote_urls <- setNames(git2r::remote_url(r, rnames), rnames))
#>                                   origin 
#>   "https://github.com/jennybc/readr.git" 
#>                                 upstream 
#> "https://github.com/tidyverse/readr.git"

m <- remote_urls %in% upstream_urls
remote_urls[m]
#>                                 upstream 
#> "https://github.com/tidyverse/readr.git"
```

<sup>Created on 2018-12-04 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

hadley:
Resolved by #602 and #533

PR #587, opened 2019-01-19
add exported pr_finish to documentation

cderv:
Related to #534 

`pr_finish` was exported but not in the documentation


hadley:
Thanks!

PR #593, opened 2019-01-23
Added Stale Bot

muschellij2:
I've added `use_stale` for the [Stale Bot](https://github.com/apps/stale).  Also there is a small bug fix for when there is a new line in the title.


hadley:
Can you please tease apart the different issues (bug fix, doc tweaks, and new feature) into different PRs? I'd also recommend starting stale bot discussion in an issue since it's the first bot, and we should discuss the overall structure before implementing a single bot.

Issue #594, opened 2019-01-23
use_travis function not showing up in the pkgdown site


apreshill:
@jennybc 

From the source code, I see two exported functions under [`ci.R`](https://github.com/r-lib/usethis/blob/master/R/ci.R) that are not showing up in the [current pkgdown site](https://usethis.r-lib.org/reference/index.html):

1. `use_travis()`
1. `use_appveyor()`

Neither are in the [`_pkgdown.yml`](https://github.com/r-lib/usethis/blob/master/_pkgdown.yml)- perhaps they belong in their own "CI" section? Note that `browse_travis()` DOES show up under both Git & GitHub, and Browse sections on the [pkgdown site](https://usethis.r-lib.org/reference/index.html).

hadley:
I think CI would be a good place. Do you want to do a PR?

PR #596, opened 2019-01-23
add ci section to index use_travis and use_appeyor on pkgdown site

apreshill:
This addresses issue #594. 

I took my best guess for where to place this section in the `_pkgdown.yml`. 

Also noting that the following are missing from the pkgdown index:
`Warning: Topics missing from index: git_sitrep, git_vaccinate, pr_init, use_citation, use_github_release, use_release_issue`


hadley:
Thanks!

PR #599, opened 2019-01-23
Re-export := in use_tidy_eval().

### LINKED ISSUE #595: Re-export := with use_tidy_eval
:= (the tidyeval assignment operator, which I pronounce "digested is") is often as useful for tidyeval as the functions exported by the file generated via usethis::use_tidyeval(). It would be helpful if it were also re-exported.

jonthegeek:
Closes #595 .

I didn't add a test, because the current test doesn't actually check the contents of the file (only that it exists), and I was already a bit lost in how that test works. I can take a stab at a more specific test if you'd like, though.


lionel-:
Thanks!

PR #600, opened 2019-01-23
Broken push with Newline in title

muschellij2:
## PR goal: fix newline
This is a smaller PR than #493 which has fixed the paths issue.  This may be fixed as I've opened issue with https://github.com/r-lib/gh/issues/92.  Other than whitespace, the main fix is simple `gsub` on `\n`.  

Here would be a `reprex` but the non-interactive choices need to be made with `ui_yeah`, so I'll use the example https://github.com/r-lib/gh/issues/92 to show that it's an issue.
```r
library(usethis)
pkgpath <- file.path(tempdir(), "testpkg")
create_package(pkgpath, open = FALSE) # creates package below temp directory
proj_set(pkgpath)

dfile = proj_path("DESCRIPTION")
desc = desc::desc(file = dfile)
desc$set("Title" = "I Want a Newline\n In My Title")
desc$write(file = dfile)
desc$get("Title")


## now, working inside "testpkg", initialize git repository
use_git()

## create github repository and configure as git remote
use_github(protocol = "https") ## to use https
```

## Typo/Whitespace 
Also, slight fix with spacing on `template.R`, where `reprex` is (before fix):
``` r
usethis::use_template("blah")
#> Error: Could not find template 'blah'in package 'usethis'.
```
and 
``` r
usethis::use_template("blah")
#> Error: Could not find template 'blah' in package 'usethis'.
```
after fix.


hadley:
This still seems to include a bunch of unrelated changes which makes it hard to review.

hadley:
It looks like you're working in master, which tends to cause problems — you might want to read https://happygitwithr.com/fork-and-clone.html#dont-touch-master

PR #468, opened 2018-09-27
Add use_rcpp_armadillo and use_rcpp_eigen; closes #421

duckmayr:
usethis already has use_rcpp() to set up packages to use Rcpp.
Similarly, it would be helpful to have use_rcpp_armadillo() and
use_rcpp_eigen() to set up packages to use RcppArmadillo and RcppEigen
(@alexpghayes requested this feature in issue #421).
This commit adds both of these functions, as well as tests for them.

use_rcpp_armadillo() ensures Rcpp and RcppArmadillo are in "LinkingTo"
in the DESCRIPTION and that Rcpp is in "Imports", creates the src/
directory and an appropriate .gitignore, as use_rcpp() does,
sets up the proper src/Makevars and src/Makevars.win files (respecting
any existing content in those files), informs the user how to fix up
their NAMESPACE file, and instructs them to run devtools::document().

Similarly, use_rcpp_eigen() ensures Rcpp and RcppEigen are in
"LinkingTo" and "Imports" in the DESCRIPTION, creates the src/
directory and an appropriate .gitignore, as use_rcpp() does,
informs the user how to fix up their NAMESPACE file, and instructs
them to run devtools::document().

PR #601, opened 2019-01-24
Add use_rcpp_armadillo() and use_rcpp_eigen(). (closes #421)

### LINKED ISSUE #421: Feature request: use_rcpp_armadillo() and use_rcpp_eigen()
RStudio provides package templates for packages that use `Rcpp` and `RcppArmadillo`, but if you decide to use `RcppArmadillo` mid-project it can be a bit tricky to remember what goes where. I use `RcppArmadillo` infrequently and can never remember the setup that needs to happen. Having helpers

- `use_rcpp_armadillo`
- `use_rcpp_eigen`

would be incredibly convenient.

coatless:
This is a cleaned up PR of the work that @duckmayr initially started in #468 based on @hadley's request to incorporate `use_rcpp()` as a starting point for both `use_rcpparmadillo()` and `use_rcppeigen()`.


NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## New features
 
+* `use_rcpp_armadillo()` and `use_rcpp_eigen()` are new (exported) functions
```
hadley: No need to say "(exported)" as that's obvious from the context

NEWS.md
```diff
@@ -2,6 +2,10 @@
 
 ## New features
 
+* `use_rcpp_armadillo()` and `use_rcpp_eigen()` are new (exported) functions 
+   that set up a package to use RcppArmadillo or RcppEigen, respectively
+   (#601, @duckmayr, @coatless).
```
hadley: This should mention the issue number, not your PR.

R/rcpp.R
```diff
@@ -51,3 +84,52 @@ use_src <- function() {
 
   invisible()
 }
+
+set_makevars <- function(path) {
```
hadley: This feels very fragile to me, and I don't want to maintain a function that attempts to modify any existing makevars. I think it would be better to have a simple path when makevars doesn't exist, and print instructions to the user when it does.

tests/testthat/test-use-rcpp.R
```diff
@@ -18,6 +18,39 @@ test_that("use_rcpp() creates files/dirs, edits DESCRIPTION and .gitignore", {
   expect_true(all(c("*.o", "*.so", "*.dll") %in% ignores))
 })
 
+test_that("use_rcpp_armadillo() creates Makevars files and edits DESCRIPTION", {
+  skip_if_not_installed("RcppArmadillo")
```
hadley: I am concerned about skipping these and not adding to DESCRIPTION. I don't particularly want these as heavy dependencies, but I do want the tests to run on travis. Maybe `use_dependency()` needs some infrastructure to facilitate testing?

coatless: As the Travis-CI builds are run by compiling all packages, I wanted to avoid adding another compilation requirement. Having said this, there is precedence elsewhere in `usethis` for:

1. adding packages to suggest _without_ skipping tests c.f. `pkgdown`. 
2. using test skips when a package is unavailable, c.f. `use_logo()` when `magick` cannot be found.

For now, I've opted to go the first route and have added the packages to the `Suggests` field of the `DESCRIPTION`. I say this because I'm not sure a valid solution would be to modify `use_dependency()` as the requirements should be kept as script as possible.

coatless:
Based on the build matrix, when I added `RcppArmadillo` to `Suggests` in `DESCRIPTION` this triggered an error for all _R_ versions less than `R >= 3.3.0`. Note: `RcppEigen` has a dependency for _R_ >= 2.15.1.

R/rcpp.R
```diff
@@ -1,7 +1,8 @@
-#' Use C and C++.
+#' Use C, C++, RcppArmadillo, or RcppEigen.
 #'
-#' Creates `src/` and adds needed packages to `DESCRIPTION`, and
-#' optionally creates `.c` or `.cpp` file.
+#' Creates `src/`, adds required packages to `DESCRIPTION`,
+#' optionally creates `.c` or `.cpp` files
+#' as well as `Makevars` and `Makevars.win` files.
```
hadley: , and where needed, `Makevars`...

R/rcpp.R
```diff
@@ -25,6 +26,41 @@ use_rcpp <- function(name = NULL) {
...
+  makevars_path <- proj_path("src/Makevars")
+  makevars_win_path <- proj_path("src/Makevars.win")
+
+  makevars_settings <- c(
+    "CXX_STD = CXX11",
+    "PKG_CXXFLAGS = $(SHLIB_OPENMP_CXXFLAGS)",
+    "PKG_LIBS = $(SHLIB_OPENMP_CXXFLAGS) $(LAPACK_LIBS) $(BLAS_LIBS) $(FLIBS)"
+  )
+
+  create_makevars(makevars_path, "RcppArmadillo", makevars_settings)
```
hadley: I think `create_makevars()` should figure out the paths to the makevars, and should probably just write both files.

R/rcpp.R
```diff
@@ -25,6 +26,41 @@ use_rcpp <- function(name = NULL) {
...
+
+#' @rdname use_rcpp
+#' @export
+use_rcpp_eigen <- function(name = NULL) {
+  use_rcpp(name)
+
+  use_dependency("RcppEigen", "LinkingTo")
+  use_dependency("RcppEigen", "Imports")
+
+  roxygen_ns_append("@import RcppEigen") && roxygen_update()
```
hadley: Can you briefly explain to me why RcppEigen needs to be imported but RcppArmadillo does not?

R/rcpp.R
```diff
@@ -51,3 +87,25 @@ use_src <- function() {
 
   invisible()
 }
+
+create_makevars <- function(path, package, settings = NULL) {
+
+  if (!is.null(settings)) {
+    makevars_settings <- paste0(settings, collapse = "\n")
+  } else {
+    makevars_settings <- c("# For help writing Makevars{.win}, please see",
```
hadley: Given that this code is not currently used, I'd recommend removing it.

tests/testthat/test-use-rcpp.R
```diff
@@ -18,6 +18,35 @@ test_that("use_rcpp() creates files/dirs, edits DESCRIPTION and .gitignore", {
...
+  use_rcpp_armadillo()
+  expect_match(desc::desc_get("LinkingTo", pkg), "RcppArmadillo")
+
+  makevars_settings <- c(
+    "CXX_STD = CXX11",
+    "PKG_CXXFLAGS = $(SHLIB_OPENMP_CXXFLAGS)",
+    "PKG_LIBS = $(SHLIB_OPENMP_CXXFLAGS) $(LAPACK_LIBS) $(BLAS_LIBS) $(FLIBS)"
+  )
+
+  makevars <- readLines(proj_path("src", "Makevars"))
```
hadley: This test is fragile because you're testing if we ever change what is in the makefile, we'll have to update all these tests. I think you'd be better of just testing that the files exist, and then creating a separate test that tests that `create_makevars()` works both when the file already exists and when it doesn't.

tests/testthat/test-use-rcpp.R
```diff
@@ -18,6 +18,35 @@ test_that("use_rcpp() creates files/dirs, edits DESCRIPTION and .gitignore", {
   expect_true(all(c("*.o", "*.so", "*.dll") %in% ignores))
 })
 
+test_that("use_rcpp_armadillo() creates Makevars files and edits DESCRIPTION", {
+  pkg <- scoped_temporary_package()
+  use_roxygen_md()
```
hadley: Is this necessary?

R/rcpp.R
```diff
@@ -51,3 +87,25 @@ use_src <- function() {
 
   invisible()
 }
+
+create_makevars <- function(path, package, settings = NULL) {
```
hadley: Should this be called `use_makevars()` and match the signature of the other `use_*` functions? I don't think we should export it at this point, but we should consider the API.

R/rcpp.R
coatless: Short Version: RcppEigen doesn't need Imports and this has been removed.

Longer version: This was an artifact of how `RcppEigen.package.skeleton()` created a new package. As a result, I've issued a patch for the RcppEigen project that addresses it and we've incorporated the change here.

tests/testthat/test-use-rcpp.R
coatless: Per the decision to deprecate supporting manually edited `NAMESPACE` files in `usethis`, the use of `roxygen2` is required for `roxygen_ns_append()` function and only `use_roxygen_md()` setups the needed components. 

It may be beneficial to add a function solely enabling `roxygen2` via `use_roxygen()`. This would be better left to a separate PR.

Note: This is also the initial setup used by the `use_rcpp()` test.

tests/testthat/test-use-rcpp.R
coatless: I've aimed to test for the existence but not the content of either `Makevar{.win}` files in the `use_rcpp_armadillo()` test.

In addition, I've also created two unit tests for `use_makevars()` (the updated name for `create_makevars()`). The first test ensures new flags are not added and the second writes `Makevars` files if none are present.

coatless:
I went through and updated the PR per the last round of suggestions by @hadley. I've responded to prior review comments with justification. 

**Please note:** The Travis-CI build will show as a failure for version 3.2 and 3.1 since `RcppArmadillo` requires _R_ >= 3.3.0. Per an earlier discussion, I added `RcppArmadillo` under `Suggests:` in `DESCRIPTION`.

tests/testthat/test-use-rcpp.R
hadley: Got it, thanks

hadley:
Thanks!

DESCRIPTION
```diff
@@ -45,6 +45,8 @@ Suggests:
     pkgdown (>= 1.1.0),
     rmarkdown,
     roxygen2,
+    RcppEigen,
+    RcppArmadillo,
```
jennybc: Why does usethis itself need these in Suggests?

coatless: @jennybc the `Suggests:` change was to allow the unit tests for `use_rcpp_armadillo()` and `use_rcpp_eigen()` to work on Travis-CI. Initially, these tests were skipped via `skip_if_not_installed()` as it would add onto the build requirements. However, after being requested by @hadley, the packages were added to `Suggests:` with the aforementioned disclaimers after each code review. I'm more than happy to remove them and add `skip_if_not_installed()`. 

c.f. https://github.com/r-lib/usethis/pull/601/files/5b0d7911cdadb396cf2c8e831ca501aa053f52a5#diff-695f4452520470c8e82fa2f184ee14b1

hadley:
I reverted because it breaks on older version of R.

coatless:
@hadley any reason why we can't just remove `RcppArmadillo` from `Suggests:` in `DESCRIPTION` and add `skip_if_not_installed()`? There is precedence elsewhere as indicated previously. 

`RcppEigen` only requires R >= 2.15.1.

hadley:
The main problem is that it will never get tested

jennybc:
Maybe something like this will work? I had to do something pretty similar to test `use_roxygen_md()`.

https://github.com/r-lib/usethis/blob/0adfb5d902cc20bba3e324e52e234695f3a81a28/tests/testthat/test-use-roxygen.R#L15-L30

coatless:
@jennybc looks good. I'm more than happy to submit a new PR with that logic.

coatless:
As the arrival of spring break nears... @hadley would you accept a PR that incorporates @jennybc's suggestion regarding the _R_ versioning trick used for `roxygen2`?

jennybc:
I had to do some fiddling to make the suggested test mocking strategy actually work @coatless.

Therefore it turns out this more recent PR is a better template: https://github.com/r-lib/usethis/pull/645

coatless:
Thanks for the update @jennybc. I'll attempt this over the weekend/Monday as the "fun" part of spring break begins!

PR #602, opened 2019-01-24
Improvements to PR flow

jennybc:
Progress towards #534 Making pr_*() functions work for anyone besides Hadley 😇

Work out more kinks in the PR flow facilitated by `usethis::pr_*()` functions. Specifically, I wanted better functionality around pushing back into PRs.

I did a thorough analysis of why this wasn't working well (see below).

TL;DR = **upstream tracking branch is primarily about `git pull`, not `git push`**. To push local PR branch into its upstream tracking branch, which might not be on `origin` and might have a different branch name, requires some explicit effort.

Also did a general pass through git utils. We are slowly reinventing a haphazard version of githug 😐, but luckily it's all internal.

Best way to review might be to just use the `pr_*()` functions and see how it goes.

---

*Optional background but since I wrote this down, I don't want to lose it.*

### Setup done by `pr_fetch()`

  * Add the PRer's fork as REMOTE.
  * Fetch the remote ref corresponding to THEIR_PR_BRANCH.
  * Create local OUR_BRANCH with REMOTE/THEIR_PR_BRANCH as start-point; in general, the branch names will differ, since many PRs come from `master`.
  * Set REMOTE/THEIR_PR_BRANCH as upstream tracking branch for OUR_BRANCH.
  * If external PR, set push config for REMOTE to do OUR_BRANCH -> THEIR_BRANCH.
    - Open question: what if we do `pr_fetch()` for >1 PR from a single fork?
  * Checkout OUR_BRANCH, to prepare for exploration and extension.

``` bash
OUR_BRANCH <--> REMOTE/THEIR_PR_BRANCH
^^ checkout
```

### Desired Pit of Success

We want easy forms of `git push` to "just work" to send new local commits into REMOTE/THEIR_PR_BRANCH:

  * Big Green Arrow in RStudio
  * `git push` at the command line
  * `pr_push()`

### Bad News

It's not that simple and we can't have all of this (yet).

`git push` does NOT simply consult the upstream tracking branch, the way `git pull` does.

### Solutions

  1. Explicit command line push: `git push REMOTE HEAD:THEIR_PR_BRANCH`
  1. Git config:
  
      ``` bash  
      push.default=upstream
      ```
      You could do this at the user level via `git config --global push.default upstream`. This has the effect that `git push` **will** consult the upstream tracking branch. Note the default these days is `push.default=current`, which seems desirable and not worth messing with.
  1. Git config:
  
      ``` bash  
      remote.REMOTE.push = refs/heads/OUR_BRANCH:refs/heads/THEIR_PR_BRANCH
      ```

The last solution makes these methods work:

  * `git push` at the command line
  * `git2r::push()` with no arguments

### RStudio is a bust currently

None of these solutions will help us push with RStudio right away. But I opened an issue and the associated PR has already been merged: <https://github.com/rstudio/rstudio/issues/4201>

### Background

https://git-scm.com/docs/git-push

The general form of a push command:

``` bash
git push <repository> <src>:<dst>
                      \_________/ <-- "refspec"
```

`<src>` can be any valid revision, such as a branch name, a SHA-1, `HEAD`, `master~4`, etc.

`<dst>` is much more constrained: it must be an actual ref, like `contributor/patch-1` or `/refs/heads/contributor/patch-1`.

> When the command line does not specify where to push with the `<repository>` argument, `branch.*.remote` configuration for the current branch is consulted to determine where to push. If the configuration is missing, it defaults to `origin`.

`pr_fetch()` does the necessary configuration and REMOTE gets picked up correctly for us.

> When the command line does not specify what to push with `<refspec>...` arguments or `--all`, `--mirror`, `--tags` options, the command finds the default `<refspec>` by consulting `remote.*.push` configuration, and if it is not found, honors `push.default` configuration to decide what to push.

In this PR, I configure `remote.*.push` and make sure `git_branch_push()` does the equivalent of `git push REMOTE HEAD:THEIR_PR_BRANCH`.


R/pr.R
```diff
@@ -58,38 +58,55 @@ pr_fetch <- function(number, owner = NULL) {
...
+  if (!git_branch_exists(our_branch)) {
+    their_refname <- git_remref(remote, their_branch)
+
+    ui_done("Creating local branch {ui_value(our_branch)}")
+    git2r::fetch(git_repo(), remote, refspec = their_branch, verbose = FALSE)
+    git_branch_create(our_branch, their_refname)
+
+    git_branch_track(our_branch, remote, their_branch)
+
+    ui_done("Caching URL for PR #{number} in config for branch {ui_value(our_branch)}")
```
hadley: I'm not sure it's useful to expose this to the user. It seems like an internal detail?

R/pr.R
```diff
@@ -58,38 +58,55 @@ pr_fetch <- function(number, owner = NULL) {
...
+    config_url <- glue("branch.{our_branch}.pr-url")
+    git_config_set(config_url, pr$html_url)
+
+    if (remote != "origin" && our_branch != their_branch) {
+      ui_done("Setting {ui_value(remote)} push config as {ui_value(glue('{our_branch}:{their_refname}'))}")
+      config_key <- glue("remote.{remote}.push")
+      config_value <- glue("refs/heads/{our_branch}:refs/heads/{their_branch}")
+      config_args <- rlang::list2(
```
hadley: I think you can just use exec here:

```R
rlang::exec(git2r::config,
  repo = git_repo(), 
  global = FALSE, 
  !!config_key := config_value
)
```

But can't you just use `git_config_set()`?

jimhester: 👍 for `git_config_set()`, that rlang call gives me the heebie jeebies ;)

jennybc:
Special thanks to @jtr13 who let me hammer her Happy Git PR with test commits as I worked through this.

PR #608, opened 2019-01-25
PR improvements

### LINKED ISSUE #575: FR: a helper function to update a fork
Starting to contribute on tidyverse dev day, I find myself in the situation of updating my existing fork with last changes.  

I don't know what is your worflow but mine is opening the terminal and 
```sh
git checkout master
git pull upstream master
git push origin master
```

and now I am ready to go (using the `pr_*` functions I guess)

Would you find it helpful to have a function in usethis for updating fork like this ? 
Seems like a helper for PR contribution

hadley:
* New tooling to update upstream master, where needed (fixes #575) — this makes it easier for PR-makers to keep up to date, and harder for them to start off from a bad place.

* Work around pain in github api

* New tools for fork and original source repo detection

* No longer try and make `git push` at the command line work, recognising that there's no way to dig a pit of success here, and instead we should be conservation with changing git config.


R/pr.R
```diff
@@ -5,15 +5,15 @@
 #'    anticipation of making a PR.
 #' * `pr_fetch(number)` downloads a remote PR so you can edit locally.
 #' * `pr_push()` pushes local changes to GitHub, after checking that there
-#'    aren't any remote changes you need to retrieve first. It will create
-#'    a PR if needed
-#' * `pr_pull()` retrieves changes from the GitHub PR. Run this if others
-#'    have made suggestions or pushed into your PR.
+#'    aren't any remote changes you need to retrieve first. It will help you
+#'    initiate a PR, if needed.
```
jennybc: I say this so people understand we won't actually open a PR automatically, i.e. they'll still get a chance to back out.

R/pr.R
```diff
@@ -5,15 +5,15 @@
...
+#' * `pr_pull()` retrieves changes from the GitHub PR. Run this if the remote PR
+#'    has commits you don't have.
```
jennybc: Reworded because it's not always "your PR", i.e. account for the maintainer's use case.

PR #610, opened 2019-01-26
Don't worry about files that are uncommitted AND untracked, by default

### LINKED ISSUE #592: Relax check for uncommitted changes?
Would it be possible to relax the check for uncommitted changes, e.g. for `pr_fetch()` and others?

I pretty regularly have purposefully untracked files that I have no plan to commit. It is a bit of a pain to `git stash --include-untracked` them every time I have to use a `pr_*()` function, or I guess I could add them to `.gitignore`.

For me at least I would prefer if this only checked if you had changes in tracked files, and did not consider untracked files.

jennybc:
Fixes #592 Relax check for uncommitted changes?

### DIFF
--- a/R/git-utils.R
+++ b/R/git-utils.R
@@ -189,19 +189,19 @@ check_uses_git <- function(base_path = proj_get()) {
   ))
 }
 
-check_uncommitted_changes <- function(path = proj_get()) {
+check_uncommitted_changes <- function(path = proj_get(), untracked = FALSE) {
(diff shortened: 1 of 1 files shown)


jennybc:
Test driving the `pr_*()` family of functions w/o this is a great trial for me. Let's put ourselves out of this particular misery.

jimhester:
LGTM, thanks!

hadley:
I think we probably didn't need to make this an option, but since you've done the work, there's no point undoing it.

PR #613, opened 2019-01-30
Check file name for non-ascii characters and remove tolower() to ensure file name fidelity in use_test()

### LINKED ISSUE #581: `usethis::use_test()` does not remain true to original *.R open source file name
With `use_test()`, if the current source editor file name is, e.g. camel case, the `test-*.R` file that is created is converted to all lower case with a call to `tolower()`. I think this is making an unnecessary decision for the user. Additionally, if the user tries to force the file name by passing a `name =` argument, this is also converted to lower case.

Example of the problem with default:
```r
source editor filename: "my_coolFunctionPrint.R"
use_test()
resulting test: "test-my_coolfunctionprint.R"
```
Example of the problem with `name = ***` passed:
```r
source editor filename: "my_coolFunctionPrint.R"
use_test("theBestFunctionEver")
resulting test: "test-thebestfunctionever.R"
```

This is likely unnoticed if strict adherence to the Tidyverse style guide, but other style guides will break the filename.R -> test-filename.R pattern. There appears to be a `tolower()` call somewhere converting the case that needs some conditional logic to it. I'd be happy to submit a PR if necessary.

stufield:
Check file name for non-ascii characters, remove `asciify()` from `slug()`, remove `tolower()` from `asciify()` to ensure name fidelity, update unit tests accordingly.

Fixes #581

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -183,6 +183,8 @@ usethis gains tooling to manage part of a file. This currently used for managing
 * `use_pkgdown()` is now configurable with site options (@jayhesselberth, #467),
   and no longer creates the `docs/` directory (#495).
 
+* `use_test()` no longer forces the filename to be lowercase (#613, @stufield).
+
 * `use_test()` will not include a `context()` in the generated file if used 
   with testthat 2.1.0 and above (the future release of testthat) (#325).
 
--- a/R/rmarkdown.R
+++ b/R/rmarkdown.R
@@ -19,7 +19,7 @@
 #' use_rmarkdown_template()
 #' }
 use_rmarkdown_template <- function(template_name = "Template Name",
-                                   template_dir = asciify(template_name),
+                                   template_dir = tolower(asciify(template_name)),
                                    template_description = "A description of the template",
                                    template_create_dir = FALSE) {
 
--- a/R/test.R
+++ b/R/test.R
@@ -13,10 +13,10 @@ use_testthat <- function() {
   check_installed("testthat")
 
   use_dependency("testthat", "Suggests")
-  use_directory("tests/testthat")
+  use_directory(path("tests", "testthat"))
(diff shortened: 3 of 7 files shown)


stufield:
Hi Hadley ... I notice that Travis CI failed one of the tests so I'm not sure if I need to fix anything on my end. I didn't notice anything obvious from the details. Please let me know what (if anything) you need me to do. Thx.

tnearey:
So is a version with this fix available? If so where exactly?
Editorial:
theforcingoflowercaseisanoyingForUsCamelCaseTypes_who_are_not_that_fond_of_underscores_either.

llrs:
I think that that error was unrelated to your changes (Although I don't understand why it failed). 
Hope this get merged soon

jennybc:
@stufield I'll be reviewing PRs soon. What's up with the large commented block here "REMOVE THIS PRIOR TO MERGE"?

stufield:
Oh geez, this was a while ago, but I *believe* this was when Hadley and I were going back and forth about how this feature should be implemented. I was unsure if we were going to change our minds, so I was hoping to have an opportunity to "clean up" before submitting a final PR. 

Specifically for this section, originally there was an `if else` branch where if the name was `NULL` we used the active file name, otherwise we checked (or something like that) ... but during this PR we thought the `name` should be checked irrespective of whether passed by the user or using the default. Removing the commented section, the `if else`, you bumped into (sorry for the confusion), would achieve that.

It's been a while, and I'm not sure where things stand at the moment, but would it be easier for me to clean it up and re-submit the PR or for you to do it on your end? Happy to do whatever you prefer.

jennybc:
If I see clear path forward when I review it, I'll just make it so (I can push commits here) and we'll get it done. Or I'll leave comments.

PR #615, opened 2019-01-30
Reflect that pkgbuild is now on CRAN

maurolepore:
Warning: This PR fixes the source but does not re-build the site.


hadley:
Thanks! The site will be re-built automatically.

Issue #625, opened 2019-02-09
Helpers for GitHub Actions


maxheld83:
[GitHub actions](http://github.com/features/actions) is a new workflow automation (~ CI/CD) service from GitHub, still in limited beta as of now.
I think it has some potential for R.

I've written a package for quickly getting started with GitHub actions: [ghactions](http://www.maxheld.de/ghactions/).

It's very closely modelled on usethis, and also includes a front line `ghactions::use_ghactions()` function that writes out sensible defaults for many projects.

It's still very early days for GitHub actions, and ghactions (the pkg) is still maturing.

But *if* things work out and the R community finds GitHub actions helpful, I'd love to write up a short PR wrapping `ghactions::use_ghactions()` in / exporting to `usethis::use_ghactions()` for easier access.

So this will be some time, just wanted to raise this here lest someone else do the same work again.

And also, of course, to gather feedback on [ghactions](http://www.maxheld.de/ghactions/).

hadley:
Fixed by #929, now that github actions have completed changed their UI since you started working on this 😬

PR #629, opened 2019-02-14
WIP: edit and browse GitHub Actions

VerenaHeld:
Adds initial support for GitHub Actions as per #625, but without any dependency on @maxheld83's and my WIP package [ghactions](https://github.com/maxheld83/ghactions).

We initially implemented this simple interaction functions (to edit and to browse) in our own package, but had trouble emulating usethis behaviour, because not all necessary functions are exported from usethis. 

We hope this will provide some initial support for GitHub Actions in good "usethis style".


maxheld83:
sorry we opened the PR too early last night 😬.
Should be all good now from our end *including tests*.
(You guys are so awesomely thorough! We didn't even consider there'd be tests for `edit_*()` and friends! But there are! 😊)

maxheld83:
one last thing we're not 100% sure is whether we got the scope thing right. `.github/main.workflow` files only meaningfully exist at the repo (~ project, package) level, *not* at the user level (as a `.gitignore` might).
So we omitted those switch statements and also the argument.

Hope this makes sense in the bigger scheme of usethis.

R/edit.R
```diff
@@ -167,3 +168,11 @@ git_ignore_path <- function(scope = c("user", "project")) {
 
   path
 }
+
+#' @export
+#' @rdname edit
+edit_workflow <- function() {
```
maxheld83: edit_workflow can be simpler than above functions (think @VerenaHeld and myself) because `.github/main.workflow` *only* exist at repo, never user level (as opposed to a `.gitignore`).

NAMESPACE
```diff
@@ -20,6 +21,7 @@ export(edit_r_environ)
 export(edit_r_makevars)
 export(edit_r_profile)
 export(edit_rstudio_snippets)
+export(edit_workflow)
```
jennybc: The name seems to have the wrong implied scope to me. It feels very global, whereas I'd like to find a way to suggest the "GitHub actions" scope.

R/browse.R
```diff
@@ -25,6 +26,8 @@
 #' browse_github_issues("backports", 1)
 #' browse_github_pulls("rprojroot")
 #' browse_github_pulls("rprojroot", 3)
+#' # Leads to 404, because usethis does not currently use GitHub Actions.
+#' browse_github_actions("usethis")
```
jennybc: I suspect it's better to `\dontrun{}` this then. In which case, I'd drag it down to the bottom, by itself.

R/edit.R
jennybc: Seems right.

tests/testthat/test-edit.R
```diff
@@ -126,3 +127,12 @@ test_that("edit_git_ignore() ensures .gitignore exists in project", {
   edit_git_ignore("project")
   expect_proj_file(".gitignore")
 })
+
+test_that("edit_workflow() ensures .github/main.workflow exists in project", {
+  # workflows only exist per project, never at user level
```
jennybc: I think these comments can be omitted.

jennybc:
Also, can you think through any `Rbuildignore` considerations?

maxheld83:
@jennybc thanks for the thoughtful suggestions, will incorporate them.

Agreed that `edit_workflow()` doesn't do enough yet; [ghactions](https://github.com/maxheld83/ghactions) is all about templates, but that stuff isn't ready for prime time yet.

I'm going to be working on ghactions in earnest for a couple of weeks under @jimhester's guidance soon.

I'll then place this PR on WIP until we have a more mature feature-rich ghactions to wrap as per #625.

jennybc:
I'm readying for release and am trying to get our PRs / issues under control for the first time in a while.

I'm going to close this in light of your planned, concerted work on GitHub actions. You can make a new PR from the same branch or, more likely, you will want to start over. See you again then!

Issue #533, opened 2018-12-05
Build protocol flexibility into pr_*() functions


jennybc:
`pr_fetch()` is currently hard-wired for ssh

https://github.com/r-lib/usethis/commit/9bc4f6d5359cf457bab7dd9489d8b194c774b29c#r31557485

hadley:
Seems like we don't need an argument for `pr_fetch()` since we already have a repo — we just need a `git_protocol()` that tells protocol of current repo. Something like this?

```R
git_protocol <- function() {
  r <- git_repo()
  if (!"origin" %in% git2r::remotes(r)) {
    ui_stop("origin remote not configured")
  }
  
  url <- git2r::remote_url(r, "origin")
  if (substr(url, 1, 4) == "git@") {
    "git"
  } else if (substr(url, 1, 8) == "https://") {
    "https"
  } else {
    ui_stop("Unknown git protocol")
  }
}
```

PR #632, opened 2019-02-17
adding https support in pr_* workflow

cderv:
To contribute in discussion in #533, here are the modification I made in this 📦 so that I can use `pr_*` functions on my windows computer where I use only https remote and github token credential.

There is a few commits because I tried several approach (like passing protocol in all functions) but I have made simpler at the end. Everything is based on the fact that 
* Default behaviour is for `ssh`
* `https` can be use if option `usethis.protocol = "https"` is set. In this case, everything in usethis will not use ssh but http
    * remote creation
    * git credentials in git2r using GITHUB_PAT or GITHUB_TOKEN, both supported by usethis

Mainly there was those two main point to deal with
* remote must be created using `clone_url` not `shh_url`
* credentials must be pass to `git2r` functions through the `credentials` argument, and created with one of `cred_*` functions from `git2r` 📦 

On other approach I had in mind was only take care that usethis creates remote using https not ssh url and that after, credentials must be provide as an option to the package using any of the ones supported in `git2r`. I stayed for now with my fist approach. 

So, this is currently working for me to contribute using usethis and https remote. 

It is note perfect yet but hoping this can contribute to the work on this topic. If you find it interesting and want to build on this PR, please tell me and I can work more to improve it, and maybe make it more clever that just defining behavior from an option.

### DIFF
--- a/R/git-utils.R
+++ b/R/git-utils.R
@@ -307,11 +307,9 @@ git_has_ssh <- function() {
 
 git_credentials <- function(protocol = getOption("usethis.protocol", default = "ssh")) {
   if(protocol == "https") {
-    env_var <- Sys.getenv(c("GITHUB_PAT", "GITHUB_TOKEN"), unset = "", names = TRUE)
-    which_set  <- env_var != ""
-    if (any(which_set)) {
-      token_env <- names(env_var)[which_set][[1]]
-      git2r::cred_token(token = token_env)
+    token_var <- git_token_var()
+    if (token_var != "") {
+      git2r::cred_token(token = token_var)
     } else {
       ui_stop(
         "
@@ -323,4 +321,15 @@ git_credentials <- function(protocol = getOption("usethis.protocol", default = "
     # use default ssh from git2r
     NULL
   }
+}
+
+git_token_var <- function() {
+  env_var <- Sys.getenv(c("GITHUB_PAT", "GITHUB_TOKEN"), unset = "", names = TRUE)
+  which_set  <- nzchar(env_var)
+  if (any(which_set)) {
+    token_env <- names(env_var)[which_set][[1]]
+  } else {
+    token_env <- ""
+  }
+  token_env
 }
--- a/R/git-utils.R
+++ b/R/git-utils.R
@@ -9,14 +9,16 @@ git_init <- function() {
   git2r::init(proj_get())
 }
 
-git_pull <- function(remote_branch = git_branch_tracking()) {
+git_pull <- function(remote_branch = git_branch_tracking(),
+                     protocol = getOption("usethis.protocol", default = "ssh")) {
   repo <- git_repo()
 
   git2r::fetch(
     git_repo(),
     name = remref_remote(remote_branch),
     refspec = remref_branch(remote_branch),
-    verbose = FALSE
+    verbose = FALSE,
+    credentials = git_credentials(protocol)
   )
   mr <- git2r::merge(git_repo(), remote_branch)
   if (isTRUE(mr$conflicts)) {
@@ -136,18 +138,24 @@ git_branch_switch <- function(branch) {
   invisible(old)
 }
 
-git_branch_compare <- function(branch = git_branch_name()) {
+git_branch_compare <- function(branch = git_branch_name(),
+                               protocol = getOption("usethis.protocol", default = "ssh")) {
   repo <- git_repo()
 
   remref <- git_branch_tracking(branch)
-  git2r::fetch(repo, remref_remote(remref), refspec = branch, verbose = FALSE)
+  git2r::fetch(repo, remref_remote(remref),
+               refspec = branch,
+               verbose = FALSE,
+               credentials = git_credentials(protocol))
   git2r::ahead_behind(
     git_commit_find(branch),
     git_commit_find(remref)
   )
 }
 
-git_branch_push <- function(branch = git_branch_name(), force = FALSE) {
+git_branch_push <- function(branch = git_branch_name(),
+                            force = FALSE,
+                            protocol = getOption("usethis.protocol", default = "ssh")) {
   remote <- git_branch_tracking(branch)
   if (is.null(remote)) {
     remote_name   <- "origin"
@@ -163,7 +171,8 @@ git_branch_push <- function(branch = git_branch_name(), force = FALSE) {
     git_repo(),
     name = remote_name,
     refspec = glue("refs/heads/{branch}:refs/heads/{remote_branch}"),
-    force = force
+    force = force,
+    credentials = git_credentials(protocol)
   )
 }
 
@@ -303,4 +312,3 @@ git_credentials <- function(protocol = getOption("usethis.protocol", default = "
     NULL
   }
 }
-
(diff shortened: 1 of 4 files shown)


jennybc:
Thanks @cderv! This is a great start. I'm going to hijack this and finish it off, which means I'll be pushing into the branch/PR. I'm also going to take notes for myself here as I work.

jennybc:
Survey of existing user-facing functions that take GitHub protocol or credentials and how:

  * `use_github(protocol = getOption("usethis.protocol", default = "ssh"), credentials = NULL, auth_token = NULL)`
    - Needs `auth_token` to create the new companion GitHub repo. The return payload contains remote URLs.
    - `protocol` is consulted to choose the right remote URL for the `origin` remote 
    - `credentials` is passed to `git2r::push()` when `protocol = "ssh"`. When `protocol = "https"`, I create a username/password credential object from the `auth_token`.
  * `create_from_github <- function(protocol = getOption("usethis.protocol", default = "ssh"), credentials = NULL, auth_token = NULL)`    
    - Needs `auth_token` to GET info on the target repo (such as URLs) and, potentially, to create a fork
    - `protocol` is consulted to choose the right remote URL for `git clone` and, for a fork, setting an `upstream` remote
    - `credentials` is passed to `git2r::clone()`, because target could possibly be a private GitHub repo or hosted on GHE. I'm not making any effort to auto-populate this, i.e. create a username/password credential object from the `auth_token` ... but I could that and do so in `use_github()` 🤔

jennybc:
Upon further study, I might need to use this to force some deeper refactoring, in a new branch. More later

jennybc:
This PR currently touches:

  * `git_pull()`: injects git credentials to `git2r::fetch()` but not provided as an argument to `git_pull()`
  * `git_branch_compare()` *ditto*
  * `git_branch_push()`: *ditto, except `git2r::push()`*

This PR adds `git_credentials()`, which is where the git credentials come from in all of the above.

  * Returns `NULL` for ssh, so routes through git2r default for getting keys from the agent. For https, figures out the name of the GitHub PAT env var, packages with `git2r::cred_token()`, or dies trying. An alternative would be to package the `auth_token` handling inside `use_github()` and `create_from_github()`. The relies on `git2r::cred_user_pass()`. That seems more general, i.e. supports but does not require direct provision of the PAT. This would mean we don't need the other new function `git_token_var()`

jennybc:
Which `pr_*()` functions are affected and how?

`pr_fetch()` needs `protocol` to set the remote URL correctly in the case of an external PR. There is also a `git2r::fetch()` from the PR. In practice, I suspect it's rare that credentials are needed, because we're usually talking about public repos. But it's conceivable that someone could be doing with with a private repo. Or GHE, but we'd also need to expose `host` to support that. There's also a token-free `gh::gh()` call in here right now.

`pr_push()` needs `credentials`. It's not in this PR, probably because @cderv isn't pushing back into PRs.

`pr_pull()`, `pr_pull_source()` probably need `credentials` (or not) in the same way as `pr_fetch()`. `pr_finish()` calls `pr_pull_source()`.

`pr_sync()` needs `credentials`. It pulls and pushes.

jennybc:
Proposal: let's support PR workflows in public repos in GitHub.com.

cderv:
Thanks for sharing notes ! It is helpful and it allows me to answer one of your question

> pr_push() needs credentials. It's not in this PR, probably because @cderv isn't pushing back into PRs.

☺️ I am pushing back into PR. And I use this PR now to do it using `pr_*` functions. 

For `pr_push` the credentials are into the inner function `git_branch_push()` where the call to `git2r::push()` that require credential is.

`pr_pull()` and `pr_pull_source()` have also credentials through `git_pull()` where the call to `git2r::fetch()` is

`pr_sync` use all the above and use also credentials. 

I put `git_credentials()` in all `git2r` calls and each function is working using credentials. At least, I have no more issue now to work with github using `pr_*` functions.

jennybc:
@cderv Do you use `pr_*()` functions with private repos or on GHE?

ijlyttle:
Hi @jennybc and @cderv

FWIW, I had a quick look yesterday, and checked that one of the foundational functions, `usethis:::uses_github()`, returns `TRUE` (as hoped) for one of my GHE packages. 

I'll be happy to take them for a full spin on GHE, if you like.

jennybc:
Do we want to support private repo and GHE for the `pr_*()` functions? If no one will use it, I'd rather keep things simpler. But it would also be silly to NOT plan for it and then immediately get requests to handle that case.

cderv:
> @cderv Do you use `pr_*()` functions with private repos or on GHE?

I don't use GHE. 
I contribute to public repo I do not own but where I have write access (maintainers role: `pr_fetch`, `pr_push`, `pr_finish`) and it works also.  
I do use private repo sometimes but not very often. So I just tested it again and it seems to work fine. 
Here the kind of test run I made:
```r
usethis::create_package("testpackage")

usethis::use_git()
usethis::use_github(private  = TRUE)
usethis::pr_init("new")
usethis::pr_push()

# create a new branch manually on github and open a pr
usethis::pr_fetch(1)
# commit something
usethis::pr_push()
# go to master
usethis::pr_pause()
# commit on master and push
# go back on pr
usethis::pr_fetch(1)
# update PR from master
usethis::pr_sync()
```

I think private repos are working as is because the credentials are set and used with any functions. However, I don't have tested 2FA with https remote... 🤔 

GHE would mean to have a host argument (or option) in every function that needs it. Could be a little bit of work, and no easy to maintain (and to test)... 
I don't really know the diversity of GHE url but `uses_github()` works because this regex `"github[^/:]*[:/]{1,2}([^/]+)/(.*?)(?:\\.git)?$"` matched on _origin_ remote. It seems it was not on purpose. ☺️ 
 
> Do we want to support private repos and GHE for the `pr_*()` functions? If no one will use it, I'd rather keep things simpler. But it would also be silly to NOT plan for it and then immediately get requests to handle that case.

I completely agree with that. 👍 
I don't know how many people will use it. Personally, I think of `usethis` at the swiss knife to ease contribution to tidyverse and more generally open R source repos. I did this PR because I find it easier on windows (and for newcomers) to connect through https to remotes rather than ssh. Also, I wanted to use `usethis` that way. So focusing on PR workflow for public repo on github.com is the main objective.

Other thought: I wonder how `credentials` 📦 can change this PR, and works with git2r credentials logic.

jennybc:
I suspect it makes sense to set GHE aside, i.e. to not thread `host` through everywhere. But, to clarify, yes, credentials are necessary, even in public repos, in order to push into a PR.

ijlyttle:
I think it makes sense to set GHE aside for now, with the wish that the door not be closed to the possibility of a community member offering a PR at some point in the future :)

jennybc:
As I hinted, this did indeed get subsumed by bigger changes I have made, concluding with PR #656 `pr_*()` love for https. Sorry about that @cderv but this was very helpful for lighting a fire under me and getting things rolling.

cderv:
> Sorry about that @cderv but this was very helpful for lighting a fire under me and getting things rolling.

This was one the purpose to contribute to your thoughts on these features. The other was to get things working at least for me ☺️ 

Glad it helps and I will now give the new functions a go. (after my vacation I guess). 

Thanks for making this a priority for happy git (https) user 😉

PR #642, opened 2019-02-25
Set NOT_CRAN environmental variable

### LINKED ISSUE #641: Consider setting NOT_CRAN: true in Appveyor?
Would you consider a pull request to the Appveyor template that sets 

```
environment:
  global:
    NOT_CRAN: true
```

I believe this will fix the behavior of Appveyor such that it does not automatically skip tests that are marked `skip_on_cran()`.  Thanks to @ldecicco-USGS who pointed this out to me.

cboettig:
Currently this is not set, leading to `testthat::skip_on_cran()` to also skip tests on Appveyor.  Users should explicitly use `skip_on_appveyor()` to also skip appveyor tests. Closes #641.

### DIFF
--- a/inst/templates/appveyor.yml
+++ b/inst/templates/appveyor.yml
@@ -13,6 +13,9 @@ install:
 cache:
   - C:\RLibrary
 
+environment:
+  NOT_CRAN: true
+
 # Adapt as necessary starting from here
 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks!

PR #584, opened 2019-01-19
Add use_tutorial() function, documentation, tests

angela-li:
As mentioned in #314, with big shoutouts to @xvrdm @jules32 @jessesadler

We love tidyverse dev day!!

PR #645, opened 2019-03-04
Use tutorial, round 2

### LINKED ISSUE #314: use_tutorial()
For learnr tutorials.

angela-li:
Closes #314 use_tutorial()

Making this PR properly from a branch instead of master of my forked repo, incorporating code review edits in #584 

Merge me instead of #584! 😄


angela-li:
Looks like builds are failing because `learnr` isn't installed in Appveyor/Travis?

DESCRIPTION
```diff
@@ -47,7 +47,7 @@ Suggests:
     roxygen2,
     spelling (>= 1.2),
     styler (>= 1.0.2),
-    testthat (>= 2.0.0)
+    testthat (>= 2.0.0),
```
jennybc: Please delete this comma.

R/tutorial.R
```diff
@@ -0,0 +1,44 @@
+#' Create a learnr tutorial.
+#'
+#' Creates a new tutorial in `inst/tutorials`. Tutorials are interactive R
+#' Markdown documents built with the `learnr` package.
+#'
+#' @section General setup:
```
jennybc: Can you think of a different heading here? This somehow made me expect a description of setup *I* needed to do versus the setup I can expect this function to do. It's probably also OK to let this be a bullet list in the first/main/only paragraph of description. In that case, you can solve the heading dilemma by deleting it 😁

R/tutorial.R
```diff
@@ -0,0 +1,44 @@
...
+#' * Adds needed packages to `DESCRIPTION`.
```
jennybc: Suggest you just be specific and say that learnr is added to Suggests.

tests/testthat/test-use-tutorial.R
```diff
@@ -0,0 +1,26 @@
+context("test-use-tutorial")
+
+test_that("use_tutorial() checks its inputs", {
+  skip_if_not_installed("rmarkdown")
+
+  scoped_temporary_package()
+  expect_error(use_tutorial(), "no default")
+  expect_error(use_tutorial(name = "tutorial-file"), "no default")
```
jennybc: Suggest you add one test with an invalid `name`, since the docs claim that certain requirements are enforced.

tests/testthat/test-use-tutorial.R
```diff
@@ -0,0 +1,26 @@
...
+  use_tutorial("tutorial-file", "Tutorial Title")
+  expect_proj_dir(path("inst", "tutorials"))
+
+  use_tutorial(file_name, "Tutorial Title")
+  expect_proj_file(path("inst", "tutorials", file_name, ext = "Rmd"))
+
+  use_tutorial("tutorial-file", title)
+  expect_equal(title, rmarkdown::yaml_front_matter(
+    path("inst", "tutorials", "tutorial-file", ext = "Rmd")
+  )$title)
```
jennybc: I don't quite follow this test. Specifically, why does it call `use_tutorial()` 3 times with, I think, effectively the same arguments? It seems you could call it once.

jennybc:
There's a test failing on AppVeyor and Travis. You probably need to use the same mocking strategy we use elsewhere to get around the need for learnr to be installed.

https://github.com/r-lib/usethis/blob/b0df5f1ca90d68fa72a9d337972d361792e94749/tests/testthat/test-use-roxygen.R#L18-L29

jennybc:
@garrettgman If you took this for a spin, that would be great. It basically looks good to me, so I plan to merge soon, after we wrap up the loose ends.

jennybc:
Can you also please add a bullet to NEWS? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jennybc:
Good grief it took me a while to get the test and mocking sorted -- probably just as well that I was the one to suffer through that.

Thanks!

angela-li:
You caught me at a busy time, but thanks for helping out with the rest of the edits! Appreciate your support on this (thanks for figuring out the mocking). It's exciting to see `use_tutorial` added into usethis 👏

PR #647, opened 2019-03-08
Add pr_*() family to the pkgdown function-references

ijlyttle:
I don't know if this is premature or not, but it will help me find the reference more-easily on the dev pkgdown site.


jennybc:
Seems like a good idea. Thanks!

PR #648, opened 2019-03-10
Improve handling of git protocol

### LINKED ISSUE #533: Build protocol flexibility into pr_*() functions

jennybc:
Laying the groundwork for another PR that will address #533 Build protocol flexibility into pr_*() functions, among other things. I see a few things to build/refactor first.

MAIN THING:

`use_git_protocol()` is a new function to summon the git protocol.

  * It can be called proactively by the user and it's also used inside functions.
  * It honours the `usethis.protocol` option, which people "like us" will want to set in `.Rprofile`.
  * In an interactive session, prompts the user to pick SSH or HTTPS and then remembers it for the session. Also suggests setting the option.
  * In a non interactive session, the option is consulted but it will fall back to "ssh".


jennybc:
@Hadley I don't think this really needs review and am going to merge so I can keep going. But if you see anything, comment and I will address.

R/git.R
```diff
@@ -132,6 +132,102 @@ use_git_config <- function(scope = c("user", "project"), ...) {
...
+#' @return "ssh" or "https"
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' use_git_protocol()
+#' use_git_protocol("ssh")
+#' use_git_protocol("https")
+#' }
+use_git_protocol <- function(protocol = NULL) {
```
hadley: My only concern is that `use` might not be the right verb here. Isn't it more like `discover` or `determine`?

jennybc: You really can call this proactively as `use_git_protocol("https")`, though.

In my mind, in a workshop setting, we might urge everyone to do that right before we try something like `use_github()`. Or I could see using it that way in docs or a non-interactive script.

But IRL, yeah, you would normally just let the menu happen to you upon first need. Or set the option.

I don't feel strongly except that `discover` or `determine` don't reflect the direct usage that is part of the design.

R/git.R
```diff
@@ -132,6 +132,102 @@ use_git_config <- function(scope = c("user", "project"), ...) {
...
 }
 
+#' Choose git protocol and credential defaults
+#'
+#' Decide whether to use the SSH or HTTPS protocol for the current session.
+#' This affects two things:
+#'   * The URL format of newly configured git remotes:
+#'     - `protocol = "ssh"` implies `git@@github.com:<OWNER>/<REPO>.git`
+#'     - `protocol = "https"` implies `https://github.com/<OWNER>/<REPO>.git`
+#'   * The strategy for creating default credentials when none are given.
```
hadley: Is this true? Why aren't we using the same protocol as the repository?

R/git.R
```diff
@@ -132,6 +132,102 @@ use_git_config <- function(scope = c("user", "project"), ...) {
   }
 }
 
+#' Choose git protocol and credential defaults
```
hadley: Would this be better as "choose default git protocol"? Shouldn't it only apply to new repositories that don't already have a protocol?

R/git.R
hadley: Ah ok, in that case it's confusing that we also use it internally to _retrieve_ the current protocol.

R/git.R
jennybc: Sure ok. Done in master.

R/git.R
jennybc: I'm just trying to explain the significance of the git protocol, i.e. how it affects usethis's actions.

This is getting reworked as I add similar functionality for `credentials`. So you will get a proper chance to review that PR.

PR #653, opened 2019-03-13
Create a way to configure git protocol and git2r credentials

jennybc:
Gives us:

  * `git_protocol()` / `use_git_protocol()` to summon or set a default protocol (so, either SSH or HTTPS)
  * `git2r_credentials()` / `use_git2r_credentials()` to summon or set git2r credentials

This addresses your earlier distaste @Hadley for getting and setting `protocol` with the same function, i.e. I've split it out and followed same design for `credentials`.

Clears the way for me to tackle #533 Build protocol flexibility into `pr_*()` functions


R/git.R
```diff
@@ -132,70 +132,70 @@ use_git_config <- function(scope = c("user", "project"), ...) {
...
+#' ## explicitly set the protocol
 #' use_git_protocol("ssh")
 #' use_git_protocol("https")
 #' }
-use_git_protocol <- function(protocol = NULL) {
-  if (!is.null(protocol)) {
-    stopifnot(
-      length(protocol) == 1,
-      tolower(protocol) %in% c("ssh", "https", NA)
+git_protocol <- function() {
```
jennybc: When I tackle the `pr_*()` stuff, this will presumably gain some smarts re: consulting the "current repo" if such a thing is well-defined. I think that will come into focus then.

jennybc: In the existing use cases (`use_github()` and `create_from_github()`) there is no existing local <--> remote relationship to consult re: protocol. But that will not be true for `pr_*()` functions.

hadley: I think we might also be able to add a bit more smarts with @jeroen's credentials package — like if an ssh key exists, we could assume that the protocol should be ssh.

R/git.R
```diff
@@ -231,6 +238,123 @@ choose_protocol <- function() {
...
+#'   given to any `auth_token` that is passed explicitly. Otherwise, the
+#'   environment variables `GITHUB_PAT` and `GITHUB_TOKEN` are consulted, in
+#'   that order. If a PAT is found, we make an HTTPS credential with
+#'   [git2r::cred_user_pass()]. The PAT is sent as the password and dummy text
+#'   is sent as the username (the PAT is what really matters in this case). You
+#'   can also register an explicit credential yourself in a similar way:
+#' ```
+#' my_cred <- git2r::cred_user_pass(
+#'   username = "janedoe",
+#'   password = "super_secret_password"
```
hadley: I think we need to be very careful with this advice because if you run this code in your current session your password will get saved to `.Rhistory`, which you might accidentally expose.

R/git.R
jennybc: Yeah I'm trying to do this in a way that will make fitting that package in a fairly pleasant task. It's on my radar.

R/git.R
jennybc: I could modify the example to show `Sys.getenv()` or `getPass::getPass()` as the way of handling the password field.

hadley: I think `getPass::getPass()` would be good

R/git.R
```diff
@@ -132,70 +132,70 @@ use_git_config <- function(scope = c("user", "project"), ...) {
...
 
-#' Choose default git protocol and credential defaults
+#' Produce or register git protocol
 #'
-#' Decide whether to use the SSH or HTTPS protocol for the current session.
-#' This affects two things:
-#'   * The URL format of newly configured git remotes:
+#' Git operations that address a remote use a so-called "transport protocol".
+#' usethis supports SSH and HTTPS. The protocol affects two things:
+#'   * The default URL format:
```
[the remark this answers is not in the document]
jennybc: Well, they're not always "clones" (consider: `use_github()`) but ok yes reworded.

R/git.R
jennybc: OK. Used for the analogous content re: `cred_ssh_key()` and the passphrase.

jennybc:
@cderv @ijlyttle With this merged, I *think* we are going to have much more success with HTTPS and on Windows, because I make a git2r credential from the GitHub PAT whenever possible. Now I will use this to wire this all in to the PR flow.

But starting now, it would be great for you two to test drive this as much as possible. It's nearly impossible to test. I do some manual tests but I don't cover many scenarios.

cderv:
First try with a ˋpr_fetch` and first error. Something does not work as intended with git2r::fetch and credentials. 
I need to dig a little further to isolate the issue.

ijlyttle:
So far, I have succeeded, no problems, on macOS with: `pr_init()`, `pr_push()`, and `pr_fetch()`.

I'll use more of the functions today and try things with Windows.

cderv:
ok my bad, this is working now with last pr updates. Will continue to test and open specific issues if needed.

PR #656, opened 2019-03-15
pr_*() love for https

### LINKED ISSUE #533: Build protocol flexibility into pr_*() functions

jennybc:
Closes #533 Build protocol flexibility into pr_*() functions

Gives the `pr_*()` functions a proper understanding of:

  * Git protocol. Specifically, we acknowledge the existence of HTTPS now 😁
  * git2r credentials. I can't imagine anyone wanting to provide them directly but the most internally consistent behaviour is to make this possible.
  * PAT a.k.a. `auth_token`. Its importance is more clear now and we're more assertive about checking for existence/validity.

I'm still doing manual tests and expect to find things to fix. But input from anyone, including @cderv and @ijlyttle, would be very welcome now.


NAMESPACE
```diff
@@ -7,6 +7,7 @@ export(browse_github)
 export(browse_github_issues)
 export(browse_github_pat)
 export(browse_github_pulls)
+export(browse_github_token)
```
jennybc: We always call this `auth_token`, so I've created this. The old `browse_github_pat()` is aliased to this. These are workflow-y functions, so this can be a pretty low key thing.

NAMESPACE
```diff
@@ -24,6 +25,7 @@ export(git2r_credentials)
 export(git_protocol)
 export(git_sitrep)
 export(git_vaccinate)
+export(github_token)
```
jennybc: Newly exported for visibility. Also, it's a frequent default value for `auth_token` and I don't like to use unexported functions in a public signature.

R/git-utils.R
```diff
@@ -152,26 +154,41 @@ git_branch_compare <- function(branch = git_branch_name()) {
   )
 }
 
-git_branch_push <- function(branch = git_branch_name(), force = FALSE) {
-  remote <- git_branch_tracking(branch)
-  if (is.null(remote)) {
```
jennybc: All this logic had to be pulled out into `git_branch_remote()` so I know / can guess the name of the remote associated with current branch. And I need that to look up the URL. And I need that to see if the current repo is managed with SSH or HTTPS.

R/github-utils.R
```diff
@@ -16,7 +34,7 @@ github_owner <- function() {
   github_origin()[["owner"]]
 }
 
-github_source <- function() {
+github_owner_upstream <- function() {
```
jennybc: This name makes much more sense to me.

R/github-utils.R
```diff
@@ -55,15 +73,7 @@ parse_github_remotes <- function(x) {
   lapply(match, function(y) list(owner = y[[2]], repo = y[[3]]))
 }
 
-## use from gh when/if exported
-## https://github.com/r-lib/gh/issues/74
-github_token <- function() {
```
jennybc: Moved to `github.R` because this is no longer just an internal utility.

R/pr.R
```diff
@@ -74,21 +80,35 @@ pr_init <- function(branch) {
...
 #' \dontrun{
 #' ## scenario: current project is a local copy of fork of a repo owned by
 #' ## 'tidyverse', not you
 #' pr_fetch(123, owner = "tidyverse")
 #' }
-pr_fetch <- function(number, owner = NULL) {
+pr_fetch <- function(number,
+                     owner = NULL,
+                     credentials = NULL,
+                     auth_token = github_token()) {
```
jennybc: In general, the `pr_*()` functions all gained the args `credentials` and `auth_token`, though I can't imagine actually using them. Anyone who uses these functions heavily will want to have a setup that plays well with default behaviour.

R/git-utils.R
```diff
@@ -152,26 +154,41 @@ git_branch_compare <- function(branch = git_branch_name()) {
...
 
+  remote <- paste0(remote_name, ":", remote_branch)
   ui_done("Pushing local {ui_value(branch)} branch to {ui_value(remote)}")
   git2r::push(
     git_repo(),
     name = remote_name,
     refspec = glue("refs/heads/{branch}:refs/heads/{remote_branch}"),
-    force = force
+    force = force,
+    credentials
```
hadley:
```suggestion
    credentials = credentials
```

R/github-utils.R
hadley: We have `pr_pull_source()`. Should this be `github_owner_source()`? Or `github_source_owner()`?

R/github.R
```diff
@@ -181,49 +178,77 @@ use_github_links <- function(auth_token = NULL,
...
-  url <- sprintf(
-    "%s/settings/tokens/new?scopes=%s&description=%s",
-    host, scopes, description
+  url <- glue(
+    "{host}/settings/tokens/new?scopes={scopes}&description={description}"
   )
   view_url(url)
 
-  ui_todo("Call {ui_code('usethis::edit_r_environ()')} to open {ui_path('.Renviron')}")
+  ui_todo(
```
hadley: Why don't we just open this file?

R/github.R
```diff
@@ -181,49 +178,77 @@ use_github_links <- function(auth_token = NULL,
...
+#' @rdname browse_github_token
+#' @export
+browse_github_pat <- browse_github_token
+
+#' @rdname browse_github_token
+#' @export
+#' @examples
+#' # for safety's sake, just reveal first 4 characters
+#' substr(github_token(), 1, 4)
+github_token <- function() {
```
hadley: I wonder if this should print a message (authenticating with GitHub using personal access token) and cleanly error if it can't find a token.

R/pr.R
hadley: I would rather not add these arguments, if the only argument is for consistency. I think we should be moving away from these arguments in general, in favour of better tooling for declaring credentials globally.

R/pr.R
```diff
@@ -74,22 +80,37 @@ pr_init <- function(branch) {
...
 
-  ui_done("Retrieving data for PR #{number}")
-  pr <- gh::gh("GET /repos/:owner/:repo/pulls/:number",
-    owner = owner %||% github_source() %||% github_owner(),
-    repo = github_repo(),
-    number = number
+  owner <- owner %||% github_owner_upstream() %||% github_owner()
+  repo <- github_repo()
+  pr_string <- glue("{owner}/{repo}/#{number}")
+  ui_done("Retrieving data for PR {ui_value(pr_string)}")
```
hadley: Having two messages for this one action seems a bit heavy to me.

R/pr.R
```diff
@@ -161,34 +197,46 @@ pr_push <- function() {
...
 
+  protocol <- github_remote_protocol()
+  credentials <- credentials %||% git2r_credentials(protocol, auth_token)
+
   source <- if (git_is_fork()) "upstream/master" else "origin/master"
+  ## TODO: this claim seems a bit like wishful thinking ... what assumptions
+  ## are we making about the setup here?
   ui_done("Pulling changes from GitHub source repo {ui_value(source)}")
-  git_pull(source)
+  ## TODO: why are we not passing `source`?
```
hadley: Why aren't you?

R/github-utils.R
jennybc: You renamed `pr_pull_source()` to `pr_pull_upstream()` in a separate PR (which I merged very recently):

Rename `pr_pull_source()` to `pr_pull_upstream()` #617

R/pr.R
jennybc: Yeah, I can see that and what you say is now possible with via `use_git2r_credentials()`. I guess it is also possible for the `auth_token` in a more indirect way, i.e. via the env vars. So we can remove them and make sure error messages / docs tell people what to do if default behaviour doesn't work for them.

R/github.R
jennybc: I'll look around and see if we are passing/using it in places where it is optional vs. mission critical.

R/pr.R
jennybc: It looks really nice though 🙂. I can combine them into 1, issued after the GET, so the interesting content is there.

R/github-utils.R
hadley: Oh I guess I must be using an old version locally 🤪

R/github.R
jennybc: This is pretty muddy right now. I will open a separate issue for it because it feel like its own mini-problem.

jennybc: #657 Where to check validity of GitHub PAT?

R/github.R
jennybc: Somehow it feels a bit like overstepping but I don't care deeply.

R/pr.R
jennybc: OK I reworked this so that we are more clearly doing what we say we are doing. This function strikes me as something people might surprise themselves with (i.e. pulling master into whatever branch they are on). But I guess that is its point.

jennybc:
I propose we merge this @Hadley and assume that our usage and that of, e.g., @cderv and @ijlyttle, will alert us to loose ends. It will get exercised more by people who are not us in `master`.

PR #661, opened 2019-03-19
Remove leading dots in extension arguments to `fs::file_temp()`

jimhester:
With the current CRAN release of fs, passing a leading dot results in a
doubled dot.

    path_file(file_temp(ext = ".txt"))
    #> file4c8b5cc4167..txt
    path_file(file_temp(ext = "txt"))
    #> file4c8bb172736.txt

However this will be fixed by the upcoming release of fs, so that any leading
dots are ignored, so `ext = ".txt` and `ext = "txt"` will be equivalent.

    path_file(file_temp(ext = ".txt"))
    #> file4c8b24d9d42a.txt
    path_file(file_temp(ext = "txt"))
    #> file4c8b58f8244c.txt

Nonetheless I think it is best practice to supply the extension without
a dot, as I am pretty sure this was just a mistake when porting the
usethis code from using `tempfile()`.

### DIFF
--- a/tests/testthat/test-use-course.R
+++ b/tests/testthat/test-use-course.R
@@ -128,19 +128,19 @@ test_that("top_directory() identifies a unique top directory (or not)", {
 })
 
 test_that("tidy_unzip() deals with loose parts, reports unpack destination", {
-  tmp <- file_temp(ext = ".zip")
+  tmp <- file_temp(ext = "zip")
   file_copy(test_file("yo-loose-regular.zip"), tmp)
   dest <- tidy_unzip(tmp)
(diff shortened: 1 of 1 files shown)


jimhester:
I uncovered this while running revdepcheck results for fs, and this bug in the usethis tests caused them to fail due to other changes in the fs release. However now that I have made fs more robust to leading `.` in the ext argument the tests pass with both versions of fs regardless of this change. But I still think it would be useful to fix.

jimhester:
Oh I just realized when going through the news for fs that I changed the behavior with the `ext` extension in `file_temp()` in the devel version. So this was working properly with the CRAN version. So maybe I will just close this for now.

Issue #664, opened 2019-03-22
Could create_from_github() use rstudioapi::selectDirectory()?


ijlyttle:
I am finding it to be a friction point to find, then type, the `destdir` argument to `create_from_github()`.

If `destdir` is `NULL` and `create_from_github()` is called interactively (and if using RStudio), could the behavior be changed to invoke `rstudioapi::selectDirectory()`?

Happy to make a PR, if you like.

PS - could this idea be extended to `create_project()`, `create_package()`, `use_course()`, or others?

jennybc:
In discussion in #665, we decided to not pursue this at this time.

ijlyttle:
I appreciate you all considering this, thanks!

PR #665, opened 2019-03-22
filesystem-dialog for create_from_github()

ijlyttle:
This PR addresses #664, to allow you to specify `destdir` using the filesystem dialog, for `create_from_github()`. 

At this point, just a first pass to see if there are any objections or other initial reactions.

If this were extended to `create_project()`, `create_package()`, or `use_course()`, I could see extracting this into its own internal-function.

I'm assuming that as a draft PR, you can react - if not, I'll be happy to change it to a "real" PR

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -144,8 +144,8 @@ create_from_github <- function(repo_spec,
 
     choice <- utils::menu(
       choices = c(
-        "filesystem dialog",
-        glue::glue("default location: {ui_value(conspicuous_place())}")
+        "Filesystem dialog",
+        glue::glue("Default location: {ui_value(conspicuous_place())}")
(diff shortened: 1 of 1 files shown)


jennybc:
I would not have initiated this, but I'm also fine with it.

I agree that this should be abstracted into a helper and used in multiple places, if it happens.

@Hadley will you weigh in?

hadley:
My gut feeling is negative — I think this changes the semantics of the existing functions too much, and hence requires a decent amount of consideration, and I don't think the improvement is sufficient for us to spend that time.

jennybc:
Ok then let’s set this aside. Thanks for the test balloon @ijlyttle.

PR #669, opened 2019-03-27
Include git_sitrep() in setup article

### LINKED ISSUE #668: use_git_config() behavior inconsistent with article
**Edit**: I did not look at this carefully enough; as @cderv points out, this is very likely a documentation issue.

---

As best as I can tell, this was working in usethis 1.4.0:

https://github.com/r-lib/usethis/blob/f08ecb6480310edd0e1c9b24e1d55359056804ec/vignettes/articles/usethis-setup.Rmd#L72-L73

I updated one of my machines to the dev version, and it returns without printing anything to the screen.

The operative code in usethis seems to be in an internal function `git_config()`, which then calls `git2r::config()`:

https://github.com/r-lib/usethis/blob/37a774ea989714619f5c6dcfaff0ece248027b56/R/git-utils.R#L297-L310 

When I updated my usethis, it updated git2r 0.24.0 -> 0.25.2

Happy to dig further, but wanted to surface this now in case you had some ideas already.

Thanks!

ijlyttle:
Fix #668 

Ready-to-go from my perspective, but happy to make any changes you may propose

### DIFF
--- a/vignettes/articles/usethis-setup.Rmd
+++ b/vignettes/articles/usethis-setup.Rmd
@@ -74,7 +74,7 @@ use_git_config(user.name = "Jane Doe", user.email = "jane@example.com")
 git_sitrep()
 ```
 
-`usethis::use_git_config()` helps you configure your `user.name` and `user.email`. Substitute **your name** and **your email address**. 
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks!

PR #674, opened 2019-03-30
Refactor handling of GitHub token

### LINKED ISSUE #657: Where to check validity of GitHub PAT?
Look at `github_token()`, `github_user()`, `check_github_token()`. Is there a way to rationalize?

  * Why do we return `""` if no token found vs. `NULL` or erroring?
    - One answer is that this is what `gh::gh_token()` does. Which is exported only in the dev version ... hence the version here.
    - Another answer is that in some of our uses we don't absolutely need the token, i.e. we're GETting a public resource.
  * @Hadley asks "Should `github_token()` print a message (authenticating with GitHub using personal access token) and cleanly error if it can't find a token?" [here](https://github.com/r-lib/usethis/pull/656#discussion_r266192898).
    - That basically describes current behaviour of `check_github_token()`.

jennybc:
Closes #657 Where to check validity of GitHub PAT?

Notes to self are in the issue. Highlights:

  * Put definitions and checks closer to where they are used.
  * `""` is the PAT when we have no PAT. This is expedient, because it's what gh expects.
  *  We have two levels of PAT checking:
    - ~`validate_github_token()`~ `check_github_token()` does a rigorous check, including against the GitHub API. Use this when we know we truly need a token or git credentials.
    - ~`maybe_github_token()`~ `check_github_token(allow_empty = TRUE)` tolerates `""` but, otherwise, does full checks. Use this when we can get away with having no token or credentials, i.e. reading a public resource.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -241,24 +241,21 @@ check_not_nested <- function(path, name) {
 }
 
 rationalize_fork <- function(fork, repo_info, auth_token) {
-  perms <- repo_info$permissions
-  owner <- repo_info$owner$login
+  have_token <- have_github_token(auth_token)
+  can_push <- isTRUE(repo_info$permissions)
+  repo_owner <- repo_info$owner$login
+  user <- if (have_token) github_user(auth_token)[["login"]]
 
   if (is.na(fork)) {
-    fork <- have_github_token(auth_token) && !isTRUE(perms$push)
+    fork <- have_token && !can_push
   }
 
-  if (fork && !have_github_token(auth_token)) {
+  if (fork && !have_token) {
     ## throw the usual error for bad/missing token
     validate_github_token(auth_token)
   }
 
-  user <- if (have_github_token(auth_token)) {
-    github_user(auth_token)[["login"]]
-  } else {
-    NULL
-  }
-  if (fork && identical(user, owner)) {
+  if (fork && identical(user, repo_owner)) {
     ui_stop(
       "Repo {ui_value(repo_info$full_name)} is owned by user \\
       {ui_value(user)}. Can't fork."
--- a/R/create.R
+++ b/R/create.R
@@ -150,9 +150,7 @@ create_from_github <- function(repo_spec,
   create_directory(repo_path)
   check_directory_is_empty(repo_path)
 
-  pat_available <- auth_token != ""
-  user <- if (pat_available) github_user()[["login"]] else NULL
-  credentials <- credentials %||% git2r_credentials(protocol, auth_token)
+  auth_token <- maybe_github_token(auth_token)
 
   gh <- function(endpoint, ...) {
     gh::gh(
@@ -165,7 +163,7 @@ create_from_github <- function(repo_spec,
 
   repo_info <- gh("GET /repos/:owner/:repo", owner = owner, repo = repo)
 
-  fork <- rationalize_fork(fork, repo_info, pat_available, user)
+  fork <- rationalize_fork(fork, repo_info, auth_token)
   if (fork) {
     ## https://developer.github.com/v3/repos/forks/#create-a-fork
     ui_done("Forking {ui_value(repo_info$full_name)}")
@@ -175,7 +173,8 @@ create_from_github <- function(repo_spec,
       ssh = repo_info$ssh_url
     )
     repo_info <- gh(
-      "POST /repos/:owner/:repo/forks", owner = owner, repo = repo
+      "POST /repos/:owner/:repo/forks",
+      owner = owner, repo = repo
     )
   }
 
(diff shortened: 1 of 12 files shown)


R/github.R
```diff
@@ -305,17 +305,54 @@ check_no_github_repo <- function(owner, repo, host, auth_token) {
...
     ui_stop(c(
-      "No GitHub {ui_code('auth_token')}.",
-      "Provide explicitly or make available as an environment variable.",
-      "See {ui_code('browse_github_token()')} for help setting this up."
+      msg,
+      "See {ui_code('browse_github_token()')} for help storing a token as an environment variable."
     ))
   }
+
+  if (length(auth_token) != 1) {
```
hadley: This feels a little over-engineered to me. I think `if(!is.character(auth_token) || length(auth_token) != 1)` would be adequate.

R/github.R
```diff
@@ -305,17 +305,54 @@ check_no_github_repo <- function(owner, repo, host, auth_token) {
...
+## tolerate "", but explicitly validate the token otherwise
+maybe_github_token <- function(auth_token = github_token()) {
+  if (!have_github_token(auth_token)) {
+    return("")
+  }
+  validate_github_token(auth_token)
+}
+
+## rigorous checks, including against the GitHub API
+validate_github_token <- function(auth_token = github_token()) {
```
hadley: I think this should be `check_github_token()` for consistency with the other `check_` functions (which throw an error)

R/github.R
```diff
@@ -305,17 +305,54 @@ check_no_github_repo <- function(owner, repo, host, auth_token) {
...
+
+# github token helpers ----------------------------------------------------
+
+## minimal check: token is not ""
+have_github_token <- function(auth_token = github_token()) {
+  !isTRUE(auth_token == "")
+}
+
+## tolerate "", but explicitly validate the token otherwise
+maybe_github_token <- function(auth_token = github_token()) {
```
hadley: I think you should eliminate this function in favour of an additional argument, e.g. `check_github_token(allow_empty = TRUE)`

R/create.R
```diff
@@ -240,22 +240,27 @@ check_not_nested <- function(path, name) {
   invisible()
 }
 
-rationalize_fork <- function(fork, repo_info, pat_available, user = NULL) {
-
+rationalize_fork <- function(fork, repo_info, auth_token) {
```
hadley: I think this function would be a little clearer if you stored `have_github_token(auth_token)` in a variable

R/create.R
```diff
@@ -240,22 +240,27 @@ check_not_nested <- function(path, name) {
...
   owner <- repo_info$owner$login
 
   if (is.na(fork)) {
-    fork <- pat_available && !isTRUE(perms$push)
+    fork <- have_github_token(auth_token) && !isTRUE(perms$push)
   }
 
-  if (fork && !pat_available) {
-    check_github_token(auth_token = NULL)
+  if (fork && !have_github_token(auth_token)) {
```
hadley: I wonder if this logic would be clearer if the outer condition tested for the presence/absence of the `auth_token`

R/github.R
```diff
@@ -305,17 +305,54 @@ check_no_github_repo <- function(owner, repo, host, auth_token) {
...
   }
+
+  if (length(auth_token) != 1) {
+    local_stop("GitHub {ui_code('auth_token')} must be a single string.")
+  }
+  if (!is_string(auth_token)) {
+    bad_class <- paste(class(auth_token), collapse = "/")
+    local_stop("GitHub {ui_code('auth_token')} must be character, not {bad_class}.")
+  }
+  if (isTRUE(auth_token == "")) {
```
hadley: I think `identical(auth_token, "")` expresses this intent more clearly

R/pr.R
```diff
@@ -93,11 +93,7 @@ pr_fetch <- function(number,
   check_uses_github()
   check_uncommitted_changes()
 
-  auth_token <- github_token()
-  check_github_token(auth_token)
-
-  protocol <- github_remote_protocol()
-  credentials <- git2r_credentials(protocol, auth_token)
+  auth_token <- maybe_github_token()
```
hadley: I think this should be moved to the arguments for consistency with the other functions

R/pr.R
```diff
@@ -322,7 +320,7 @@ pr_find <- function(owner,
     owner = owner,
     repo = repo,
     .limit = Inf,
-    .token = github_token()
+    .token = maybe_github_token()
```
hadley: Again, should be put in arguments, for consistency

R/pr.R
jennybc: Two weeks ago you said:
> I would rather not add these arguments, if the only argument is for consistency. I think we should be moving away from these arguments in general, in favour of better tooling for declaring credentials globally.

and I agree with Past Hadley 😉

https://github.com/r-lib/usethis/pull/656#discussion_r266192941

R/github.R
jennybc: Yeah, this was a leftover from experimenting with two different levels of checks.

Issue #572, opened 2019-01-16
use_template and devtools::load_all() docs?


maelle:
Reg https://github.com/r-lib/usethis/issues/347 are you still interested in "a PR that adds a small note about this to the use_template() docs, to reduce future head-scratching."?

At least since reading that thread I know I'm not crazy. :grin:

jennybc:
@maelle Now would be a really good time for this ⏰ we're pushing towards a release.

maelle:
Oops I had missed your answer in January, on it now!

Issue #347, opened 2018-05-11
use_template doesn't work as intended when inside a package


lwjohnst86:
I've tried to use a reprex but the effects are as intended. But when the `use_template` is inside of a package, it can't find the template... debugging it seems to point to `system.file()` using the location that the developing package is... and not being smart enough to know that "templates" is found in the "inst" folder. 

While in debug mode inside `usethis:::find_template`, running `system.file("inst", "templates", ..., package = ...)` correctly points to the template... but without the "inst" (as it currently is) is returns a character vector of length 0.

I struggled to understand this issue for several hours, and I don't really know what the correct effect `use_template` should have here... Not sure if this is a bug, or if it is an intended effect. Either way, just wanted to report it, since I can't actually use `use_template` because of this ... bug? Even though `use_template` is exactly what I am looking for... Oh well.

ijlyttle:
Please forgive me for what may be a silly question, but have you installed the package from which you want to `use_template()`, i.e. hit the "Build & Install" button on the RStudio IDE?

lwjohnst86:
Haha, yes, a silly question, but analogous to the "did you restart the computer" for IT :P Yes, its a package I am developing, so I've done all the "load_all", "build and install" things I could and it doesn't do anything. Problem even occurs on travis.

jennybc:
I believe @ijlyttle is already using `use_template()` in a package (yes?), so maybe he can point at his source to help figure out if there's a difference in usage.

ijlyttle:
Thanks for your good humor on the question 😀

@jennybc responded as I was typing - I am using the CRAN version of `usethis` and it is working OK for me. The package for which I `use_template()` is ghentr, and you can see my templates folder here: https://github.com/ijlyttle/ghentr/tree/master/inst/templates

Having installed the package (using `install_github("ijlyttle/ghentr")`), these work for me:

```
system.file("templates", "install_ghe.R", package = "ghentr")
usethis:::find_template("install_ghe.R", package = "ghentr")
```

Does this shed any light?

ijlyttle:
I think it is doing the same as for you: 

```R
> system.file("templates", "install_ghe.R", package = "ghentr")
[1] "/Library/Frameworks/R.framework/Versions/3.4/Resources/library/ghentr/templates/install_ghe.R"
> usethis:::find_template("install_ghe.R", package = "ghentr")
[...]

PR #684, opened 2019-04-01
Reduce future use_template()/load_all() head-scratching

maelle:
cf #347 #572 

sorry for the delay!

### DIFF
--- a/R/template.R
+++ b/R/template.R
@@ -13,6 +13,16 @@
 #' the `data` argument. Internally, this function uses
 #' [whisker::whisker.render()] to combine your template file with your data.
 #'
+#' @details When using `use_template()` inside your package, you won't be able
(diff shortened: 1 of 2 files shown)


jennybc:
Reviewing this PR made me ask internally whether we could just fix the root problem and @jimhester did!

As of https://github.com/r-lib/fs/commit/de56a640e874b642d002108d4bb3f67f89bf7843, `fs::path_package()` is a better alternative to `system.file()`.

Locally, you should be able to enjoy the benefits of this already once you install the dev version of fs.

I can't use it unconditionally here in usethis yet, because this fs feature isn't in the CRAN version and I want usethis (including our **own** dev workflow) to work with CRAN fs.

maelle:
Best possible outcome! :tada:

PR #691, opened 2019-04-03
UI changes

jennybc:
A few comments inline


R/ui.R
```diff
@@ -70,7 +86,7 @@ ui_code_block <- function(x, copy = interactive(), .envir = parent.frame()) {
   x <- glue(x, .envir = .envir)
 
   block <- indent(x, "  ")
-  block <- crayon::make_style("darkgrey")(block)
+  block <- crayon::silver(block)
```
jennybc: If I can't have a decent red, then you can't have dark grey 😈

More seriously, is there some principle that makes this use of `crayon::make_style()` acceptable but not others?

jennybc:
Screen Shot 2019-04-03 at 11 25 23 AM

R/ui.R
hadley: No 😐

PR #693, opened 2019-04-04
Record "created-by" info in git config; use it to offer remote deletion

jennybc:


R/pr.R
```diff
@@ -272,6 +281,25 @@ pr_finish <- function() {
...
+  created_by <- git_config_get(glue("remote.{remote}.created-by"))
+  if (is.null(created_by) || !grepl("^usethis::pr_", created_by)) {
+    return()
+  }
+
+  b <- git2r::branches(git_repo(), flags = "local")
+  remote_specs <- purrr::map(b, ~ git2r::branch_get_upstream(.x)$name)
+  remote_specs <- purrr::compact(remote_specs)
+  remote_specs <- purrr::keep(remote_specs, ~ grepl(glue("^{remote}/"), .x))
+  if (length(remote_specs) == 0 &&
```
hadley: I don't think you even need to ask here. We didn't ask if it was ok to create so we shouldn't ask if it's ok to delete. I don't think most people will know what the "correct" answer is here, and I think it's bad practice to ask the user to adjudicate (it's like when the plumber asked me if I wanted copper or plex piping).

jennybc: Yeah. I'll have to think about whether there's any possible way we'd delete a remote that will be needed. Like `upstream` or `master`. It's easy to write these functions assuming someone got here "the right way", using the `pr_*()` functions as designed. But I suspect real usage will be more chaotic.

Issue #694, opened 2019-04-04
assert conspicuous-place for RStudio Cloud


ijlyttle:
Follow-on from #685:

> This is adjacent to #466, but I wonder what should be the default "conspicuous place" on RStudio Cloud. Currently, it seems to be `home/rstudio-user`, which I think is meant for dotfiles and the like. I wonder if the default conspicuous-place for RStudio Cloud should be `/cloud/project`.

To me, the "simplest" solution would be to add `path_real("/cloud/project")` to the top of the vector of paths, here:

https://github.com/r-lib/usethis/blob/274902535ea489dbaa4b57e56ad8c6d81e62a988/R/course.R#L290-L297

Questions:

- this could introduce a problem if someone happens to have that path, but not be on RStudio Cloud, but I don't know how to ask `is_this_rstudio_cloud()`. Thoughts?

- would `path_real()` be the right path-creator?

I will start a PR along these lines, subject to your guidance.

jennybc:
Notes from Slack conversation with the Cloud Team:

  * Current best way to detect that we're on Cloud: "keying off of the existence of `/cloud/project`".
  * Project-switching (while in a Cloud project) goes against the RStudio Cloud paradigm. They generally think of 1-1 relationship between an RStudio Cloud project and an RStudio Project. And the current Cloud project is the one and only project.
  * Most logical location for `use_course()` to write to is within the project, i.e. within `/cloud/project`. However, it is not logical to attempt to *open* any `.Rproj` file found within the payload. This sort of project-switching goes against the RStudio Cloud paradigm.
  * The use of `create_*()` functions is somewhat illogical within RStudio Cloud. It's conceivable we just shouldn't do it? If we do it anyway, it's not blindingly obvious whether the target dir should be below `/home/rstudio-user` or `/cloud/project`. Again, it would not make sense to attempt to activate the new project.

jennybc:
cc @mine-cetinkaya-runde @cwickham @hadley l in case you have any relevant experience or thoughts.

Have you taught something where you used or wanted to use `create_*()` functions on Cloud? It seems like a clash of paradigms, i.e. if you're designing a learning experience you would either provide the scaffold via a Cloud Project or participants would use their own computer (or an RStudio server instance) + `create_*()`.

jennybc:
let me try that again: cc @mine-cetinkaya-rundel

ijlyttle:
My use case, which may not be typical, revolves around a package-development workshop followed by an unconf. 

[...]

PR #698, opened 2019-04-04
Address #694: Advice for paths and RStudio Cloud

ijlyttle:
This is a first attempt at providing guidance for how to deal with paths and destination-directories for RStudio Cloud. 

I don't know if this is best addressed in the function-documentation, as done here, or if a bullet point in the setup-article might also (or instead) be merited.

cc @cwickham, to make sure I've got it "right"

### DIFF
--- a/R/course.R
+++ b/R/course.R
@@ -26,6 +26,9 @@
 #'   mind, but should work for ZIP files generally. If no "http" prefix is
 #'   found, "https://" is prepended. See [use_course_details] for more.
 #' @param destdir The new folder is stored here. Defaults to user's Desktop.
+#'   If you are using  [RStudio Cloud](https://rstudio.cloud), our current
+#'   advice is to use `"/cloud/project"` as a home directory, and to specify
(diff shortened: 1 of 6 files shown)


cwickham:
This is what I do, but I'm not sure how I feel about this advice.  I feel like it should convey something more like "If you are using RStudio Cloud you might not want to use `use_course()`, but if you must, we recommend `"/cloud/project"` as a home directory."

Or perhaps, not have this in the documentation, but somewhere else as "problems encountered in teaching usethis".

ijlyttle:
@cwickham Thanks - I'll have another go at it.

I imagine this will keep until after the CRAN release, esp. to decide where the documentation should "live".

jennybc:
I suspect there should be an article about usethis functions that are useful when running a course or workshop. And there could be a section on RStudio Cloud.

ijlyttle:
Sounds good! Will head in that direction.

jennybc:
I'm going to close this, with the expectation that this new content (and probably some of the existing) can move into this future article.

PR #703, opened 2019-04-05
Add use_circleci()

jdblischak:
I wrote a function `use_circleci()` to setup continuous integration with [CircleCI](https://circleci.com/). I modeled it after `use_travis()`.

## Features

The configuration file it creates enables the following features:

* Uses the Docker image `rocker/verse:latest` ([DockerHub](https://hub.docker.com/r/rocker/verse)) as the environment for building and checking the package. This image contains TeX Live, pandoc, and the tidyverse packages, so it should be sufficient for many R packages. However, the user has the option of specifying the Docker image to use via the argument `image`.

* Very simple caching of packages. The CircleCI platform provides a [complex caching mechanism](https://circleci.com/docs/2.0/caching/), but I chose not to use this since `devtools::install_deps()` can update packages as needed. Also, the default is for the cache to expire after 30 days. If desired, a more sophisticated option would be to invalidate the cache whenever a specific file is changed, e.g. using a key like `cache-{{ checksum "DESCRIPTION" }}`.

* The `*.Rcheck/` directory is uploaded as an artifact at the end of build.

## Other notes

* The tests I added passed on my local machine and on the [AppVeyor](https://ci.appveyor.com/project/jdblischak/usethis/builds/23616082#L289) and [Travis](https://travis-ci.org/jdblischak/usethis/jobs/516014236#L1239) builds of my forked repository
* I couldn't find the function `ui_` that is used in `check_uses_travis()` ([line 67](https://github.com/r-lib/usethis/blob/master/R/ci.R#L67)), so I used `ui_field()` in `check_uses_circleci()`
* If this new function is accepted, I'll add a bullet point to `NEWS.md`


jennybc:
I'm releasing usethis today, so it's about to rest for a little while. But we anticipate another release before useR!

jennybc:
Hi @jdblischak I used this PR as a demo of `pr_*()` functions in usethis during our group meeting today, which is why you see this odd little commit from me.

jdblischak:
> I used this PR as a demo of pr_*() functions in usethis during our group meeting today, which is why you see this odd little commit from me.

@jennybc Understood. Thanks!

jdblischak:
Some potential ideas to improve the default configuration (with the downside of increased complexity):

## Report installation information

Similar to the current setup for Travis or Appveyor, we could add a step that reports the installation information, e.g.

```
      - run:
          name: Session information and installed package versions
          command: |
            Rscript -e 'sessionInfo()'
            Rscript -e 'installed.packages()[, c("Package", "Version")]'
            Rscript -e 'rmarkdown::pandoc_version()'
```

## Separate cache for each job

If a user tests multiple versions of R in different jobs, they won't want to use the same cache. We could make the default caching mechanism slightly more complicated by default to anticipate this use case. The following will store a separate cache for each job:

```
    steps:
      - restore_cache:
          keys:
            - cache-{{ .Environment.CIRCLE_JOB }}

      - save_cache:
          key: cache-{{ .Environment.CIRCLE_JOB }}
          paths:
            - "/usr/local/lib/R/site-library"
```

inst/templates/circleci-config.yml
```diff
@@ -0,0 +1,25 @@
...
+    docker:
+      - image: {{{ image }}}
+    steps:
+      - restore_cache:
+          keys:
+            - cache
+      - checkout
+      - run:
+          name: Install package dependencies
+          command: R -e "devtools::install_deps(dependencies = TRUE)"
```
jimhester: Probably best to use the remotes packages for this, something like

```r
R -e 'if (!require("remotes")) install.packages("remotes")' -e 'remotes::install_deps(dependencies = TRUE)'
```

We might even just want to unconditionally call `install.packages("remotes")` if for no other reason that to ensure it is installed and is the latest version from CRAN.

inst/templates/circleci-config.yml
```diff
@@ -0,0 +1,25 @@
...
+          command: R CMD build .
+      - run:
+          name: Check package
+          command: R CMD check --as-cran --no-manual *tar.gz
+      - store_artifacts:
+          path: {{{ package }}}.Rcheck/
+      - save_cache:
+          key: cache
+          paths:
+            - "/usr/local/lib/R/site-library"
```
jimhester: At least the rocker containers set the `R_LIBS_USER` environment variable (to this location), maybe this should be $R_LIBS_USER ? Although I am not sure if circleci will expand the environment variable here...

inst/templates/circleci-config.yml
```diff
@@ -0,0 +1,25 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: {{{ image }}}
+    steps:
+      - restore_cache:
+          keys:
+            - cache
```
jimhester: From reading https://circleci.com/docs/2.0/caching/#using-keys-and-templates I think we would want something like

```
 - r-pkg-cache-{{ arch }}-{{ .Branch }}
 - r-pkg-cache-{{ arch }}-
```

This would give you specific caches for the architecture and branch, but would fall back to caches if they are available not available.

inst/templates/circleci-config.yml
jimhester: Also IIRC because this is the default library location for the verse docker container, the CircleCI cache will contain all of the package already installed in the library. It might be better to do something like `R_LIBS_USER=/foo/bar:${R_LIBS_USER}` and setup new library that is only dependencies installed by CircleCI.

inst/templates/circleci-config.yml
jdblischak: To ensure it was always up-to-date, I think we'd need to unconditionally install it. The rocker image verse already has remotes/devtools installed.

inst/templates/circleci-config.yml
jdblischak: This might be possible using [parameters](https://circleci.com/docs/2.0/reusing-config/#using-the-parameters-declaration).

But I do think that most users will use a rocker image, and any user that can create an R environment from a more minimal Docker image (e.g. a standard Ubuntu image) should be able to handle updating the library path.

jdblischak: `R_LIBS_USER` has to be set in `.Renviron`, right? Do you want me to add a step to create this file?

inst/templates/circleci-config.yml
jdblischak: OK. I can update the caching logic.

inst/templates/circleci-config.yml
jimhester: No, it is just a normal shell environment variable, it can be set like you would normally in the CI.

jdblischak: Thanks for the confirmation. That's what I had thought, but I'm having trouble getting the rocker image to recognize `R_LIBS_USER`, even after creating that directory. It works on my local Ubuntu, so I'm not sure what is going wrong. Maybe this discussion is related: https://github.com/rocker-org/rocker-versioned/issues/108#issuecomment-445003537 I'll keep investigating.

jdblischak: Starting a Docker container with `docker run --rm -it rocker/verse:3.5.2 bash`

```
export R_LIBS_USER=~/R/Library
echo $R_LIBS_USER
## /root/R/Library/
mkdir -p $R_LIBS_USER
Rscript -e 'Sys.getenv("R_LIBS_USER")'
## [1] "/usr/local/lib/R/site-library"
```

jdblischak: OK. I confirmed that the problem is that the rocker image is setting `R_LIBS_USER` in a system `Renviron` ([line](https://github.com/rocker-org/rocker-versioned/blob/9ad0922a7cb6932dfacc5ecc387048a1e1cd6e59/r-ver/devel/Dockerfile#L115)). When I delete that line, then it respects the setting of the environment variable.

jdblischak: @jimhester Since setting `R_LIBS_USER` as a standard environment variable didn't work in the rocker image, I tried defining it in `.Renviron`. This was at least recognized by R, but it was added to the *end* of the of the library path.

```
mkdir -p ~/R/Library
echo R_LIBS_USER=~/R/Library > ~/.Renviron
Rscript -e '.libPaths()'
## [1] "/usr/local/lib/R/site-library" "/usr/local/lib/R/library"     
## [3] "/root/R/Library"              
```

Thus I then tried using `.Rprofile`. I've had to resort to this hack before in order to configure R properly on my HPC. Note that only the call to `.libPaths()` is actually required. Including the call to `Sys.setenv()` is more for an end user trying to understand how the library paths were setup.

```
mkdir -p ~/R/Library
echo 'Sys.setenv(R_LIBS_USER = "~/R/Library")' > ~/.Rprofile
echo '.libPaths("~/R/Library")' >> ~/.Rprofile
cat ~/.Rprofile 
## Sys.setenv(R_LIBS_USER = "~/R/Library")
## .libPaths("~/R/Library")
Rscript -e '.libPaths()'
## [1] "/root/R/Library"               "/usr/local/lib/R/site-library"
## [3] "/usr/local/lib/R/library" 
```

Another alternative would be to set `lib = "~/R/Library"` when calling `install.packages()` and `remotes::install_deps()`.

Thoughts on which solution I should implement?

jimhester: The problem is the rocker container is also setting `R_LIBS` and `R_LIBS` takes precedence over `R_LIBS_USER`, so you have to use that in this case.

jdblischak: Setting `R_LIBS` worked. Thanks! By setting `R_LIBS=~/R/library` and creating this directory, `.libPaths()` returns the desired order:

```
export R_LIBS=~/R/library
mkdir -p ~/R/library
Rscript -e '.libPaths()'
## [1] "/root/R/library"               "/usr/local/lib/R/site-library"
## [3] "/usr/local/lib/R/library" 
```

jdblischak:
@jimhester This PR is ready for another round of review. Thanks!

inst/WORDLIST
```diff
@@ -74,15 +76,23 @@ minimise
 msg
 nano
 organisation
+<<<<<<< HEAD
 organisations
 pkgbuild
+=======
```
jimhester: looks like you have some unresolved conflict markers here

jimhester:
LGTM other than the conflict marker issue.

inst/WORDLIST
jdblischak: Strange. I didn't think Git would allow me to add the file and `git rebase --continue` if the file still contained the conflict markers. I'll fix that real quick.

jimhester: No by default, but you can use `git diff --check` to warn about them, or put this in a pre-commit hook if you want to automate it.

R/ci.R
```diff
@@ -147,3 +147,66 @@ check_uses_gitlab_ci <- function(base_path = proj_get()) {
...
+use_circleci <- function(browse = interactive(), image = "rocker/verse:latest") {
+  check_uses_github()
+  use_directory(".circleci")
+  new <- use_template(
+    "circleci-config.yml",
+    ".circleci/config.yml",
+    data = list(package = project_name(), image = image),
+    ignore = TRUE
+  )
+  if (!new) return(invisible(FALSE))
```
jennybc: Can you run styler over this? I think we want this in curly braces, with newlines.

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
+context("use_circleci")
+
+test_that("uses_circleci() reports usage of CircleCI", {
```
jennybc: Can you add `skip_if_no_git_config()`?

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
...
+  scoped_temporary_package()
+  expect_false(uses_circleci())
+  use_git()
+  git2r::remote_add(name = "origin", url = "https://github.com/fake/fake")
```
jennybc: Please switch over to `usethis::use_git_remote()`. We will probably switch over to gert at some point, so I don't want to acquire any new git2r calls.

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
...
+  use_circleci(browse = FALSE)
+  expect_true(uses_circleci())
+})
+
+test_that("check_uses_circleci() can throw error", {
+  scoped_temporary_package()
+  expect_error(check_uses_circleci(),
```
jennybc: I *think* styler would re-format this too.

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
...
+               "Do you need to run `use_circleci\\(\\)`")
+})
+
+test_that("use_circleci() configures CircleCI", {
```
jennybc: Can you add `skip_if_no_git_config()`?

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
...
+  scoped_temporary_package()
+  use_git()
+  git2r::remote_add(name = "origin", url = "https://github.com/fake/fake")
```
jennybc: Please switch over to `usethis::use_git_remote()`.

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
...
+  use_circleci(browse = FALSE)
+  expect_proj_dir(".circleci")
+  expect_proj_file(".circleci/config.yml")
+  yml <- yaml::yaml.load_file(".circleci/config.yml")
+  expect_identical(yml$jobs$build$steps[[7]]$store_artifacts$path,
```
jennybc: styler might change this

tests/testthat/test-use-circleci.R
```diff
@@ -0,0 +1,38 @@
...
+                   paste0(project_name(), ".Rcheck/"))
+})
+
+test_that("use_circleci() specifies Docker image", {
```
jennybc: Same tweaks as above

jdblischak:
@jennybc Thanks for the review! I had clearly overlooked the bullet point about the code style. I think I've addressed everything, so please let me know if I missed anything.

jennybc:
Thanks!

PR #713, opened 2019-04-14
Fix bug in news_latest().

rorynolan:
This is a simple code typo, the `if (length(headings == 1))` statement shouldn't be there. The idea is that if the user's `NEWS.md` file has no top-level headings, then they get the error "No top-level headings found in {ui_value(path)}" but right now what actually happens is that the code falls through to `text <- which(news != "")` and `news` doesn't exist there so the user gets an uninformative error from that.

Sorry for no reprex, kinda hard when the function is searching for a file with a certain name within the current project.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -151,14 +151,12 @@ news_latest <- function() {
   lines <- readLines(path)
   headings <- which(grepl("^#\\s+", lines))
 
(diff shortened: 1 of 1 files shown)


jennybc:
Yeah I agree this is just a mistake and this PR puts it back to the previous state.

PR #715, opened 2019-04-14
pr_pause just switches back to master if no remote branch

### LINKED ISSUE #714: pr_pause throw error if branch not already pushed
```r
usethis::pr_init("my-next-pr-branch")
# commit some stuff
# but not ready to open PR yet so no `usethis::pr_push()` 
# I want to pause the PR to get other things done
usethis::pr_pause()
✔ 
#> Error in git2r::fetch(repo, name = remref_remote(remref), refspec = branch,  : 
#>   Error in 'git2r_remote_fetch': 'name' must be a character vector of length one with non NA value
#> Call `rlang::last_error()` to see a backtrace 
```

I was expecting the go back to master with no error but there an issue with `check_branch_pulled(use = "pr_pull()")` Currently, `pr_pause()` only work if the current branch has been pushed once. (if not `remote` is `NULL` and it is not handled)

I believe to be reasonnable to be able to pause a PR even if there is no remote to the PR and the branch is only local. 

I can make a PR for this if you want.

cderv:
Fixes #714 

Small fix as you can see. I just return early if no remote. 

One think I am not sure: Should there be a message or no message as there is no remote to check ? 

```r
> pr_pause()
✔ No remote branch yet for local branch 'fix-714-pause-no-remote'.
✔ Switching back to 'master' branch
```
or 
```r
> pr_pause()
✔ Switching back to 'master' branch
```


jennybc:
Thanks!

Issue #523, opened 2018-11-30
Handle multiple existing types in use_dependency()


hadley:
This arises when calling `use_rcpp()` twice, because the `LinkingTo` field is handled specially, but is a general problem if the package already appears in multiple fields.

PR #719, opened 2019-04-18
`use_dependency` fix

rorynolan:
The issue is that `use_dependency()` doesn't account for the fact that you could have the package passed to `use_dependency()` already in two places in `DESCRIPTION` (most likely `LinkingTo` and `Imports`) (I guess some people could erroneously have a package in three or more places). This could cause the variable `delta` (which is used in a few `if` statements e.g. `if (delta > 0)`) to have length two and then the `if` generates a warning. I often get this warning because I use `use_rcpp()` every time I want to create another `.cpp` file in `/src`.
I've fixed this with a `for` loop but that's just an idea. I know you're rightly quite particular about how this package is authored and I'm happy to implement a fix to this as instructed, if you so wish.

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -55,37 +55,37 @@ use_dependency <- function(package, type, min_version = NULL) {
     desc::desc_set_dep(package, type, version = version, file = proj_get())
     return(invisible())
   }
-
-  delta <- sign(match(existing_type, types) - match(type, types))
-  if (delta < 0) {
-    # don't downgrade
-    ui_warn(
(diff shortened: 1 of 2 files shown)


jennybc:
Related to #523

hadley:
This code is already quite complex and I'm worried that this for loop makes it more complex. I don't have a good idea of exactly what the solution should look like, but I suspect it will require a bigger rewrite so is best handled by @jennybc or me. 

@rorynolan — we appreciate your work on this and we're sorry it didn't make it in to usethis this time.

PR #727, opened 2019-04-22
Fix #725: git_sitrep() token-checking, issue feedback according to error

### LINKED ISSUE #725: Could git_sitrep() check for network (proxy) connection?
This was the "edgiest" of cases, but I thought it was interesting.

Someone at the workshop used a proxy but did not have their environment variables set up for the proxy. I think that proxy-setup is clearly beyond the scope of usethis, but it did seem that from the `git_sitrep()` perspective, a network failure is indistinguishable from an invalid token.

With my network off:

``` r
usethis::git_sitrep()
#> Git user
#> * Name: 'Ian Lyttle'
#> * Email: 'ian.lyttle@schneider-electric.com'
#> * Vaccinated: TRUE
#> usethis + git2r
#> * Default usethis protocol: <unset>
#> * git2r supports SSH: TRUE
#> * Credentials: '<usethis + git2r default behaviour>'
#> GitHub
#> * Personal access token: '<found in env var>'
#> ✖ Token is invalid.
#> Repo
#> ℹ No active usethis project.
```

<sup>Created on 2019-04-20 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

As a first idea, could `git_sitrep()` look for a successful call to something like `httr::GET("https://httpbin.org/get")` to determine if it has network access?

ijlyttle:
This fixes #725. 

Following @hadley's suggestion, issues `ui_oops()` feedback according to the error returned from the `gh::gh_whoami()` call.

Because we are working with side-effects, I thought to keep everything at the level of the `git_sitrep()` function.

Happy to make any changes you might suggest.


R/git.R
```diff
@@ -479,11 +479,23 @@ git_sitrep <- function() {
...
-    who <- github_user()
-    if (is.null(who)) {
-      ui_oops("Token is invalid.")
-      have_token <- FALSE
-    } else {
+
+    who <-
+      suppressMessages(
+        tryCatch(
+          gh::gh_whoami(),
```
hadley: I think you should pass the github token explicitly here, which I believe will allow you to eliminate the `suppressMessages()`.

hadley: And I'd suggest making this block

```R
who <- gh::gh_whoami()
kv_line("User", who$login)
kv_line("Name", who$name)
```

Then you can simplify the error handlers, and remove the if block below.

ijlyttle: Thanks for the suggestions - it looks much better now!

ijlyttle:
Should be ready to go, thanks again!

jennybc:
Thanks!

PR #728, opened 2019-04-24
Fix broken link

jmgirard:
https://adv-r.hadley.nz/introduction-16.html
should be: 
https://adv-r.hadley.nz/metaprogramming.html


jennybc:
Thanks!

PR #730, opened 2019-04-25
Link to topic file rather than alias (tidy eval)

jmgirard:
This change will prevent Rd warnings during "Install and Restart" on Windows. However, it may make the links more fragile.
https://github.com/klutometis/roxygen/issues/707
https://stackoverflow.com/questions/48430093/how-do-i-resolve-rd-warning-missing-file-link-when-building-packages-in-rstudi/48478698#48478698


jennybc:
Thanks!

Issue #732, opened 2019-04-26
Change order of the instructions to users in edit_r_profile and edit_r_environ


llrs:
When I want to edit my .Rprofile I see this:
```r
edit_r_profile()
● Restart R for changes to take effect
● Modify '/home/lrevilla/.Rprofile'
```
However I think that the order should be the opposite, first Modify then Restart. Similarly the order can be changed in `edit_r_environ`. 

```r
edit_r_profile()
● Modify '/home/llrs/.Rprofile'
● Restart R for changes to take effect
```

I'm happy to submit a small PR if you agree on this

https://github.com/r-lib/usethis/blob/e8fb9d75e1ad0193ce9687a774296f703a9e37a0/R/edit.R#L63

jennybc:
I suspect there's a bit more to think about here, because certain things are arranged they way they are to accommodate context where we can and cannot open the file for editing. But it's certainly OK for you to poke around and see if there's an obvious rationalization.

llrs:
The only reference I found to `edit_r_profile` in `usethis` is in the `use_rprofile_package` function, which affects the `use_reprex`, and other functions of the family rprofile-helper. Which currently suggest to:
```
use_reprex()
● Restart R for changes to take effect
● Modify '/home/llrs/.Rprofile'
● Include this code in '.Rprofile' to make reprex available in all interactive sessions.
  if (interactive()) {
    suppressMessages(require(reprex))
  }
```

Which ideally I think would be easier to read 

```
● Include this code in '.Rprofile' to make reprex available in all interactive sessions.
  if (interactive()) {
    suppressMessages(require(reprex))
  }
● Modify '/home/llrs/.Rprofile'
● Restart R for changes to take effect
```

I couldn't find (but maybe I need to think longer and not on Friday evening) a situation where it cannot be possible to open the file for editing. I'll submit a PR by this weekend.

jennybc:
> I couldn't find (but maybe I need to think longer and not on Friday evening) a situation where it cannot be possible to open the file for editing. I'll submit a PR by this weekend.

We can't assume people are using RStudio or an IDE of any sort.

cderv:
FWIW this change happened in e04499b770010689f8968bf2bb983fe1adb71369 where `edit_r_profile` and `edit_r_environ` were made compatible with environment variable. 
This is where the order has changed. 

[...]

PR #734, opened 2019-04-27
Reorder the messages in edit_r_profile, edit_renviron and rprofile-helper functions

llrs:
This PR changes the order of the messages that the user sees when editing the Rprofile and the Renviron file as proposed on #732.

### DIFF
--- a/R/edit.R
+++ b/R/edit.R
@@ -60,17 +60,17 @@ NULL
 #' @export
 #' @rdname edit
 edit_r_profile <- function(scope = c("user", "project")) {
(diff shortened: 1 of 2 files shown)


hadley:
Can you please change the title of this issue to be informative?

llrs:
Done

jennybc:
I no longer remember why I was so anxious about this. I'll merge and try to play with a bit to make sure I'm not missing something obvious.

llrs:
Great! (I also didn't understand what were the problems with this reordering, but let's see if there is any trouble down the line)

PR #739, opened 2019-04-29
Update gitlab-ci.yml

### LINKED ISSUE #738: Gitlab-ci issues with package_coverage
This won't work at all as-as, you are not actually calling `package_coverage`, just assigning the function to a different name.

I think this could be simplified to just

```r
covr::gitlab(quiet = FALSE)
```

With the mentioned regex above if you want the automatic coverage badge.

_Originally posted by @jimhester in https://github.com/r-lib/usethis/pull/562_

overmar:
Updates to covr::gitlab() per @jimhester fixes #738

More background: https://github.com/r-lib/usethis/pull/562#discussion_r279391626

### DIFF
--- a/inst/templates/gitlab-ci.yml
+++ b/inst/templates/gitlab-ci.yml
@@ -8,10 +8,11 @@ stages:
 building:
   stage: build
   script:
-    - R -e "remotes::install_deps(dependencies = T)"
+    - R -e "remotes::install_deps(dependencies = TRUE)"
     - R -e 'devtools::check()'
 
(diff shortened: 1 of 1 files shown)


jimhester:
LGTM, @overmar could you verify this does what you want in an example GitLab project?

overmar:
https://gitlab.com/remotes.test.testthat/test, the issue with the pages site is you have to add coverage.html, or you need to use pkgdown to create the whole site that is then passed out to public. Since coverage.html isn't index.html it doesn't initially know how to deal with it.

PR #743, opened 2019-05-01
Avoid resetting project after with_project() and local_project()

krlmlr:
by returning old project path on all execution paths of `proj_set()`.

## Revised behavior in this PR

``` r
here::set_here()
#> Created file .here in /tmp/RtmpS1nTRM/reprex3084505b53ad

usethis::proj_get()
#> ✔ Setting active project to '/tmp/RtmpS1nTRM/reprex3084505b53ad'
#> /tmp/RtmpS1nTRM/reprex3084505b53ad

usethis::with_project(code = print("Hi!"))
#> [1] "Hi!"
usethis::with_project(code = print("Hi!"))
#> [1] "Hi!"
usethis::with_project(code = print("Hi!"))
#> [1] "Hi!"
```

<sup>Created on 2019-05-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>

## Original, problematic behavior in 5482403

``` r
here::set_here()
#> Created file .here in /tmp/Rtmp2bJVqd/reprex2d9e2cea0669

usethis::proj_get()
#> ✔ Setting active project to '/tmp/Rtmp2bJVqd/reprex2d9e2cea0669'
#> /tmp/Rtmp2bJVqd/reprex2d9e2cea0669

usethis::with_project(code = print("Hi!"))
#> [1] "Hi!"
usethis::with_project(code = print("Hi!"))
#> ✔ Setting active project to '/tmp/Rtmp2bJVqd/reprex2d9e2cea0669'
#> [1] "Hi!"
usethis::with_project(code = print("Hi!"))
#> ✔ Setting active project to '/tmp/Rtmp2bJVqd/reprex2d9e2cea0669'
#> [1] "Hi!"
```

<sup>Created on 2019-05-02 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1.9000)</sup>


jennybc:
Thanks!

This seems correct and it's a small change. And yet I feel like @Hadley and I sweated over these bits. So I'm just going to "be" with it for a little while, in case any subtle reason reveals itself as to why we didn't do this way from the start.

tests/testthat/test-proj.R
```diff
@@ -128,20 +128,60 @@ test_that("proj_sitrep() reports current working/project state", {
...
+  proj_set(host)
+  expect_identical(proj_get_(), host)
+
+  res <- with_project(path = guest, proj_get_())
+
+  expect_identical(res, guest)
+  expect_identical(proj_get(), host)
+})
+
+test_that("with_project() works when temp proj == original proj", {
```
jennybc: I wondered why my existing test didn't catch the behaviour you pointed out @krlmlr.

So I added this one, which does, and another as well.

R/proj.R
```diff
@@ -70,7 +70,7 @@ proj_get <- function() {
 #' @export
 proj_set <- function(path = ".", force = FALSE) {
   if (dir_exists(path %||% "") && is_in_proj(path)) {
-    return(invisible())
+    return(invisible(proj_get_()))
```
jennybc: This also makes `proj_set()` type stable, because the only other exit point returns a path, invisibly.

krlmlr:
Thanks for looking into it. Type (and return value) stability is the most important thing here I think. I'm not perfectly happy with the control flow here, maybe the logic should be in a subfunction whose return value is ignored. But then we have tests (thanks!).

The original use case was a function that has a `path` argument and was oblivious to the state of `usethis` and shouldn't alter it too. This function now fully interacts with `proj_get()` and simply uses what it finds.

jennybc:
> I'm not perfectly happy with the control flow here

Yeah, I know. This was my best effort last time I looked at it. But the more usage and tests we accumulate the better. Maybe we'll re-express more elegantly one day.

PR #744, opened 2019-05-02
Update binder badge URL

nuest:
Binder has a new badge now with a logo, see https://discourse.jupyter.org/t/help-us-choose-an-updated-launch-binder-badge/100/21

Also changed the protocol to https, to which http is redirected.


jennybc:
Thanks!

PR #753, opened 2019-05-08
Update glue version to >= 1.3.0

### LINKED ISSUE #752: Bump glue dependency to 1.3.0
I could not install usethis with glue 1.2.0

```r
> install.packages("usethis")
* installing *source* package ‘usethis’ ...
** package ‘usethis’ successfully unpacked and MD5 sums checked
** R
** inst
** byte-compile and prepare package for lazy loading
Error : object ‘glue_collapse’ is not exported by 'namespace:glue'
ERROR: lazy loading failed for package ‘usethis’
```
This is because `glue_collapse` is new in glue 1.3.0. Could you please update the version in the DESCRIPTION file to reflect that?

batpigandme:
* Fixes #752

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Imports:
     fs (>= 1.2.0),
     gh,
     git2r (>= 0.23),
-    glue (>= 1.2.0),
+    glue (>= 1.3.0),
     purrr,
     rlang,
     rprojroot (>= 1.2),


jennybc:
@jimhester @hadley 

Quick check in. I suspect usethis may be the thing that pulls glue in for the devtools-verse.

The two choices and some facts:

  1. Bump the minimum glue version to v1.3.0 = this PR.
    - Current min of v1.2.0 was released 2017-10-29.
    - Proposed min of v1.3.0 was released 2018-07-17.
    - BTW I don't see anything in v1.3.1 (released 2019-03-12) that suggests we need to require it. That is, v1.3.0 is enough to sort out `glue_collapse()`.
  1. Keep the minimum glue version at v1.2.0 and use the solution @jimhester contributed to [googledrive](https://github.com/tidyverse/googledrive/blob/c0a62bd7720c5bb85a349e5f8107b3f2a01ee5d0/R/utils.R#L117-L123).

jimhester:
I think we can just bump the version of glue.

PR #757, opened 2019-05-09
Add use_tidy_versions() to release issue

krlmlr:
Do we generally recommend this?


jennybc:
No, as per a semi-recent Slack discussion, we do not generally recommend this and, in fact, I had already opened an issue about renaming / reworking `use_tidy_versions()` (#771). I am eager to see what happens with your [minver](https://github.com/krlmlr/minver) efforts!

PR #758, opened 2019-05-10
ui_path(): Avoid putting a slash on a directory that has already been specified with a slash on the end.

rorynolan:
Currently, for a directory, if specified with a slash on the end, `ui_path()` appends another one. I think it shouldn't.
For example, `ui_path("R/")` outputs `R//` but I think it shouldn't. With my new implementation, `ui_path("R/")` outputs `R/`. 
My implementation is very minimal. It still has `ui_path("R//")` outputting `R//`. To fix that to output `R/`, regex would be needed. Up to you whether or not to do that.


R/ui.R
```diff
@@ -186,7 +186,9 @@ ui_path <- function(x, base = NULL) {
     x <- path_rel(x, base)
   }
 
-  x <- paste0(x, ifelse(is_directory, "/", ""))
+  if (!endsWith(x, "/")) {
```
jennybc: FYI `endsWith()` only appeared in R 3.3, so we can't use it w/o also using backports, which usethis does not. Right now, we still promise to support 3.2.

Anyway, I did a bit of refactoring and this is one of the things I changed.

jennybc:
Thanks!

PR #768, opened 2019-05-20
`use_version()` directly replaces development version NEWS headers

DavisVaughan:
This PR allows you to call `use_version("major")` when you are currently in a state where the first line of NEWS is `# pkg (development version)` and directly replace that header with the bumped version. This means you no longer have to manually delete the temporary dev header.


jennybc:
Thanks!

PR #778, opened 2019-06-05
Replace devtools with remotes for install_ templates

Robinlovelace:


Robinlovelace:
Heads-up @jennybc I've just been using `usethis` and noticed it's still using the old `devtools::install_github()` rather than the new `remotes::install_github()`. Think this fixes it. Any comments welcome. Many thanks!

jennybc:
The optics of this (emphasizing devtools over remotes) is intentional. We think of devtools as the user-facing storefront for interactive use, so in this context, devtools still feels like the right way for most people to get at this function.

Robinlovelace:
Aha, that makes sense. In line with the 'conscious decoupling' of `devtools`. I think it could be confusing to some people that there are seemingly 2 versions of `install_github()`, but the +s of having everything in the `devtools` bag outweigh that -.

jennybc:
We are very actively trying to 1) reach internal consensus and 2) be consistent in our messaging about this.

Here's my current best take on it:
https://r-pkgs.org/intro-why.html#intro-usage

Feedback on the r-pkgs repo via an issue is always welcome.

Robinlovelace:
Thanks for the additional explanation. Building on the updated text, I guess the aim is for as many useRs to become developRs as possible. I think there could be an argument for encouraging people to start moving in that direction. See the point that having to remember function names scattered across 7 different packages is not ideal! In interactive mode `devtools::load_all()` is much easier to remember than `pkgload::load_all()`, so the explanation makes sense. Not sure about the useR/developR category labels. I imagine saying that gradually moving from being a useR to a developeR is one of the aims of the book although in reality many people are useRs and developeR, just at different times and on different projects. How about something like 'when in useR mode', so it's categorising what the person is doing not the person? Could put that in a PR if interested. 

In any case great to see the 'r-pkgs' book being updated, it was a valuable resource when I was developing `stplanr` (which is undergoing its own conscious uncoupling process, resulting in new CRAN packages `stats19` and `cyclestreets`) with limited programming experience.

jennybc:
> Not sure about the useR/developR category labels

Yeah I am using them more to describe one's state of mind at any given moment.

If you want to try out the dev version of someone else's package, just call `install_github()`, assuming you've attached devtools one way or another. When you're writing functions inside an R package, take the trouble to identify the package where that function is actually maintained and import it from the true source.

Robinlovelace:
>  I am using them more to describe one's state of mind at any given moment.

That doesn't come through in the current prose for me. I read: "For interactive use, useRs should attach devtools and think of it as the provider of your favorite functions for package development." Could be useful to define that a useR is more a state of mind than a type of person. or could be something like "For interactive use, in 'useR mode', devtools can be considered the provider of functions for package development."

Just a suggestion. Hope it's of use/interest, best of luck with it and look forward to reading the finalised second edition!

PR #780, opened 2019-06-10
Minor edits to docs

mine-cetinkaya-rundel:
Three minor edits to docs that I came across while working my way through them


jennybc:
Thanks! I took some, left some, tweaked some.

PR #809, opened 2019-07-08
allow 'tidyverse' in use_package if type='depends' (For data analysis…

### LINKED ISSUE #760: Tone down refusal of using tidyverse as a dependency
I have put together a minimal set of "best practices" that help when you start using functions in data analysis projects: https://krlmlr.github.io/tidyprog/best-practices.html. This includes declaring the packages your analysis uses: https://github.com/krlmlr/tidyprog/blob/master/08-Best-practices.Rmd#L37.

`use_package("tidyverse")` is "verboten". This is for a good reason, but I think that a data analysis project that uses parts of R's packaging infrastructure should be allowed to at least depend on the tidyverse. Would you support an amendment of this check?

VincentGuyader:
… projects). fix #760

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -76,15 +76,14 @@ package_remote <- function(package) {
 
 refuse_package <- function(package, verboten) {
   if (identical(package, verboten)) {
+    code <- glue("use_package(\"{package}\", type = \"depends\")")
     ui_stop(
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks!

PR #810, opened 2019-07-08
Revise `ui_yeah` / `ui_nope`

### LINKED ISSUE #796: `ui_yeah` and `ui_nope` option choices and ordering
I like both of these functions and would like to use them in `ghclass` when users request potentially destructive actions. However, I'd like the option to control what choices are presented to the user and whether or not they are randomized.

Both of these behaviors were also unclear to me from the docs. 

Happy to propose a modified version of both if that would be helpful.

rundel:
Closes #796

Here is one possible starting point for enabling more control over `ui_yeah` and `ui_nope`. The goal here is to let the package author specific the exact options that will be shown for both the yes and no options and control whether they are randomly shuffled.


R/ui.R
```diff
@@ -124,9 +124,43 @@ ui_warn <- function(x, .envir = parent.frame()) {
...
+#' ui_yeah("Do you like R?")
+#' ui_nope("Have you tried turning it off and on again?", n_yes = 1, n_no = 1)
+#' ui_yeah("Are you sure its plugged in?", yes = "Yes", no = "No", shuffle = FALSE)
+#' }
+NULL
+
+#' @rdname ui_question
 #' @export
-ui_yeah <- function(x, .envir = parent.frame()) {
+ui_yeah <- function(
```
jennybc: Before this is done, you should run this through styler, because we format this differently.

R/ui.R
```diff
@@ -137,21 +171,32 @@ ui_yeah <- function(x, .envir = parent.frame()) {
...
-  nays <- c("No way", "Not now", "Negative", "No", "Nope", "Absolutely not")
+  n_yes <- min(n_yes, length(yes))
+  n_no <- min(n_no, length(no))
 
-  qs <- c(sample(ayes, 1), sample(nays, 2))
-  ord <- sample(length(qs))
+  qs <- c(sample(yes, n_yes), sample(no, n_no))
+
+  if (shuffle)
+    qs <- sample(qs)
```
jennybc: styler will put brackets around this.

R/ui.R
```diff
@@ -137,21 +171,32 @@ ui_yeah <- function(x, .envir = parent.frame()) {
...
 #' @export
-ui_nope <- function(x, .envir = parent.frame()) {
-  !ui_yeah(x, .envir = .envir)
+ui_nope <- function(
+  x,
+  yes =  c("Yes", "Definitely", "For sure", "Yup", "Yeah", "I agree", "Absolutely"),
+  no  =  c("No way", "Not now", "Negative", "No", "Nope", "Absolutely not"),
+  n_yes = 1, n_no = 2, shuffle = TRUE,
+  .envir = parent.frame()
+) {
```
jennybc: See styling comment from above.

R/ui.R
```diff
@@ -124,9 +124,43 @@ ui_warn <- function(x, .envir = parent.frame()) {
...
+#'   when either a no option or no selection is made (canceled). `ui_nope` is the logical
+#'   opposite of `ui_yeah`.
+#' @name ui_question
+#' @examples
+#' \dontrun{
+#' ui_yeah("Do you like R?")
+#' ui_nope("Have you tried turning it off and on again?", n_yes = 1, n_no = 1)
+#' ui_yeah("Are you sure its plugged in?", yes = "Yes", no = "No", shuffle = FALSE)
+#' }
+NULL
```
jennybc: I think this and the help from the rest of the `ui_*()` functions should link to each other, if we split them up.

R/ui.R
rundel: Sounds good, just ran things through and updated

R/ui.R
rundel: Makes sense, revisions include explicit links in the text as well as the inclusion of a user interface family

rundel:
I'm happy to contribute tests, but I don't have any experience in working around the interactive issue. Any suggestions on spoofing `interactive()` and `menu()`?

jennybc:
Thanks! Don't worry about the tests. I can pursue that myself.

PR #812, opened 2019-07-08
Add LICENSE file to the package and add "+ file LICENSE" to the description file for gpl-3 license

### LINKED ISSUE #683: use_gpl3_license does not create GPL3+ file LICENSE like MIT
When we use, `use_gpl3_license()` the description file does not contain CRAN compliant `License: GPL-3 + file LICENSE` in description file. 

Without this feature, when we run `R CMD check --as-cran mypackage` we get the following NOTE:

```
* checking top-level files ... NOTE
File
  LICENSE
```
Having `use_gpl3_license()` output `GPL-3+ file LICENSE` in description file would be very helpful similar to the output of `use_mit_license()`

Cervangirard:
Add LICENSE file to the package and add "+ file LICENSE" to the description file


R/license.R
```diff
@@ -55,8 +55,14 @@ use_gpl3_license <- function(name = find_name()) {
   force(name)
   check_is_package("use_gpl3_license()")
 
-  use_description_field("License", "GPL-3", overwrite = TRUE)
+  use_description_field("License", "GPL-3 + file LICENSE", overwrite = TRUE)
   use_license_template("GPL-3", name)
+
+  use_template(
+    "license-mit.txt",
```
jennybc: This is confusing (although it produces the correct result). I think we should rename this file, so the code is less puzzling.

I propose `year-copyright.txt`. After you rename it, you'll have to update the MIT license code to use the new name too.

Cervangirard: It should be good :)

jennybc:
Can you also change the test? I think the test for `use_gpl3_license()` should like the one for `use_mit_license()` in terms of the match in the `License` field of DESCRIPTION.

Then I think this is done!

Cervangirard:
Shoud be ok for the test now.

jennybc:
Thanks!

PR #813, opened 2019-07-08
Document use_r and use_test in one documentation file

friep:
This changes the documentation for `use_test` and `use_r`, putting them in the same documentation file.

For that, I create a dummy documentation file `creating-files.R` that keeps the description and the common parameter `name` . I then add to that using `@rdname` in the documentation of the two functions. 

As a consequence, this PR removes `use_test` from the testing documentation that now only contains `use_testthat`, so its documentation was changed to reflect that.


hadley:
Thanks @friep!

PR #815, opened 2019-07-08
Remove unexported function from use_rmarkdown_template signature

### LINKED ISSUE #761: error when running code as is from use_rmarkdown_template
If you run the code as given in the default function signature for `use_markdown_template()`, you encounter an internal function, `asciify()` and therefore an error, which can be confusing to new users

``` r
usethis::use_rmarkdown_template(template_name = "Basic Template",
                        template_dir = tolower(asciify(template_name)),
                        template_description = "the most basic of templates",
                        template_create_dir = FALSE)
#> Error in asciify(template_name): could not find function "asciify"
```
<sup>Created on 2019-05-13 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

fmichonneau:
Fixes #761

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 
 * Fix quoting of dataset name in template for `use_data_raw()` (#736, @mitchelloharawild).
 
-* Remove unexported function from `use_markdown_template()` (#761, @fmichonneau)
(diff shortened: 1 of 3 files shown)


jennybc:
Thanks!

Issue #756, opened 2019-05-09
update issue template to current github format


maxheld83:
split off from #755 as suggested by @jennybc.

<img width="1237" alt="old-template" src="https://user-images.githubusercontent.com/5372770/57480004-f27a2080-729e-11e9-8f5d-eac841c03d7e.png">

maxheld83:
maybe *this* might be a tdd candidate?

jennybc:
The intent of this issue was to change the issue template used by `use_tidy_issue_template()`, which #818 PR doesn't do. I.e. we have to change the file `inst/templates/tidy-issue.md`.

I'm going to reopen.

@Maschette Do you want to take another crack at it? I'm going to point out a few things in the original PR.

Issue #755, opened 2019-05-09
support multiple issue issue templates


maxheld83:
It might be nice to add *several* issue templates, from which people can than choose the type of issue (say, bug report or feature request). 

Like so:

![new-issue-page-with-multiple-templates](https://user-images.githubusercontent.com/5372770/57464040-48899c80-727c-11e9-95c3-d9aa2d6bed66.png)

This could help people to self-filter their issues a bit, maybe even reducing the issue load.

FYI, current issue template is also now superseded by a new format, says github when inspecting the source for `usethis::use_tidy_issue_template()`:

<img width="1237" alt="old-template" src="https://user-images.githubusercontent.com/5372770/57464188-956d7300-727c-11e9-8ce1-89f5dc5f1b64.png">

cc @batpigandme 

Also, maybe this would be a nice tdd-issue? (can't tag it).

jennybc:
I think updating the current, sole issue template should be in a separate issue from the possibility of having multiple templates.

hadley:
Yeah, and the multiple template issue isn't a great fit for TDD since we need to decide on which templates first.

hadley:
I don't find these templates terribly useful, and we don't have a concrete proposal here, so I think it's time to close this issue.

PR #818, opened 2019-07-08
tidy issue template update for github changes

Maschette:
This addresses the issue #756. The main changes  were adding a "title" and "about" header to the issue template, and moving the template into a folder called "ISSUE_TEMPLATE", and update the function to account for the directory change.. 

The latter is the important one as if it is not in this folder than for some reason github thinks you are still using an old version. It also gives the location to consider multiple reporting templates as highlighted by @maxheld83 in issue #755


R/tidyverse.R
```diff
@@ -198,7 +198,7 @@ use_tidy_issue_template <- function() {
   use_directory(".github", ignore = TRUE)
   use_template(
     "tidy-issue.md",
-    ".github/ISSUE_TEMPLATE.md"
+    ".github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.md"
```
lionel-: Maybe the file should be in lower case, like in the [github documentation](https://help.github.com/en/articles/manually-creating-a-single-issue-template-for-your-repository)? Seems a bit less shouty :)

Maschette: I just kept it the same as what is was but I am happy to change it if you want.

lionel-: What do you think @batpigandme?

batpigandme: For sure! Their old rec was to "yell" `ISSUE_TEMPLATE`, so I'm happy to see it lowercase-ified!

lionel-:
Thanks!

Maschette:
Thanks @lionel- and @batpigandme!

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.md
```diff
@@ -1,9 +1,12 @@
+---
```
jennybc: This sort of change needs to happen in `inst/templates/tidy-issue.md`. Then we could apply the revised function to usethis itself, to change its issue template(s).

R/tidyverse.R
```diff
@@ -198,7 +198,7 @@ use_tidy_issue_template <- function() {
   use_directory(".github", ignore = TRUE)
```
jennybc: What about the `.github/ISSUE_TEMPLATE` subdirectory?

jennybc:
OK I'm just catching up now and I see that @lionel- made subsequent commits addressing some of these points, i.e. the subdirectory and the tests.

So that just leaves the matter of editing the template used by `use_tidy_issue_template()` vs. the issue template in the usethis repo.

Maschette:
I sent a separate pull request which copied the issue template header from the usethis repo into the issue template used by use_tidy_issue_template(). What I am not sure on as I am not familiar with use_directory is making sure that when you use `use_tidy_issue_template()` for a package it will put the template in .github/ISSUE_TEMPLATE/ which is where it needs to be for github to not complain.

PR #819, opened 2019-07-08
Add example to use_build_ignore()

### LINKED ISSUE #722: use_build_ignore generates invalid regular expression
When using a wildcard in the files argument the use_build_ignore function generates invalid regex.

For example `use_build_ignore("*.xlsx")` will print `Adding '^*\\.xlsx$' to '.Rbuildignore'` in the console but in fact adds `^*\.xlsx$` (note the difference) which will break devtools::check(). Either way, I believe both are invalid. I believe the correct regex would be `^.*\.xlsx$`. 

What I was trying to do was to ignore all Excel files. There might be some reason not use a wildcard, but in that case the function should just print an error or warning. 

```r
usethis::use_build_ignore("*.xlsx")
```

ijlyttle:
Fixes #722 

Ready for review.

Adds an example to `use_build_ignore()` to show how to ignore all files that have a given extension.

I tested this manually, the example works for me. It is unclear how to test this, but I am happy to attempt it if you would like.

As well, I did not think this merits a NEWS item, but will be happy to add an item if you like.

### DIFF
--- a/R/ignore.R
+++ b/R/ignore.R
@@ -6,10 +6,18 @@
 #' expression that will only match that path. Repeated entries will be
 #' silently removed.
 #'
+#' `use_build_ignore` is designed to ignore *individual* files. If you
+#' want to ignore *all* files with a given extension, consider providing
+#' an "as-is" regular expression, using `escape = FALSE`; see examples.
+#'
 #' @param files Character vector of path names.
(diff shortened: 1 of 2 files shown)


hadley:
Thanks @ijlyttle — it doesn't need a news bullet since it's not a user facing change.

PR #821, opened 2019-07-08
helper for preparing URLs for use_course()

fmichonneau:
Closes #406

Issue #406, opened 2018-07-03
Helper for constructing links suitable for use_course()


jennybc:
Provide more help (and a convenient place of docs) re: how to get "download" links for GitHub, DropBox, Google Drive.

If possible, also a way to make short links from usethis.

jennybc:
Relevant to RStudio folks, note that Rebrandly has an API:
https://developers.rebrandly.com/docs

jennybc:
Example: consider this GitHub repo

https://github.com/jennybc/explore-libraries

The correct `use_course()` call is

use_course("https://github.com/jennybc/explore-libraries/archive/master.zip")

It would be nicer to say `use_course("jennybc/explore-libraries"). This reduces the need to create short links.

cwickham:
I'm going to work on this for tidy-dev-day

cwickham:
OK, to not bite off too much at once, I'll focus on the github repo use case, `use_course("jennybc/explore-libraries")`.  Essentially turning the recommendations about URLs for Github (and Dropbox?) in `?use_course_details` into things that happen automatically.

API questions for @jennybc , how should this work from a user's perspective:

* Somehow automatically identify that a user is specifying a `user/repo` form of link: `use_course("jennybc/explore-libraries")`?  But how to distinguish this from a shortlink minus the `http`, e.g. `use_course("bit.ly/explore-libraries")`?

* Make the user identify this is a github link via an argument (which might have a useful default?), `use_course("jennybc/explore-libraries", github = TRUE)`, or `use_course("jennybc/explore-libraries", host = "github")`?  But this defeats the purpose of less typing.

* A different function name, `use_github_course("jennybc/explore-libraries")`?

Either way, we need to keep mind how to generalise to other common sources, i.e. Dropbox.

cwickham:
Answers based on in-person discussion:

* `use_course()` should identify a github style shortlink and expand to the required URL: `use_course("https://github.com/jennybc/explore-libraries/archive/master.zip")`

* Distinguish between github and shortlinks via a shortlink whitelist `bit.ly`, `rtsd.io`

* Look into repo spec to help.

PR #824, opened 2019-07-08
Document how to use use_course() with a folder on Google Drive

perezp44:
Document how to use use_course() with a folder on Google Drive in course.R file.


hadley:
Part of #406

jennybc:
Thanks!

But I'm afraid we were seeking a way to use `use_course()` with a **folder** on Google Drive, the same way that it works with DropBox or. GitHub repo. Your example uses a **ZIP file** stored on Google Drive, which is an existing use case.

In #821, @fmichonneau and I are, sadly, concluding that what we want isn't really possible 😔

PR #826, opened 2019-07-08
Check GitHub email with sitrep

### LINKED ISSUE #724: Could git_sitrep() check git vs. GitHub email?
This is one of the edge cases that came up during the workshop - someone's git email (`user.email`) was not registered with their GitHub account.

I wonder if something like this could be checked in `git_sitrep()`. I had a quick look, I don't *think* you're doing it now. Further, I would think that the GITHUB_PAT would need read access to `user:email`, which does not appear to be currently among the scopes in the default for `browse_github_pat()`.

Just a thought.

dragosmg:
closes #724


dragosmg:
Is there anything else I need to do - should I add a bullet point in news?

dragosmg:
does the example on the package website need to be modified? 
https://usethis.r-lib.org/reference/git_sitrep.html

jennybc:
Nice! The example will update itself when the website is re-built and deployed, although it won't be very satisfying anyway, since the dummy travis user has no PAT.

Issue #822, opened 2019-07-08
`use_tutorial()` - add option to import learnr


isteves:
Currently, `use_tutorial()` places `learnr` in "Suggests", which is ideal for package developers who want to include `learnr` tutorials as documentation. 

In a teaching context, it is often more important to reduce user friction than to reduce dependencies. Thus, we'd like to create a "heavyweight"/learner-friendly option for `use_tutorial()` that would do the following:

1. place `learnr` in "Imports" instead
2. re-export `learnr::run_tutorial()`

@jules32 @angela-li

hadley:
This just feels a bit too specific for usethis — we'd need to distinguish between the package-developer and teacher uses cases either with different function names or arguments, and I just doubt that it would be discoverable enough to actually help people. I think it's a good idea, it just doesn't feel like usethis is the right home for it to me.

PR #827, opened 2019-07-08
Added learner-friendly option to use_tutorial

irenetlv:
Added a new argument, `learner_friendly`, to `use_tutorial()` with @angela-li and @jules32. Defaults to FALSE. If TRUE, then:

- re-exports `learnr::run_tutorial()`
- places `learnr` in Imports instead of Suggests

@mine-cetinkaya-rundel any further comments/suggestions?
https://github.com/r-lib/usethis/issues/822

### DIFF
--- a/R/tutorial.R
+++ b/R/tutorial.R
@@ -14,6 +14,7 @@
 #' @param name Base for file name to use for new `.Rmd` tutorial. Should consist
 #'   only of numbers, letters, `_` and `-`. We recommend using lower case.
 #' @param title The human-facing title of the tutorial.
+#' @param learner_friendly If `TRUE`, adds `learnr` to Imports instead of Suggests to install `learnr` alongside the tutorial package. Re-exports `run_tutorial()` to make it immediately available to tutorial package users.
 #' @inheritParams use_template
 #' @seealso The [learnr package
 #'   documentation](https://rstudio.github.io/learnr/index.html).
@@ -22,7 +23,8 @@
 #' \dontrun{
 #' use_tutorial("learn-to-do-stuff", "Learn to do stuff")
 #' }
-use_tutorial <- function(name, title, open = interactive()) {
+use_tutorial <- function(name, title, learner_friendly = FALSE,
+                         open = interactive()) {
   stopifnot(is_string(name))
   stopifnot(is_string(title))
 
@@ -31,8 +33,14 @@ use_tutorial <- function(name, title, open = interactive()) {
 
   use_directory(dir_path)
   use_git_ignore("*.html", directory = dir_path)
-  use_dependency("learnr", "Suggests")
-
+  
+  if(learner_friendly) {
+    use_dependency("learnr", "Imports")
+    use_template("run-tutorial.R", "R/utils-run-tutorial.R")
+  } else {
+    use_dependency("learnr", "Suggests")
+  }
+  
   path <- path(dir_path, asciify(name), ext = "Rmd")
 
   data <- project_data()
@@ -45,6 +53,6 @@ use_tutorial <- function(name, title, open = interactive()) {
     ignore = TRUE,
     open = open
   )
-
+  
   invisible(new)
 }
--- a/inst/templates/run-tutorial.R
+++ b/inst/templates/run-tutorial.R
@@ -1,0 +1,5 @@
+#' Run a tutorial
+#' @keywords NULL
+#' @export
+#' @name run_tutorial
+learnr::run_tutorial
--- a/tests/testthat/test-use-tutorial.R
+++ b/tests/testthat/test-use-tutorial.R
@@ -2,7 +2,7 @@ context("test-use-tutorial")
 
 test_that("use_tutorial() checks its inputs", {
   skip_if_not_installed("rmarkdown")
-
+  
   scoped_temporary_package()
   expect_error(use_tutorial(), "no default")
   expect_error(use_tutorial(name = "tutorial-file"), "no default")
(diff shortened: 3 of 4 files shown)


mine-cetinkaya-rundel:
This is along the lines of what I was thinking!

tests/testthat/test-use-tutorial.R
```diff
@@ -20,3 +20,14 @@ test_that("use_tutorial() creates a tutorial", {
     }
   )
 })
+
+test_that("use_tutorial(learner_friendly = TRUE) adds promised file, Imports learnr", {
+  with_mock(
+    `usethis:::uses_roxygen` = function(base_path) TRUE, {
```
jennybc: This also needs to use a mock for  `usethis:::is_installed`, as the test above it does. See the test failures on travis and appveyor.

inst/templates/run-tutorial.R
```diff
@@ -0,0 +1,5 @@
+#' Run a tutorial
+#' @keywords NULL
+#' @export
+#' @name run_tutorial
+learnr::run_tutorial
```
jennybc: The form of this seems a bit surprising to me. Have you created a test package and used `use_tutorial(name = "aaa", title = "bbb", learner_friendly = TRUE)`, run `document()`, installed it and checked that it does what you intend?

If you look at the template added by `use_tidy_eval()` and the way devtools re-exports functions from other packages, there are some differences. Based on those, I would expect to see an `@importFrom` directive that imports `run_tutorial` from learnr.

What is the intent with `@keywords NULL`?

R/tutorial.R
```diff
@@ -14,6 +14,7 @@
 #' @param name Base for file name to use for new `.Rmd` tutorial. Should consist
 #'   only of numbers, letters, `_` and `-`. We recommend using lower case.
 #' @param title The human-facing title of the tutorial.
+#' @param learner_friendly If `TRUE`, adds `learnr` to Imports instead of Suggests to install `learnr` alongside the tutorial package. Re-exports `run_tutorial()` to make it immediately available to tutorial package users.
```
jennybc: Did you have any other ideas for argument name? My thought is that it could be more factual / descriptive. Because who wouldn't want to be `learner_friendly`? It feels odd to default to something that seems learner-unfriendly.

A possible re-framing: instead of the new argument being `TRUE` / `FALSE` and learner friendly vs ?unfriendly?, it could take values "Suggests" vs. "Imports". That might suggest another name for the argument.

jennybc:
@irenetlv BTW can you grant "maintainer can modify" permission in this PR?

https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork

I'm curious: do you remember explicitly disallowing that? I _think_ the default is for it to be allowed. I'm asking because of baseline curiosity about this setting and the `pr_*()` family of functions, not motivated by anything specific to you or this PR.

inst/templates/run-tutorial.R
isteves: @jennybc I copied this bit from @lionel-.  This is supposed to create a documentation page for each re-exported function (in this case just one) that links to the original function.

https://github.com/tidyverse/purrr/blob/master/R/reexport-rlang.R

R/tutorial.R
isteves: Our thought process behind this is that `learner_friendly` would address two types of users:

1. FALSE (lightweight option) -- for package developers who want to include tutorials as documentation but not absorb extra dependencies
2. TRUE (heavyweight option) -- for tutorial-only packages (for educators, etc)

We wanted a function name that would make it clear what this option was meant for.  Perhaps using the default `lightweight = TRUE` could be relatively neutral (unlike `unfriendly`) and still evocative?  @angela-li perhaps you have some thoughts on this?

isteves:
@jennybc I was actually working on that fork from this account (isteves) but somehow managed to accidentally make a PR with my work account (irenetlv).  (I had no idea that was even possible.)  Could that have also contributed to the mess?  Angela and I have edit permissions on Julie's fork.

Would it help if @jules32 opened a new PR for this?

inst/templates/run-tutorial.R
jennybc: Ok so I guess these are two different styles of exposing a function from package B in package A.

That makes my question/comment: are you sure you want this style as opposed to the one used by other re-export helpers in usethis? I don't have an opinion here, just want to know that you like the end result.

jules32:
Hi All, I'm back – please let me know if I need to open a new PR or followup in another way!

jennybc:
The main things I care about are the substance of the review and having "maintainer can modify" switched on for the PR.

jules32:
So unfortunately I do not have the option to "Allow edits from maintainers" as illustrated in the help (https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so I'm not sure if it's something I control?

image

jennybc:
Maybe when this gets revised, it should be as a new PR, given the puzzling nature of this one, in the sense that no one appears to have the power to turn on "maintainer can modify".

hadley:
I'm going to close this PR since it seems like its fundamentally stuck. I'm sorry it didn't get merged into usethis, but hopefully everyone learned something along the way 😄

jules32:
Good idea. Definitely learned a lot, but sorry we (I) couldn't help take it to the finish line

isteves:
Yep definitely did!  Sorry I also wasn't able to move it forward, but I guess a good lesson is to stick with standard PR flows 😉

PR #835, opened 2019-07-12
Fix use_circleci

jdblischak:
I used `use_circleci()` today and found a few bugs:

* The mustache delimiters caused the cache keys to be corrupted. I fixed this by using alternative delimiters, as suggested [here](https://mustache.github.io/mustache.5.html)

* `R CMD build` mentioned that it removed the empty `.circleci` directory. Apparently ignoring the only file it contains doesn't also ignore the entire directory. Thus I also explicitly ignored the directory itself

I added tests for both of these updates and also ran styler on the code changes.


jdblischak:
@jennybc Thanks for cleaning up my PR and merging it!

Issue #834, opened 2019-07-11
use_github() enterprise API problem 


pieterjanvc:
Hi,

I noticed that when trying to us the use_github() function with an enterprise account, it will give me an error saying 
```
Error: GitHub `auth_token` is invalid.
See `browse_github_token()` for help storing a token as an environment variable.
```
After a long search I found out this is because I have an enterprise account through my university and it uses a different host address. When I do use the host api in the function (e.g.  use_github(host = "github.uc.edu/api/v3") ), it still produces the same error.

I know the token is working as I tested it with the gh function:

```r
gh::gh_whoami(.api_url = "https://github.uc.edu/api/v3")
```

I dug in the code and think I found the source of the problem in the following function:

```r
github_user <- function(auth_token = github_token()) {
  suppressMessages(
    tryCatch(gh::gh_whoami(auth_token), error = function(e) NULL)
  )
}
```
This will generate a NULL value if the correct host variable is not given. I think it can simply be fixed by adding the .api_url = host (where host is the variable specified in the use_github ) like so:

```r
github_user <- function(auth_token = github_token()) {
  suppressMessages(
    tryCatch(gh::gh_whoami(auth_token, .api_url = host), error = function(e) NULL)
  )
}
```
Let me know what you think!

Grtz,
PJ

cderv:
I don't think this is really documented but `usethis` should be compliant with `gh` default behaviour. And you can configure an environment variable for `gh` to change the default url. 
Try: 
```r
Sys.setenv(GITHUB_API_URL="https://github.uc.edu/api/v3")
gh::gh_whoami()
```
This should work - it works for me.

I looked for documentation on this and there is none at my knowledge, unless you read code, PR and commits. 

I don't know if this could be change in usethis. I find it convenient that it uses the default behavior of `gh` and not introduce more arguments. 
However, I think this should be documented in usethis and gh.

pieterjanvc:
Thank you so much! 
It worked like a charm!!
But I agree it should be documented better so others will not suffer over it anymore :)

Thanks
PJ

cderv:
This is now documented in `gh`. 

For usethis, I see two things

* Document that `gh` default behaviour is used with `GITHUB_API_URL` env var.
* add a host argument in `github_user()` as you suggest because every other github funtion in usethis have a host argument.

cderv:
After a tour in the code, I think compatibility with github enterprise is not completely full yet.

[...]

PR #836, opened 2019-07-13
GitHub host improvement handling for GHE

cderv:
This is to see what it would mean to support GHE on usethis. a `host` argument has been introduce in some functions, but not all. This follow an issue in #834 that has been solved by providing `GITHUB_API_URL` env var that `gh` depends on. 

usethis can also depends on that by default as it uses `gh` and not offer a `host` argument in each function. It could also take the same approach as for token and check internally `github_api_host()` for this environment variable, default to GH if not set. 

```r
github_api_host() <- function() {
  # NULL means use gh default
  Sys.getenv("GITHUB_API_URL", NULL)
}
```
As `gh` does already this, this could be unnecessary in usethis.

Adding a host argument everywhere it is required seems to be a lot (from what I found already - see the changes)

Some functions (`pr_create_gh`) are using a url not configurable for now and won't work with GHE. It is not the API url but the server url. A `GITHUB_URL` env var maybe ? 🤔 

This is just to put the thoughts somewhere. Hence the draft PR for now.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -178,7 +178,7 @@ create_from_github <- function(repo_spec,
   create_directory(repo_path)
   check_directory_is_empty(repo_path)
 
-  auth_token <- check_github_token(auth_token, allow_empty = TRUE)
+  auth_token <- check_github_token(auth_token, allow_empty = TRUE, host = host)
 
   gh <- function(endpoint, ...) {
     gh::gh(
@@ -191,7 +191,7 @@ create_from_github <- function(repo_spec,
 
   repo_info <- gh("GET /repos/:owner/:repo", owner = owner, repo = repo)
 
-  fork <- rationalize_fork(fork, repo_info, auth_token)
+  fork <- rationalize_fork(fork, repo_info, auth_token, host = host)
   if (fork) {
     ## https://developer.github.com/v3/repos/forks/#create-a-fork
     ui_done("Forking {ui_value(repo_info$full_name)}")
@@ -278,11 +278,11 @@ check_not_nested <- function(path, name) {
   invisible()
 }
 
-rationalize_fork <- function(fork, repo_info, auth_token) {
+rationalize_fork <- function(fork, repo_info, auth_token, host = NULL) {
   have_token <- have_github_token(auth_token)
   can_push <- isTRUE(repo_info$permissions$push)
   repo_owner <- repo_info$owner$login
-  user <- if (have_token) github_user(auth_token)[["login"]]
+  user <- if (have_token) github_user(auth_token, host = host)[["login"]]
 
   if (is.na(fork)) {
     fork <- have_token && !can_push
@@ -290,7 +290,7 @@ rationalize_fork <- function(fork, repo_info, auth_token) {
 
   if (fork && !have_token) {
     ## throw the usual error for bad/missing token
-    check_github_token(auth_token)
+    check_github_token(auth_token, host = NULL)
   }
 
   if (fork && identical(user, repo_owner)) {
--- a/R/git.R
+++ b/R/git.R
@@ -402,6 +402,7 @@ have_git2r_credentials <- function() rlang::env_has(git2r_env, "credentials")
 #'   `my_cred`.
 #'
 #' @inheritParams git_protocol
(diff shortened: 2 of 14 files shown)


ijlyttle:
This is a very interesting topic to me; just a small note to commit (at least to myself) to contribute my thoughts here in the next couple days :)

ijlyttle:
First to apologize provisionally to @cderv, as this post seems like it is hijacking the thread. I am happy to remove this and to post it elsewhere, at the group's suggestion.

I have been thinking about an approach is likely too ambitious to be practical; I can appreciate the discipline of taking modest steps. Perhaps I can present it here, among colleagues, in the hopes that more discussion will follow towards the right practical solution.

There are a few r-lib packages (usethis, remotes, and pak come to mind; there may be others) that rely on or interact with git and github. I think that this idea could be used by other packages, should they wish.

I wonder if there could be a list of "r-lib" github options, like follows:

```r
rlib.github.default <- 
  list(
    service = "github",
    name = "GitHub"
    host_git = "github.com",
    url_api = "https://api.github.com",
    url_web = "https://github.com",
    env_token = c("GITHUB_PAT", "GITHUB_TOKEN")
  )
```

I have "renamed" some of the elements to try to identify more clearly what each element does; I note that the definition of `host` can vary from function to function (e.g. `remotes::install_github()`, `usethis::create_from_github()`, `usethis::browse_github_pat()`).

Perhaps this option could be a hardwired default, so that if someone wanted to use an instance of GitHub Enterprise (GHE), they could add something like this to their `.Rprofile`:

```r
options(
  rlib.github = list(
    list(
      service = "ghe_acme",
      name = "Acme GitHub Enterprise"
      host_git = "github.acme.com",
      url_api = "https://github.acme.com/api/v3",
      url_web = "https://github.acme.com",
      env_token = "GITHUB_PAT_ACME"
    )
  )
)
```

The overall idea is that public-facing functions that touch github (public or GHE) could add a `service` argument that could be used to determine the rest of the information. For organizations that use GHE, it would be their responsibility to publicize internally the "thing" that folks would put in their `.Rprofile`.

I think of functions in four categories:

- creating at github: `usethis::use_github()`
- creating from github: `usethis::create_from_github()`, `remotes::install_github()`
- interacting with github: `usethis::pr_*()`
- utilities: `usethis::github_token()`, `usethis::browse_github_token()`, `usethis::git_sitrep()`

For creating at github, I might suggest a default of `service = NULL`. If there is only one service available (github), it would use it. If there are multiple services available, and none is specified, the user could be told to supply an explicit choice.

For creating from github, I perhaps `service = "github"` could work as a default. I could also see wanting to be explicit.

For interacting with github, the `pr_*` functions, I don't think a `service` argument would be appropriate. Instead, I think these functions could look at the git-remote address for `origin` or `upstream`, e.g. `git@github.com:r-lib/usethis.git`, pull out the `host_git`, then use the matching service.

For the utility functions, I think `service = "github"` would work OK. For `usethis::git_sitrep()`, perhaps it could report on the status of all the registered services, and let the user know which of the services is being used for a local repo.

To wrap up, I recognize that this idea may not be a practical way forward, and I am happy to move the discussion to wherever it should go...

jennybc:
Thanks to you both for pushing this along.

@ijlyttle Your proposal sounds quite interesting and feels compatible with where we are w.r.t. the `auth_token` and git credentials. I think you should open a new issue and copy what you above there, as a better place to store it.

The next big thing I will do here, after pushing a few more gargle-related things out, will be to (hopefully) move usethis onto gert + credentials, to eliminate a lot of our git credentials pain. I will be touching all the git bits here and that could be a time to do something like what you propose.

I'll also raise this with others re: whether we could have a multi-package approach to some of this.

ijlyttle:
@jennybc Will do, thanks!

jennybc:
The landscape has changed here and ultimately I think this problem will be solved in other ways. For now I'm leaving this open as an interesting resource for when I get to solving the GHE problem.

cderv:
Glad it is still useful 😄 

I have seen the different evolution from afar as I did not have time yet to contribute in discussion. I will have more soon if you need 👀 on anything.

Issue #781, opened 2019-06-14
Can't post release issue


lionel-:
My first problem is that usethis tries to post on origin rather than upstream. Since posting issues in forks is unlikely, would the following patch make sense?

```diff
--- a/R/release.R
+++ b/R/release.R
@@ -25,7 +25,7 @@ use_release_issue <- function(version = NULL) {
   checklist <- release_checklist(version)
 
   issue <- gh::gh("POST /repos/:owner/:repo/issues",
-    owner = github_owner(),
+    owner = github_owner_upstream() %||% github_owner(),
     repo = github_repo(),
     title = glue("Release {project_name()} {version}"),
     body = paste(checklist, "\n", collapse = "")
```

Then when I try with this patch, I get:

```
Error in gh_process_response(raw) : 
GitHub API error (404): 404 Not Found
Message: Not Found
Read more at https://developer.github.com/v3/issues/#create-an-issue

URL not found: https://api.github.com/repos/r-lib/vctrs/issues
```

Not sure what to do from here.

jennybc:
OK I'll have a look at what's wrong.

But, as you have discovered, this function clearly currently expects that you are working on a repo that you "own", i.e. that `origin` is the repo where the issue should be opened. It's pretty rare for someone to be in a fork **and** have the authority to release a package. I think we (members of tidyverse & r-lib) should usually be working in the main repo and not have personal forks.

lionel-:
I prefer working from personal forks, this keeps the main repo tidy (we often end up with dangling branches otherwise because the author forgot to remove it, and the other members don't know what it's about). Also when I own the repo (i.e. personal repos in `github.com/lionel-`), I still call it `upstream`. IIRC this strategy simplified github management in my Emacs config (mainly for interacting with the main github repo). Maybe I could find a way to change it.

jennybc:
No, I think you're right that it should prefer `upstream` over `origin`, if both exist. It just doesn't come up in my personal workflow, which is why I've never noticed.

lionel-:
I see the same issue with:

```r
usethis::use_tidy_thanks(from = "v0.3.1", to = "v0.4.0")
#> Resolving timestamp for ref 
#> Error in parse_github_remotes(url)[[1]] : subscript out of bounds
```

The error goes away when I temporarily rename `upstream` to `origin`.

jennybc:
Yeah it makes sense that your workflow continues to expose cases where we've over-customized to the remote set up Hadley and I seem to share.

[...]

PR #838, opened 2019-07-16
WIP: Introduce concept of main and user repositories

lionel-:
Part of #781.

I'm currently creating a new package so I thought I'd go ahead and try to make usethis compatible with the origin + upstream workflow.

One idea explored in this tentative PR is to introduce the concepts of main and user repos.

- The main repo is where official development takes place. By default it is `upstream || origin`.
- The user repo is where local development takes place. By default it is `origin || upstream`.

For instance, Travis activation should use the main repo by default (it is occasionally useful to activate it in the user repo but I'm not sure how to accomodate this with the current UI). On the other hand, checking if Github is properly configured should be done with the user repo. I expect most usethis functions should interact with the main repo rather than the user repo.

### DIFF
--- a/R/ci.R
+++ b/R/ci.R
@@ -49,7 +49,8 @@ use_travis_badge <- function(ext = "org") {
 }
 
 travis_activate <- function(browse = interactive(), ext = "org") {
-  url <- glue("https://travis-ci.{ext}/profile/{github_owner()}")
+  owner <- github_main_owner()
+  url <- glue("https://travis-ci.{ext}/profile/{owner}")
 
   ui_todo("Turn on travis for your repo at {url}")
   if (browse) {
--- a/R/github-utils.R
+++ b/R/github-utils.R
@@ -40,14 +40,21 @@ github_upstream <- function() {
   parse_github_remotes(r)[[1]]
 }
 
+# The main repository is upstream || origin.
+# The user repository is origin || upstream.
 github_main <- function() {
   r <- github_remote("upstream") %||% github_remote("origin")
   if (is.null(r)) return(r)
   parse_github_remotes(r)[[1]]
 }
+github_user <- function() {
+  r <- github_remote("origin") %||% github_remote("upstream")
+  if (is.null(r)) return(r)
+  parse_github_remotes(r)[[1]]
+}
 
 github_owner <- function() {
-  github_main()[["owner"]]
+  github_user()[["owner"]]
 }
 
 github_owner_upstream <- function() {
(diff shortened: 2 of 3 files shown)


jennybc:
Just recording some of my own thoughts to park here.

I see three different scenarios:

  * Maintainer 1:
    - User is package maintainer or has write or admin permissions.
    - `origin` is the main repo. If there are other remotes, they are contributor remotes. No `upstream`.
    - Jenny and Hadley usually work this way.
  * Maintainer 2:
    - User is package maintainer or has write or admin permissions.
    - `origin` is a personal fork. `upstream` in the main or source repo.
    - Lionel usually works this way.
  * Contributor:
    - User does not have write or admin permissions.
    - `origin` is a personal fork. `upstream` in the main or source repo.
    - Typical contributor.

Basically our mentality so far has accounted for Maintainer 1 and Contributor. We didn't really plan for Maintainer 2.

hadley:
If we go down this route, I think it probably makes sense to eliminate `github_upstream()` (and friends) and only refer to `github_main()` or `github_user()`.

hadley:
Or maybe `github_upstream()` should always fall back to origin if upstream doesn't exist?

@lionel- when would you expect an upstream remote without an origin remote?

lionel-:
If I have created the github repo instead of cloning it, I would call it "upstream" and not "origin". I would only call clones "origin". What do you think about these conventions @jimhester?

jimhester:
The most widely used convention and what git defaults to is.

1. The place you push to is `origin`
2. The place you pull from is `upstream`
3. If there is only one of the above it is `origin`

I think doing anything else in usethis will lead people down a bad path.

lionel-:
That makes sense. This suggests these rules for the main and user repos:

- Main repo: `upstream || origin`
- User repo: `origin`

Note that this is compatible with the rules suggested above:

- Main repo: `upstream || origin`
- User repo: `origin || upstream`

hadley:
I think maybe using "push" and "pull" as name might help us remember. (But the main change we need to make is for upstream to always fall back to origin)

jennybc:
> I think doing anything else or using other terms in usethis will lead people down a bad path.

I agree that adopting any sort of nonstandard convention, no matter how sane, will ultimately cause more pain than good. I tried to have my own convention around remote names for a while and eventually just gave up.

> I think maybe using "push" and "pull" as name might help us remember. (But the main change we need to make is for upstream to always fall back to origin)

One wrinkle is also that the push/pull setup will vary by branch. In a fork, `master` will pull from `upstream` (at least it will if it was created by `create_from_github()`. But `master` will push to `origin` (and this push should really never happen because no one should ever work in `master` of a fork). Non-`master` branches will pull from **and** push to `origin`.

jennybc:
This is now handled by the changes that came as part of the huge gert PR #1144

lionel-:
Thanks Jenny!

PR #840, opened 2019-07-17
Add use_ui(), imports usethis ui_* in your package

CorradoLanera:
User interfaces provided by **usethis** are very useful and beautiful!

However, to use them developing a package, one has to type `usethis::use_*()` every single time. On the one hand, that is quite frustrating; on the other hand, it produces toooo long lines on common entries, e.g.:
```
usethis::ui_stop('field {usethis::ui_field("foo")} must has value {usethis::ui_value(value)}, ...')
```
Which is a small example, but with width 100, and even split into three lines, e.g.
```
usethis::ui_stop(
    'field {usethis::ui_field("foo")} must has value {usethis::ui_value(value)}, ...'
)
```
The central one reach 84. So, if one would stay on 80 (or less), that snippet likely becomes like
```
usethis::ui_stop(paste(
    'field {usethis::ui_field("foo")} must has value',
    ' {usethis::ui_value(value)}, ...'
))
```

:-) :-) :-) quite fanny :-) :-) :-)

with `use_ui()` it would be simply:

```
ui_stop('field {ui_field("foo")} must has value {ui_value(value)}, ...')
```

73, wonderful! (And much more readable too!)

I have also written some tests for it, and test it locally on one of the packages I'm developing. It seems to work fine.

For the documentation, code, and tests, I tried to use style and functions used for `use_pipe()` and `use_tibble()` too.


jennybc:
I'm glad you are finding the `ui_*()` functions useful!

I'm not going to merge it for two reasons:

1. The `ui_*()` family is slated to move elsewhere, probably to the [cli package](https://github.com/r-lib/cli), so I don't want to accumulate more code around it here.
1. I'm not convinced the usage of these functions is "special" enough to get this treatment. For example, if they get it, why not `use_template()` as well? The other related functions (`use_tidy_eval()`, `use_pipe()`, `use_tibble()`) are all examples where there's something a bit tricky about it, that maintainers often miss. Also they are examples of re-exporting functionality from a different package.

I hope that makes sense. Thanks for working this up, even though this won't go in, it's part of the GitHub history and others can find it as a good example of one way to adopt the `ui_*()` functions in a package.

PR #842, opened 2019-07-18
Add `use_lifecycle()`

lionel-:
`use_lifecycle()` adds a dependency on the upcoming lifecycle package. It also imports the SVG badges to `man/figures`, so they can be displayed in the documentation.

I didn't know where to put this, so I have added an `r-lib.R` file. Perhaps there is a more appropriate place?


R/helpers.R
```diff
@@ -104,3 +104,21 @@ view_url <- function(..., open = interactive()) {
   }
   invisible(url)
 }
+
+use_rd_macros <- function(package) {
+  if (desc::desc_has_fields("RdMacros")) {
```
jennybc: Should this also have `file = project_get()`?

R/r-lib.R
```diff
@@ -0,0 +1,49 @@
+#' Helpers from the r-lib organisation
```
jennybc: I propose we call this file `lifecycle.R` and reword the docs to talk about the "lifecycle terminology used by the tidyverse and r-lib packages".  This way we explain where it comes from but others are welcome to use it too.

I think this link should appear somewhere:
https://www.tidyverse.org/lifecycle/

R/r-lib.R
```diff
@@ -0,0 +1,49 @@
...
+
+#' @rdname r-lib
+#' @export
+use_lifecycle <- function() {
+  check_is_package("use_lifecycle()")
+
+  use_package("lifecycle")
+  use_rd_macros("lifecycle")
+
+  dest_dir <- fs::dir_create(fs::path(proj_get(), "man", "figures"))
```
jennybc:
```suggestion
  dest_dir <- create_directory(proj_path("man", "figures"))
```

R/r-lib.R
```diff
@@ -0,0 +1,49 @@
...
+  templ_dir <- fs::path_package("usethis", "templates")
```
jennybc:
```suggestion
  templ_dir <- path_package("usethis", "templates")
```

We use fs so much I've imported the whole namespace.

R/r-lib.R
```diff
@@ -0,0 +1,49 @@
...
+  templ_files <- fs::dir_ls(templ_dir, glob = "*/lifecycle-*.svg")
```
jennybc:
```suggestion
  templ_files <- dir_ls(templ_dir, glob = "*/lifecycle-*.svg")
```

R/r-lib.R
```diff
@@ -0,0 +1,49 @@
...
+
+  purrr::walk(templ_files, fs::file_copy, dest_dir, overwrite = TRUE)
```
jennybc:
```suggestion
  purrr::walk(templ_files, file_copy, dest_dir, overwrite = TRUE)
```

tests/testthat/test-r-lib.R
```diff
@@ -0,0 +1,28 @@
+context("r-lib")
+
+test_that("use_lifecycle() imports badges", {
+  scoped_temporary_package()
+  use_lifecycle()
+  expect_proj_file("man", "figures", "lifecycle-stable.svg")
+
+  # Idempotent
+  expect_error(use_lifecycle(), regexp = NA)
```
jennybc:
```suggestion
  expect_error_free(use_lifecycle())
```

tests/testthat/test-r-lib.R
```diff
@@ -0,0 +1,28 @@
+context("r-lib")
```
jennybc: If we *don't* put lifecycle in Suggests, all 3 of these tests need to mock `usethis:::is_installed` the way we do for packages like learnr and RcppArmadillo:

https://github.com/r-lib/usethis/blob/187bad13a10c8a85e517205936d45338978b1cb6/tests/testthat/test-use-tutorial.R#L14

https://github.com/r-lib/usethis/blob/187bad13a10c8a85e517205936d45338978b1cb6/tests/testthat/test-use-rcpp.R#L24

DESCRIPTION
```diff
@@ -41,13 +41,16 @@ Imports:
 Suggests: 
     covr,
     knitr,
+    lifecycle,
```
jennybc: I'm not sure usethis itself needs to add lifecycle to Suggests?

lionel-: Since we link to a documentation topic in the lifecycle package, not having it in Suggest causes a CMD check failure on Travis.

lionel-: oops actually we now only link to the website, so we can remove it!

lionel-:
Thanks for the review @jennybc, I think all is now addressed.

jennybc:
Thanks!

PR #862, opened 2019-08-02
Restrict the number of file names provided in `use_r()` and `use_test()`

### LINKED ISSUE #861: `use_r()` and `use_test()` fail when multiple names are provided in the `name` argument
When providing a vector input with which its length is greater than one, `use_r()` throws a warning and create files with faulty names. For instance;

```r
usethis::use_r(c("fileA", "fileB"))
#> ● Modify 'R/fileA.R', 'R/fileB.NA'
#> Warning messages:
#> 1: In if (!valid_file_name(path_ext_remove(name))) { :
#>   the condition has length > 1 and only the first element will be used
#> 2: In if (dir_exists(path)) { :
#>   the condition has length > 1 and only the first element will be used
#> 3: In if (!file.exists(filePath)) { :
#>   the condition has length > 1 and only the first element will be used
```
Here above, `use_r()` creates two files; the first one is fine, but the second one is created with an `NA` extension. Besides that, such a warning is thrown into the R console.

If we provide the same input to `use_test()`, it fails with the following error message:

```r
usethis::use_test(c("fileA", "fileB"))
#> ✔ Writing 'tests/testthat/test-fileA.R', 'tests/testthat/test-fileB.NA'
#> Error in file(path, open = file_mode, encoding = "utf-8") : 
#>   invalid 'description' argument
#> In addition: Warning messages:
#> 1: In if (!valid_file_name(path_ext_remove(name))) { :
#>   the condition has length > 1 and only the first element will be used
#> 2: In if (file_exists(proj_path(path))) { :
#>   the condition has length > 1 and only the first element will be used
#> 3: In if (!file_exists(path)) { :
#>   the condition has length > 1 and only the first element will be used
#> 4: In if (!file_exists(path)) { :
#>   the condition has length > 1 and only the first element will be used
```

At the first glance, the problem seems to happen around that call, `check_file_name()`, which is placed in the `R/r.R` file.

The usethis version I run is `1.5.1.9000`.

strboul:
Although it will be quite useful that if we have a way to create multiple files at once - I might use them as stubs and fill them in once I feel ready, it is not so easy to do the implementation at the moment.

So I thought that we can limit these calls that they will stop the execution if a vector provided with which its length is more than one. That limitation is also aligned with the current API of these functions because the main argument either in `use_r()` or `use_test()` states 'name' (not the plural version 'names').

This is my first PR in usethis. I am happy to modify my changes accordingly based on your review.

Resolves #861


R/r.R
```diff
@@ -8,6 +8,7 @@
 #'   Packages](https://r-pkgs.org).
 #' @export
 use_r <- function(name = NULL) {
+  check_file_name_length(name)
```
jennybc: I think it would be pretty self-consistent of us to do something simpler in this case. I propose `stopifnot(is_string(name))` instead of a custom helper.

tests/testthat/test-use-r.R
```diff
@@ -5,3 +5,8 @@ test_that("use_r() creates a .R file below R/", {
   use_r("foo")
   expect_proj_file("R/foo.R")
 })
+
+test_that("use_r doesn't accept multiple file names", {
```
jennybc:
```suggestion
test_that("use_r() doesn't accept multiple file names", {
```

R/r.R
jennybc: Also, the check should happen after the `name <- name %||% ...` line, because `NULL` input is valid.

R/test.R
```diff
@@ -47,6 +47,7 @@ use_test <- function(name = NULL, open = interactive()) {
   if (is.null(name)) {
     name <- get_active_r_file(path = "R")
   } else {
+    check_file_name_length(name)
```
jennybc: This should be edited to look more like the comparable locus in `use_r()`. It looks like we could be using `%||%` here too, actually.

R/r.R
strboul: I agree all except that we use `stopifnot()`. A regular if-not-true-stop branch might be more useful to give better error messages for the end-users.

strboul: On the other hand, I see that `stopifnot(is_string(name))` is often used in the package to check inputs.

R/test.R
strboul: Good idea! Fixed in 147a626

hadley:
Thanks @strboul!

PR #869, opened 2019-08-08
Initial draft of azure-pipelines usethis functions

jimhester:
This is not really done, but it will give us a place to discuss how we want it to work, and let someone other than me try it out.

Right now it uses the [azure-tidyverse.yml](https://github.com/r-lib/r-azure-pipelines/blob/master/azure-tidyverse.yml) template, which is more equivalent to `use_tidy_ci()` in the builds and functionality it uses. Maybe that is OK, or maybe we should provide a more barebones configuration for `use_azure_pipelines()`?

The main manual steps you have to do is creating the azure project and (for users not in the r-lib org) specifying the endpoint to point to their endpoint's name. I am also not entirely sure if I should document these things in usethis or at https://github.com/r-lib/r-azure-pipelines, maybe the latter?


maxheld83:
Just a small thing: if *covr* isn't installed, `usethis::use_azure_pipelines()` fails (towards the end) with:

```
Error: 'covr' must be installed before you can take a dependency on it.
```

The working directory is already unclean at this point.
Perhaps *covr* could be asserted *before* the working directory is altered?

hadley:
Can we merge this so we can try out from dev version of usethis without having to install a branch?

jennybc:
> Can we merge this so we can try out from dev version of usethis without having to install a branch?

Sure, here we go.

PR #883, opened 2019-09-08
Allow invalid CRAN names

### LINKED ISSUE #528: Optionally allow non-CRAN names in create_package and use_description
It's often useful to have package structures or even just DESCRIPTION files in projects that aren't fully packages or will never go to a repository, but the namespace is only so big and underscores/dashes are useful in project names.  I would like a `check_name=TRUE` argument for `use_description()` and `create_package()` so I can skip the name-checking error and use these functions for more projects.  It could also default to `getOptions(usethis.checkname)`

Happy to write the PR for this if you think the feature should exist.

noamross:
- `stop_for_name` controls whether to throw error
  for invalid CRAN package names
- Default to warning for `use_description()`, error
  for `create_package()`
- Add tests
- update NEWS

I am unsure if `stop_for_name` is the best argument name, open to other suggestions!

Closes #528

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* `use_description()` and `create_package()` gain the argument `stop_for_name` to control whether to allow names invalid for CRAN (#883 @noamross)
+
 * New `use_lifecycle()` helper to import the lifecycle badges for functions and arguments in your package. See https://lifecycle.r-lib.org/.
 
 * `git_sitrep()` reports email(s) associated with your GitHub account(#724, @dragosmg).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # usethis (development version)
 
-* `use_description()` and `create_package()` gain the argument `stop_for_name` to control whether to allow names invalid for CRAN (#883 @noamross)
+* `use_description()` and `create_package()` gain the argument `check_name` to control whether to check for package names invalid for CRAN (#883, @noamross).
 
 * New `use_lifecycle()` helper to import the lifecycle badges for functions and arguments in your package. See https://lifecycle.r-lib.org/.
 
--- a/R/create.R
+++ b/R/create.R
@@ -30,14 +30,14 @@
 create_package <- function(path,
                            fields = NULL,
                            rstudio = rstudioapi::isAvailable(),
-                           stop_for_name = TRUE,
+                           check_name = TRUE,
                            open = interactive()) {
   path <- user_path_prep(path)
   check_path_is_directory(path_dir(path))
 
   name <- path_file(path)
-  if (stop_for_name) {
-    check_package_name(name, stop_for_name = stop_for_name)
+  if (check_name) {
+    check_package_name(name)
   }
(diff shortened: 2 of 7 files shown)


R/create.R
```diff
@@ -30,20 +30,23 @@
...
                            fields = NULL,
                            rstudio = rstudioapi::isAvailable(),
+                           stop_for_name = TRUE,
                            open = interactive()) {
   path <- user_path_prep(path)
   check_path_is_directory(path_dir(path))
 
   name <- path_file(path)
-  check_package_name(name)
+  if (stop_for_name) {
```
jennybc: Do we need this condition? Since we pass `stop_for_name` along, seems like we could execute `check_package_name()` unconditionally.

R/description.R
```diff
@@ -97,15 +99,20 @@ build_description_list <- function(fields = list()) {
...
       "* Start with a letter",
       "* Not end with '.'"
-    ))
+    )
+    if (stop_for_name) {
+      ui_stop(msg)
+    } else {
+      ui_warn(stop_for_name)
+    }
   }
```
jennybc: I lean towards only checking when `stop_for_name = TRUE`. If people go out of their way to call `stop_for_name = FALSE`, let's assume they know what they're doing.

This makes this function simpler and the tests.

tests/testthat/test-use-description.R
```diff
@@ -67,3 +67,13 @@ test_that("default description is tidy", {
...
   expect_identical(desc_lines_before, desc_lines_after)
 })
+
+test_that("valid CRAN names checked", {
+  withr::local_options(list(usethis.description = NULL, devtools.desc = NULL))
+  suppressWarnings(
+    scoped_temporary_package(dir = file_temp(pattern = "invalid_pkg_name"))
+  )
+  expect_warning(use_description())
+  expect_error(use_description(stop_for_name = TRUE))
```
jennybc: Look around at other `expect_error()` calls. In general we're checking class and/or against a message fragment.

R/description.R
noamross: Current behavior gives a warning because the default is `use_description(stop_for_name = FALSE)`, but making it TRUE (and called `check_name`) will probably reduce surprise.

R/create.R
noamross: I think the package name needs to be checked before the directory is created to avoid creating an invalid directory and then erroring out, but `use_description()` needs to be called after the directory is created.  I eliminating the warning (below), I've simplified the check below, so still need the conditional here.

tests/testthat/test-use-description.R
noamross: Done in the changes below.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* `use_description()` and `create_package()` gain the argument `stop_for_name` to control whether to allow names invalid for CRAN (#883 @noamross)
```
```suggestion
* `use_description()` and `create_package()` gain the argument `check_name` to control whether to check for package names invalid for CRAN (#883 @noamross)
```

jennybc:
Thanks!

PR #890, opened 2019-09-12
allow "." to be used with create_package()

### LINKED ISSUE #777: usethis::create_package() needs path, empty path does not create default path
Related to #92, help file suggests that create_package(), with no path specified, will create a path, but it does not. 

This might be the intended behavior, but I was assuming if I was in an empty R project, I could run create_package(), and it would create a package inside the project folder, using the project name. Instead, create_package("../projectName") was needed. 

Create an empty R project, then
```r
usethis::create_package() #doesn't create a new package with "default" path
Error in user_path_prep(path) : 
  argument "path" is missing, with no default
```

Not sure if #92 was closed because the proposed "." was implemented, but if so, that doesn't appear to work either:

```r
usethis::create_package(".")
Error: '.' is not a valid package name. It should:
* Contain only ASCII letters, numbers, and '.'
* Have at least two characters
* Start with a letter
* Not end with '.'
```

MilesMcBain:
- fixes #777 
- replaced `path_expand` with `path_real` in `user_path_prep`
- added a to check a valid package can be created using "." within a temp dir.

### DIFF
--- a/R/proj.R
+++ b/R/proj.R
@@ -150,7 +150,7 @@ proj_path_prep <- function(path) {
 user_path_prep <- function(path) {
   ## usethis uses fs's notion of home directory
   ## this ensures we are consistent about that
-  path_expand(path)
+  path_real(path)
 }
 
 proj_rel_path <- function(path) {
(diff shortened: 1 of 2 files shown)


hadley:
I'm going to close this because I think unfortunately that it requires considerably more thinking.

PR #902, opened 2019-09-28
Restrict options of snippet types

### LINKED ISSUE #885: [Feature request] Limit the values of `type` in  edit_rstudio_snippets() to reasonable options only
As I understand, RStudio has a limited number of different types of snippets:
snippet

But `edit_rstudio_snippets()` argument `type` can have any value. For example, once I forgot that I have to use "markdown" and not "rmarkdown" and created a wrong type of document. Snippets in it did not have an effect. 
I expect that only reasonable values of `type` should be accepted. In other cases, either a warning or an error message should be issued.

GegznaV:
Restrict options of snippet types and warn if the default snippets are masked.
Closes #885

### DIFF
--- a/R/edit.R
+++ b/R/edit.R
@@ -107,7 +107,7 @@ edit_rstudio_snippets <- function(type = "r") {
   if (!file_exists(path)) {
     ui_done("File for snippets created {ui_path(path)}")
     ui_warn(c("The default snippets for {ui_field(type)} may not work!",
-      "To restore the default snippets, this file should be deleted and RStudio restarted."))
+      "If it is needed to restore the default snippets, this file should be deleted and RStudio restarted."))
   }
   edit_file(path)
 }
--- a/R/edit.R
+++ b/R/edit.R
@@ -95,12 +95,15 @@ edit_r_makevars <- function(scope = c("user", "project")) {
 
 #' @export
 #' @rdname edit
-#' @param type Snippet type. One of: "R", "markdown", "C_Cpp", "Tex",
-#'   "Javascript", "HTML", "SQL"
-edit_rstudio_snippets <- function(type = "R") {
+#' @param type Snippet type. One of: "r", "markdown", "c_cpp", "css", "html",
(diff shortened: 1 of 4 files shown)


R/edit.R
```diff
@@ -95,12 +95,20 @@ edit_r_makevars <- function(scope = c("user", "project")) {
...
-#' @param type Snippet type. One of: "R", "markdown", "C_Cpp", "Tex",
-#'   "Javascript", "HTML", "SQL"
-edit_rstudio_snippets <- function(type = "R") {
+#' @param type Snippet type. One of: "r", "markdown", "c_cpp", "css", "html",
+#' "java", "javascript", "python", "sql", "stan", "tex" (case insensitive).
+edit_rstudio_snippets <- function(type = "r") {
   # RStudio snippets stored using R's definition of ~
   # https://github.com/rstudio/rstudio/blob/4febd2feba912b2a9f8e77e3454a95c23a09d0a2/src/cpp/core/system/Win32System.cpp#L411-L458
-  path <- path_home_r(".R", "snippets", path_ext_set(tolower(type), "snippets"))
+  available_types <- c("r", "markdown", "c_cpp", "css", "html", "java",
```
jennybc: Please move this set of values into the function signature, in the same style as:

https://github.com/r-lib/usethis/blob/a2342b8a937e32d17de01b29b42154e92ccb5e6b/R/rstudio.R#L43-L44

You can do the `tolower()` on user input like so:

``` r
...
type <- tolower(type)
type <- match.arg(type)
...
```

Then you can also remove that list from the documentation for the `type` parameter and eliminate the possibility of the docs and the code differing from each other. In fact, they disagree already: `"markdown"` appears in the code but not in the docs.

jennybc:
Can you also reword your PR description to use one of the Magic Words for closing the associated issue?

https://help.github.com/en/articles/closing-issues-using-keywords

R/edit.R
GegznaV: Done

jennybc:
Thanks!

PR #903, opened 2019-09-30
Make get_active_r_file() more robust to `~/` issues on Windows

### LINKED ISSUE #901: use_test() without arguments fails with RStudio project under ~ on Windows
File path checking seems to go wrong when `use_test()` tries to get the path to the active file in RStudio in a project under `~` on Windows:
``` r
usethis::use_r("foo.R")
# ✔ Setting active project to 'C:/Users/mikko/Documents/R/dev/foo'
# ● Modify 'R/foo.R'
usethis::use_test()
# ✔ Adding 'testthat' to Suggests field in DESCRIPTION
# ✔ Creating 'tests/testthat/'
# ✔ Writing 'tests/testthat.R'
# ● Call `use_test()` to initialize a basic test file and open it for editing.
# Error: Open file must be in the 'R/' directory of the active package.
# * Actual path: 'Users/mikko/R/dev/foo/R/foo.R'
rstudioapi::getSourceEditorContext()$path
# [1] "~/R/dev/foo/R/foo.R"
usethis:::proj_path_prep(rstudioapi::getSourceEditorContext()$path)
# C:/Users/mikko/Documents/R/dev/foo/Users/mikko/R/dev/foo/R/foo.R
```
Related https://github.com/r-lib/fs/issues/221. Maybe related https://github.com/r-lib/usethis/pull/485?

jennybc:
Closes #901

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* `use_test()` should work again on Windows, when called with no argument, i.e. when the active `.R` file is determined from RStudio's source editor context (#901).
+
 * `use_description()` and `create_package()` gain the argument `check_name` to control whether to check for package names invalid for CRAN (#883, @noamross).
 
 * New `use_lifecycle()` helper to import the lifecycle badges for functions and arguments in your package. See https://lifecycle.r-lib.org/.
--- a/R/proj.R
+++ b/R/proj.R
@@ -153,6 +153,25 @@ user_path_prep <- function(path) {
   path_expand(path)
 }
 
+## rstudioapi can return a path like '~/path/to/file' where '~' means
+## R's notion of user's home directory
+rstudioapi_path_prep <- function(path) {
+  if(!grepl("^~", path)) {
+    return(path)
+  }
+
+  possibilities <- c(path_expand(path), path_expand_r(path))
+
+  exists <- file_exists(possibilities)
+  if (!any(exists)) {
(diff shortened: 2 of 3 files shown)


jennybc:
OK I've simplified this, as there is only a need to ever call `path_expand_r()` on this particular path from rstudioapi. And _then_ we need to process with `proj_path_prep()`.

I think @jimhester's suggestion in #479 is a good one, as `path_real()` seems to expose us to various difficulties with mapped drives on Windows. I'll open a new issue to think harder about this and see what I would have to do to work around rprojroot's `normalizePath()` call.

dhofstetter:
Hello all together ,
is it possible that this changes broke the `usethis::use_test()` functionalities on Windows (usually this don't bother me because of using linux, but at work we can only use Windows)?

Actually at the moment when using the `use_test()` method within RStudio i receive an output like:

> >usethis::use_test()
>   Setting active project to 'C:/Users/username/Documents/Pkg_Dir_Name'
> Error: Open file must be in the 'R/' directory of the active package.
> * Actual path: 'Users/username/Pkg_Dir_Name/R/vector_util.R'

The file was opened in the editor from the R folder in the Files pane ... so it's definetly within the current package.

And interesting at all is `rstudioapi::getSourceEditorContext()$path` returns '~/Pkg_dir_name'

This all is quite strange, as I'm sure, that prior to my holidays it was all working like a charme.

BR Daniel

PR #910, opened 2019-10-13
Avoid byte-compiling when installing from RStudio

krlmlr:
This settings shaves off ~1 second or more for installing a package using Ctrl + Shift + B (Cmd + Shift + B).

### DIFF
--- a/inst/templates/template.Rproj
+++ b/inst/templates/template.Rproj
@@ -12,5 +12,5 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
-PackageInstallArgs: --no-multiarch --with-keep.source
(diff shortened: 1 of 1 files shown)


hadley:
Do we want to do this? We usually assume that Cmd + Shift + B provides a higher quality simulation —  wouldn't this change would mess up benchmarking?

krlmlr:
We could frame it like this:

- `load_all()` is for very frequent iterations when developing the package
- Ctrl + Shift + B is for testing a package you develop against other packages
- `install_git(".")` or `install_local()` is for creating a production-like install

This is for new templates only, in existing packages we'd need to reapply the template anyway. We can experiment a bit before updating the template. I could add an option that triggers this behavior, so that we can merge and enable/disable easier.

hadley:
We _could_ frame it that way, but we currently don't, so I'd prefer to leave off for now.

PR #913, opened 2019-10-14
Add `browse_circle()`

### LINKED ISSUE #914: Add the ability to browse builds on Circle CI
Similar to `browse_travis()`.

pat-s:
fixes #914


hadley:
Could you please start by filing an issue that describes the problem you're trying to solve?

pat-s:
Well, the context of this addition to the already existing functions including the change to `NEWS.md` should be descriptive enough. At least that's what I thought. 

A direct close is also answer enough. Sorry for contributing.

batpigandme:
Hi @pat-s,

Sorry, this is probably my fault for not having a clearer link to the  contributing guidelines. The most succinct version is in just the final section here: https://www.tidyverse.org/contribute/ (apologies for lack of formatting in this comment/not pulling the text, I'm on my phone right now). 

The flow is usually something to the effect of Issue pitching the idea (usethis is pretty broad, so I think this would fit in), then PR closing the issue. 

I don't use Circle CI, but I think I can tell what you're doing. Do you want me to open an issue for it? (You're right, the issue would basically be your news bullet, the order of operations our way of keeping track of things given the volume and number of repos etc).

pat-s:
Hi @batpigandme,

thanks for taking the time.
I am aware that my response was also a bit sensitive.

>  The flow is usually something to the effect of Issue pitching the idea (usethis is pretty broad, so I think this would fit in), then PR closing the issue.

Sure, I respect contributing guidelines. We also have some and I get quite some issues/PR per day. 

I am aware that "hard actions" might be taken if one does not comply with such. Yes, I was not aware of having to open an issue first in the moment when I opened that PR.
Still think that the behavior of a direct close here was a little rough.

> I don't use Circle CI, but I think I can tell what you're doing. Do you want me to open an issue for it? (You're right, the issue would basically be your news bullet, the order of operations our way of keeping track of things given the volume and number of repos etc).

I was really thinking of just letting it go at first, but thanks to your gentle reply I will open an issue and maybe this PR can be re-opened.

batpigandme:
Thanks, @pat-s!

R/browse.R
```diff
@@ -59,6 +59,14 @@ browse_travis <- function(package = NULL, ext = c("org", "com")) {
   view_url(sub("github.com", travis_url, gh))
 }
 
+#' @export
+#' @rdname browse-this
```
hadley: I think this would be clearer as `browse_circle_ci()` or `browse_circleci()`? (whichever punctuation you think is better)

R/browse.R
```diff
@@ -59,6 +59,14 @@ browse_travis <- function(package = NULL, ext = c("org", "com")) {
   view_url(sub("github.com", travis_url, gh))
 }
 
+#' @export
```
hadley: Can you please add a bullet point to the documentation above, following the existing pattern?

R/browse.R
```diff
@@ -59,6 +59,14 @@ browse_travis <- function(package = NULL, ext = c("org", "com")) {
...
+#' @rdname browse-this
+browse_circle <- function(package = NULL) {
+  gh <- github_home(package)
+  circle_url <- glue::glue("circleci.com/gh")
```
hadley: I don't think glue is needed here?

batpigandme: Would this make more sense?
```suggestion
  circle_url <- glue::glue("circleci.com/{gh}")
```

hadley: I don't think so — we're currently using `github_home()` as a round about way to get the github repo; I'll fix that problem after this PR is merged.

batpigandme: Ah, 10-4. My bad

R/browse.R
pat-s: I don't know; Travis functions also do not have the "ci" suffix.
I don't have a strong preference on this though.

R/browse.R
pat-s: Nothing is glued here so yes, this can just be normal string.

R/browse.R
jennybc: We have an existing `use_circleci()` function FWIW.

pat-s: Just saw it, will therefore use the `circleci` pattern.

hadley:
Thanks!

PR #916, opened 2019-10-17
Allow create_package() to use "." once more

### LINKED ISSUE #777: usethis::create_package() needs path, empty path does not create default path

hadley:
Fixes #777

I think this should be safe because it's only affects the name of the package, and it doesn't break any tests, but I'd appreciate a second pair of eyes

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -35,7 +35,7 @@ create_package <- function(path,
   path <- user_path_prep(path)
   check_path_is_directory(path_dir(path))
 
-  name <- path_file(path)
+  name <- path_file(path_real(path))
   if (check_name) {
     check_package_name(name)
   }
--- a/tests/testthat/test-create.R
+++ b/tests/testthat/test-create.R
@@ -37,6 +37,12 @@ test_that("nested project is disallowed, by default", {
   expect_usethis_error(scoped_temporary_project(path(dir, "abcde")), "anyway")
 })
 
+test_that("can create package in current directory", {
+  dir <- dir_create(path(file_temp(), "mypackage"))
+  withr::local_dir(dir)
+  expect_error_free(create_package("."))
+})
+
 ## https://github.com/r-lib/usethis/issues/227
 test_that("create_* works w/ non-existing rel path and absolutizes it", {
   ## take care to provide a **non-absolute** path


MichaelChirico:
Would benefit from this. I reported in #977:

My workflow was:

   1. Navigate to my github directory cd ~/github
   2. Create my pkg dir: mkdir pkg
   3. RStudio : File > New Project... > Existing Directory > [select pkg]
   4. In new RStudio session within pkg.Rproj, try to run usethis::create_package('.'), but it failed.

This issue would allow this workflow which doesn't seem to crazy to me.

PS I also filed an incidental improvement to check_valid_package(), feel free to add that here or I can file a separate PR

jennybc:
Checking something else, I just used `create_project(".")`, which *does* work.

But has some regrettable path infelicities in the messaging. For lack of a better place to record this, whenever I come through a check/merge this, I should check the paths in the messages for both `create_project(".")` and `create_package(".")`.

---

Inside RStudio:

``` r
> setwd("~/tmp/xyz")
> create_project(".")
✔ Setting active project to '/Users/jenny/tmp/xyz'
✔ Creating 'R/'
✔ Writing 'xyz.Rproj'
✔ Adding '.Rproj.user' to '.gitignore'
✔ Opening './' in new RStudio session
✔ Setting active project to '<no active project>'
```

This is not so great: "Opening './' in new RStudio session".
Ideally it would say: "Opening 'xyz' in new RStudio session", I think?

---

Running R in a shell:

```r
> setwd("~/tmp/xyz")
> usethis::create_project(".")
✔ Setting active project to '/Users/jenny/tmp/xyz'
✔ Creating 'R/'
✔ Writing a sentinel file '.here'
● Build robust paths within your project via `here::here()`
● Learn more at <https://here.r-lib.org>
✔ Changing working directory to './'
```

Similar problem with "Changing working directory to './'".

Issue #771, opened 2019-05-24
Rename use_tidy_versions()


jennybc:
`use_tidy_versions()` is *not* parallel to `use_tidy_ci()` or `use_tidy_github()`, i.e. it's not enforcing a policy we embrace unconditionally. Aggressive minimum versions are another feature that can make a package feel like a heavy dependency (in addition to having lots of dependencies).

jennybc:
This should possibly get split into 2 functions, instead of 1 function with `source "local"` vs `source = "CRAN"`.

Brainstorming names:

  * `use_installed_versions()`
  * `use_local_versions()`
  * `use_cran_versions()`
  * `use_released_versions()`

hadley:
I gave relatively little thought to the name because I don't think it's something anyone should be doing routinely. I added `@keyword internal` to hopefully reduce the number of people who stumble across it accidentally.

PR #917, opened 2019-10-17
Generate absolute link to CoC

### LINKED ISSUE #772: Code of Conduct link in README needs to be full URL
There's a new CRAN check that gets triggered by README links to files below `.github/`, such as the code of conduct.

Make sure usethis's tools replace this internal link with a full URL. Probably to the file on GitHub? The one I just did "by hand" links to the file in the pkgdown site, which is probably a nice solution, but that can't really be the default strategy.

hadley:
Fixes #772


R/code-of-conduct.R
```diff
@@ -7,6 +7,12 @@
 #' <https://contributor-covenant.org>, version 1:
 #' <https://contributor-covenant.org/version/1/0/0>.
 #'
+#' If your package is going to CRAN the link to the CoC in your README must
```
batpigandme:
```suggestion
#' If your package is going to CRAN, the link to the CoC in your README must
```

man/use_code_of_conduct.Rd
```diff
@@ -19,3 +19,10 @@ inappropriate behaviour. The template comes from
 \url{https://contributor-covenant.org}, version 1:
 \url{https://contributor-covenant.org/version/1/0/0}.
 }
+\details{
+If your package is going to CRAN the link to the CoC in your README must
```
batpigandme:
```suggestion
If your package is going to CRAN, the link to the CoC in your README must
```

R/code-of-conduct.R
```diff
@@ -7,6 +7,12 @@
...
+#' be an absolute link to a rendered website as `CODE_OF_CONDUCT.md` is not
+#' included in the package sent to CRAN. `use_code_of_conduct()` will
+#' automatically generate this link if you use pkgdown and have set the
+#' `url` field in `pkgdown.yml`; otherwise you'll need to fill it in yourself.
```
jennybc: Maybe we should default to the source of our template and always leave people in a working state?

So: <https://contributor-covenant.org/version/1/0/0>

R/latest-dependencies.R
```diff
@@ -0,0 +1,39 @@
+#' Use "latest" versions of all dependencies
+#'
+#' Pins minimum version of dependencies to latest (as determined by "source)"
+#' Useful for tidyverse package, but should otherwise be used with extreme
+#' care.
+#'
+#' @keywords internal
```
jennybc: Re: "extreme care". We could add a `@seealso` to R Packages > Package metadata > Dependencies> Versioning.

> Think carefully if you declare a minimum version. In some sense, the safest thing to do is to require a version greater than or equal to the package’s current version. For public work, this is most naturally defined as the current CRAN version of a package; private or personal projects may adopt some other convention. But it’s important to appreciate the implications for people who try to install your package: if their local installation doesn’t fulfill all of your requirements around versions, installation will either fail or force upgrades of these dependencies. This is desirable if your minimum version requirements are genuine, i.e. your package would be broken otherwise. But if your stated requirements have a less solid rationale, this may be unnecessarily conservative and inconvenient.
>
> In the absence of clear, hard requirements, you should set minimum versions (or not) based on your expected user base, the package versions they are likely to have, and a cost-benefit analysis of being too lax versus too conservative.

Ha! This TODO reveals R Packages should change in response to this PR:

> TODO: return here and insert concrete details closer to publication. Thing 1: mention relevant usethis function(s). Today that is `use_tidy_version()` but it needs some work (https://github.com/r-lib/usethis/issues/771). Thing 2: See if Kirill’s prototype of studying minimum version has matured into anything we can recommend for general use (https://github.com/r-lib/pillar/tree/c017f20476fce431ceee18cc3a637f7ed2884d3a/minver#readme)

R/latest-dependencies.R
```diff
@@ -0,0 +1,39 @@
...
+#' @export
+#' @param overwrite By default (`FALSE`), only dependencies without version
+#'   specifications will be modified. Set to `TRUE` to modify all dependencies.
+#' @param source Use "local" or "CRAN" package versions.
+use_latest_dependencies <- function(overwrite = FALSE, source = c("local", "CRAN")) {
```
jennybc: I would have done CRAN then local. Do we really want local to be the default?

R/latest-dependencies.R
```diff
@@ -0,0 +1,39 @@
...
+  deps <- desc::desc_get_deps(proj_get())
+  deps <- update_versions(deps, overwrite = overwrite, source = source)
+  desc::desc_set_deps(deps, file = proj_get())
+
+  invisible(TRUE)
+}
+
+update_versions <- function(deps, overwrite = FALSE, source = c("local", "CRAN")) {
```
jennybc: If we changed order of `source` values, would need to happen here too.

R/r.R
```diff
@@ -49,14 +49,14 @@ use_test <- function(name = NULL, open = NULL) {
       use_template(
         "test-example-2.1.R",
         save_as = path,
-        open = open
+        open = FALSE
```
jennybc: Why does this (and its friend just below) change?

tests/testthat/test-latest-dependencies.R
```diff
@@ -0,0 +1,22 @@
+test_that("use_tidy_versions() specifies a version for dependencies", {
```
jennybc:
```suggestion
test_that("use_latest_dependencies() specifies a version for dependencies", {
```

tests/testthat/test-latest-dependencies.R
```diff
@@ -0,0 +1,22 @@
...
+  use_package("desc")
+  use_package("withr", "Suggests")
+  use_package("gh", "Suggests")
+  use_latest_dependencies()
+  desc <- readLines(proj_path("DESCRIPTION"))
+  desc <- grep("usethis|desc|withr|gh", desc, value = TRUE)
+  expect_true(all(grepl("\\(>= [0-9.]+\\)", desc)))
+})
+
+test_that("use_tidy_versions() does nothing for a base package", {
```
jennybc:
```suggestion
test_that("use_latest_dependencies() does nothing for a base package", {
```

R/latest-dependencies.R
hadley: I don't think it matters (especially since I only accidentally included in this PR)

R/code-of-conduct.R
hadley: Oh, good idea!

R/code-of-conduct.R
```diff
@@ -24,10 +30,17 @@ use_code_of_conduct <- function(path = NULL) {
...
+  href <- project_pkgdown_url()
+  if (is.null(href)) {
+    ui_todo("Couldn't find pkgdown website; make sure to adjust block below")
+    href <- "{INSERT WEBSITE HERE}"
+  }
+  href <- paste0(href, "/CODE_OF_CONDUCT.html")
+
   ui_todo("Don't forget to describe the code of conduct in your README:")
   ui_code_block("
     Please note that the {ui_value(project_name())} project is released with a
```
hadley: Should we add a heading here?

jennybc: I don't have a strong opinion.

PR #922, opened 2019-10-17
Fix #918: add parameter remote to use_dev_package()

### LINKED ISSUE #918: Feature request: extend use_dev_package() beyond github.com
As far as I can tell, right now, `use_dev_package()` assumes that the remote is located at github.com, and it constructs a Remotes entry based on that assumption, e.g. `r-lib/usethis`.

To extend to other situations, could `use_dev_package()` be an additional argument, `remote` with a default of `NULL`? Of course, the default behavior would be the current behavior. A non-NULL `remote` provided to the function could become the entry in Remotes.

I would not think this useful for direct interactive use; my use case is to write a wrapper function for institutional use, for example:

```r
use_package_acme <- function(package) {

  remote <- get_acme_remote(package) # Acme's responsibility

  usethis::use_dev_package(package, remote = remote)
}
```

If you think this worthwhile, I'd be happy to make a PR. Thanks!

ijlyttle:
Fix #918, add `remote` as parameter to `use_dev_package()`:

- syntax for `remote` expected to follow the [remotes documentation](https://remotes.r-lib.org/articles/dependencies.html#other-sources).
- default behavior unchanged.
- no testing (I could not see infrastructure to build on), but *it works on my computer* ™️. 
- add NEWS item.

Happy to amend as you see fit.


antoine-sachet:
Why not automate the RemoteType detection from the installed package? It is really straightforward, see #970 for example code. This would make it more user-friendly IMHO.

ijlyttle:
@antoine-sachet: I read what you propose in #970. I think this could be helpful in a number of cases, as you support GitHub, GitLab, and BitBucket.

However, in my case, I am using GitHub Enterprise. I can use a variant of `devtools::install_guthub()` to install, but the best way for me to specify this as a `remote` is using a URL. For me, having this extra flexibility of a `remote` parameter remains useful.

jennybc:
Thanks!

PR #928, opened 2019-10-24
Use messages rather than cat to output

jimhester:
But output the messages on stdout instead of stderr, to avoid red text
in RStudio and the Windows GUI.

This works around the issues with `message()` in RStudio by outputting the results on standard output rather than stderr, but still lets us emit message conditions objects, so they can be easily muffled by users with `suppressMessages()` and friends.

I recently used a similar approach in vroom, I think it handles the largest downside of using `message()` while giving the benefits of muffling without having to resort to `capture.output()`.

I did not try to edit the relevant sections of principles.md in regards to this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -49,7 +49,7 @@ Suggests:
     styler (>= 1.0.2),
     testthat (>= 2.1.0)
 Remotes:
-    rlang
+    r-lib/rlang
 Encoding: UTF-8
 Language: en-US
 LazyData: true
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -54,4 +54,4 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 6.1.99.9001
+RoxygenNote: 6.1.1
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,7 +30,7 @@ Imports:
     git2r (>= 0.23),
     glue (>= 1.3.0),
     purrr,
-    rlang,
+    rlang (>= 0.4.1.9000),
     rprojroot (>= 1.2),
     rstudioapi,
     stats,
@@ -48,6 +48,8 @@ Suggests:
     spelling (>= 1.2),
     styler (>= 1.0.2),
     testthat (>= 2.1.0)
+Remotes:
+    rlang
 Encoding: UTF-8
 Language: en-US
 LazyData: true
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,7 @@
 * New `use_github_actions()`, `use_github_action_check_release()`, `use_github_action_check_full()`,
   `use_github_action_pr_commands()`, `use_github_actions_tidy()` to set up a GitHub Actions
   for a package (@jimhester).
+* usethis now outputs with message conditions on standard output. (@jimhester)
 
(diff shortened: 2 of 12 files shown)


jimhester:
Related to this I opened https://github.com/r-lib/rlang/issues/852, if rlang does add this behavior either as the default or via an option in the future the code here could be changed to use `rlang::inform()` directly.

jimhester:
As discussed in the workweek, once `rlang::inform()` does this for interactive sessions we can simplify this PR and use it.

jennybc:
So I should wait, yes?

I can imagine this will have some downstream consequences for anyone who has built upon usethis, so I should run revdep checks after we do this.

jimhester:
Now that this behavior is in rlang https://github.com/r-lib/rlang/commit/5a0b80a477c80313effb636e471b77a4015ed5a8 we can just use `rlang::inform()` directly.

@jennybc would you like me to run the reverse dependency checks and see how many dependencies would be affected by this change?

jennybc:
> would you like me to run the reverse dependency checks

Yes, that would be awesome.

jimhester:
I added the revdep results. The devtools issue is just due to API changes in the `create_package()` function in devel usethis, not due to changes in this PR.

Two packages failed tests due to the changes, `DataPackageR` and `portalr`, the minor diffs which fix the tests are below.

1. DataPackageR
```diff
diff --git a/tests/testthat/test-ignore.R b/tests/testthat/test-ignore.R
index 9834ee2..cb1afa4 100644
--- a/tests/testthat/test-ignore.R
+++ b/tests/testthat/test-ignore.R
@@ -13,6 +13,8 @@ test_that("use_ignore works", {
       r_object_names = c("cars_over_20")
     )
   )
-  expect_output(use_ignore(file = "mydata.csv",path = "inst/extdata"),"Adding 'mydata.csv' to 'inst/extdata/\\.gitignore'|Adding '\\^inst/extdata/mydata\\\\.csv\\$' to '\\.Rbuildignore'")
+  withr::with_options(c(crayon.enabled = FALSE), {
+    expect_message(use_ignore(file = "mydata.csv",path = "inst/extdata"),"Adding 'mydata.csv' to 'inst/extdata/\\.gitignore'|Adding '\\^inst/extdata/mydata\\\\.csv\\$' to '\\.Rbuildignore'")
+  })
   expect_message(use_ignore(),"No file name provided to ignore.")
 })
```

2. portalr
```diff
diff --git a/tests/testthat/test-01-data-retrieval.R b/tests/testthat/test-01-data-retrieval.R
index d6fcf43..75450c6 100644
--- a/tests/testthat/test-01-data-retrieval.R
+++ b/tests/testthat/test-01-data-retrieval.R
@@ -103,7 +103,7 @@ test_that("default data path functions work if unset", {
   expect_error(use_default_data_path())
 
   data_path <- tempdir()
-  expect_error(m <- capture_output(use_default_data_path(data_path)), NA)
+  expect_error(m <- capture_messages(use_default_data_path(data_path)), NA)
   expect_match(m, "Call `usethis::edit_r_environ\\(\\)` to open '.Renviron'")
   expect_match(m, "Store your data path with a line like:")
   expect_match(m, "PORTALR_DATA_PATH=")
```

jennybc:
So shall we merge this?

jimhester:
I think so, it would probably be good to have it marinating in devel for a bit before the next release.

jennybc:
OK, here we go. Thanks!

PR #929, opened 2019-10-24
Add support for GitHub actions

jimhester:
This adds basic support for GitHub actions, with a minimal workflow. Presumably we would also need to add the more complex workflow as well, (https://github.com/r-lib/actions/tree/master/examples#tidyverse-ci-workflow), either adapting `use_tidy_ci()` to use GitHub Actions or creating a new function...


R/write.R
```diff
@@ -70,6 +70,7 @@ write_union <- function(path, lines, quiet = FALSE) {
 write_over <- function(path, lines, quiet = FALSE) {
   stopifnot(is.character(lines), length(lines) > 0)
   path <- user_path_prep(path)
+  fs::dir_create(dirname(path))
```
jimhester: We need this because the `.github/workflows` directory likely won't exist, I don't believe it should cause any problems with existing code, we may also want to use this in `write_union()` above.

R/ci.R
```diff
@@ -290,3 +290,34 @@ use_azure_coverage_badge <- function() {
   img <- glue("https://img.shields.io/azure-devops/coverage/{github_repo_spec()}/2")
   use_badge("Azure pipelines coverage status", url, img)
 }
+
+use_github_actions <- function(name = "R") {
```
jimhester: I am not really sure what `name` should be by default. This is the name of the workflow, in GitHub's [starter workflows](https://github.com/actions/starter-workflows/tree/master/ci) these are all just the name of the language, which is why I used "R" for this here. But I am open to changing it to something more specific as well.

jimhester:
I have reworked this a bit, adding a `use_action()` function, which lets you easily use actions from https://github.com/r-lib/actions/tree/master/examples, so that we can define all actions in that repository, and don't have to include the templates in usethis.

R/ci.R
```diff
@@ -290,3 +296,93 @@ use_azure_coverage_badge <- function() {
   img <- glue("https://img.shields.io/azure-devops/coverage/{github_repo_spec()}/2")
   use_badge("Azure pipelines coverage status", url, img)
 }
+
+#' @section `use_github_actions()`:
+#' Adds a basic ‘r.yaml’ file to the `.github/workflows` directory of a
+#'  package. This is a configuration file for the GitHub Actions service'
```
jennybc:
```suggestion
#'  package. This is a configuration file for the GitHub Actions service.
```

Is there a good primary link for us to include re: GitHub Actions? Maybe this, which I see below?

<https://github.com/features/actions>

R/ci.R
```diff
@@ -290,3 +296,93 @@ use_azure_coverage_badge <- function() {
   img <- glue("https://img.shields.io/azure-devops/coverage/{github_repo_spec()}/2")
   use_badge("Azure pipelines coverage status", url, img)
 }
+
+#' @section `use_github_actions()`:
+#' Adds a basic ‘r.yaml’ file to the `.github/workflows` directory of a
```
jennybc:
```suggestion
#' Adds a basic ‘R.yaml’ file to the `.github/workflows` directory of a
```

R/ci.R
```diff
@@ -290,3 +296,93 @@ use_azure_coverage_badge <- function() {
...
+  name <- utils::URLencode(name)
+  url <- glue("{github_home()}/actions?workflow={name}")
+  img <- glue("{github_home()}/workflows/{name}/badge.svg")
+
+  use_badge("R build status", url, img)
+}
+
+#' @section `use_action()`:
+#' Use a specific action, either one fo the example actions from
+#'   [@r-lib/actions/examples](https://github.com/r-lib/actions/tree/master/examples) or a custom action
```
jennybc:
```suggestion
#'   [r-lib/actions/examples](https://github.com/r-lib/actions/tree/master/examples) or a custom action
```

R/write.R
jennybc: This comes up in many places and we've basically decided to keep `write_over()` and `write_union()` as dumb as possible. I think we're pretty consistent about making the caller responsible for creating the parent directory. We use the internal helper `create_directory()` everywhere for this.

R/ci.R
```diff
@@ -290,3 +296,93 @@ use_azure_coverage_badge <- function() {
...
+
+  if (!new) {
+    return(invisible(FALSE))
+  }
+  use_github_actions_badge("R")
+
+  invisible(TRUE)
+}
+
+use_tidy_actions <- function() {
```
jennybc: Is this meant to be exported?

R/ci.R
hadley: I think something around `R cmd check`would be useful, since we'd have different checks for different project types.

R/actions.R
```diff
@@ -0,0 +1,110 @@
...
+
+#' @rdname actions
+#' @export
+use_action_ci_full <- function(save_as = "R.yaml", ignore = TRUE, open = FALSE) {
+  use_action("ci-full.yaml", save_as = save_as, ignore = ignore, open = open)
+}
+
+#' @rdname actions
+#' @export
+use_action_pr_commands <- function(save_as = "pr-commands.yaml", ignore = true, open = FALSE) {
```
DavisVaughan: `ignore = true`! should be `ignore = TRUE`

NAMESPACE
```diff
@@ -60,6 +60,9 @@ export(ui_todo)
 export(ui_value)
 export(ui_warn)
 export(ui_yeah)
+export(use_action_ci_full)
```
hadley: I think it would better to use `check` here rather than `ci`, or maybe even `rcmdcheck`, because eventually we'll want `use_action_bookdown()`, `use_action_blogdown()` etc

jimhester: Ok I did that in 9ac19994

R/write.R
jimhester: Thanks, I made this change in 8568602b

R/actions.R
```diff
@@ -0,0 +1,129 @@
...
+  if (!new) {
+    return(invisible(FALSE))
+  }
+  use_github_actions_badge("R")
+
+  invisible(TRUE)
+}
+
+#' @section `use_tidy_actions()`
+#' Sets up tidyverse actions that support the officially supported R versions
```
hadley:
```suggestion
#' Sets up tidyverse actions that check our officially supported R versions
```

R/actions.R
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  if (open && new) {
+    edit_file(proj_path(save_as))
+  }
+
+  invisible(new)
+}
+
+#' @section `use_action_check_quick()`:
```
hadley: I'm not sure quick has quite the right implication here. Maybe `_single()`? `_stable()`?

hadley:
Actually, I'm currently seeing:

```R
[master] > use_tidy_actions()
✔ Creating '.github/workflows/R.yaml/'
Error in file(con, "r") : cannot open the connection
In addition: Warning messages:
1: In file(con, "r") :
  'raw = FALSE' but '/Users/hadley/Documents/r-lib/scales/.github/workflows/R.yaml' is not a regular file
2: In file(con, "r") :
  cannot open file '/Users/hadley/Documents/r-lib/scales/.github/workflows/R.yaml': it is a directory
```

I have latest latest and restarted R.

R/actions.R
jimhester: Maybe `release()`?

jimhester:
Sorry, there was a bug when I just changed how the directory was being created, should be fixed now.

jimhester:
Do you want me to separate the roxygen-devel changes from the meat of this PR?

R/actions.R
hadley: 👍

hadley:
Oh, yes please.

jimhester:
Ok diff should be much more manageable now.

R/actions.R
```diff
@@ -0,0 +1,131 @@
...
+#' @rdname actions
+#' @export
+use_action_check_full <- function(save_as = "R.yaml", ignore = TRUE, open = FALSE) {
+  use_action("check-full.yaml", save_as = save_as, ignore = ignore, open = open)
+
+  use_github_actions_badge("R-CMD-check")
+}
+
+#' @section `use_action_pr_commands()`:
+#' This workflow enables the use of 2 R specific commands in pull request
```
jennybc:
```suggestion
#' This workflow enables the use of 2 R-specific commands in pull request
```

NAMESPACE
```diff
@@ -60,6 +60,9 @@ export(ui_todo)
 export(ui_value)
 export(ui_warn)
 export(ui_yeah)
+export(use_action_check_full)
+export(use_action_check_release)
+export(use_action_pr_commands)
```
jennybc: Maybe it's too long, but should these be `use_github_action_*()` for consistency and auto-complete predictability?

NAMESPACE
```diff
@@ -141,6 +146,7 @@ export(use_template)
 export(use_test)
 export(use_testthat)
 export(use_tibble)
+export(use_tidy_actions)
```
jennybc: Ditto re: not having same prefix as the rest.

NAMESPACE
jimhester: I don't think so, that is way too long IMHO, I think `use_action` is clear

NAMESPACE
jimhester: This is meant to be consistent with `use_tidy_ci()`.

R/actions.R
```diff
@@ -19,15 +19,15 @@ use_github_actions <- function() {
...
 
   if (!new) {
-    return(invisible(FALSE))
+    return(invisible(new))
   }
-  use_github_actions_badge("R")
+  use_github_actions_badge("R-CMD-check")
 
-  invisible(TRUE)
+  invisible(new)
```
jennybc: If this is in response to my comment, it wasn't about this instance. Rather, my comment was about lines 41-45.

jimhester: It is the same point though, they are both changed now

R/actions.R
```diff
@@ -39,10 +39,10 @@ use_tidy_actions <- function() {
   new <- use_action_check_full() && use_action_pr_commands()
 
   if (!new) {
-    return(invisible(FALSE))
+    return(invisible(new))
   }
 
-  invisible(TRUE)
+  invisible(new)
 }
```
jennybc: Sorry 😂 I meant: can't this be replaced with a simple `invisible(new)`? It covers the `TRUE` and `FALSE` case and we do it elsewhere when there's no additional work done for one case.

jimhester: Oh right, I did this now

NAMESPACE
jennybc: Do you think the GitHub action stuff should *be* `use_tidy_ci()` now?

jennybc:
Shall I merge this?

I note there are no tests.

I suppose we can add them after the fact.

hadley:
I think we should merge.

jimhester:
Oh, I just actually decided to take Jenny's suggestion and make all the prefixes consistent. And I was working on adding some tests. I will send a new PR once it is done.

jennybc:
Sounds good.

NAMESPACE
orchid00: hi, where do we find `use_action_bookdown()` or `use_action_blogdown()`? @murraycadcow @jimhester ?

jimhester: They don't exist currently

PR #936, opened 2019-11-12
Rename GitHub Actions functions to have a common prefix

jimhester:
Also add tests for all of the functions.

I renamed the filenames so that the test file matches the source file as well.


R/github-actions.R
```diff
@@ -4,51 +4,44 @@
...
   check_uses_github()
 
-  new <- use_action_check_release()
-
-  if (!new) {
-    return(invisible(new))
-  }
-  use_github_actions_badge("R-CMD-check")
-
-  invisible(new)
```
jimhester: Test coverage revealed the bottom lines would never get run, as `use_action_check_release()` already adds the badge, so we don't need to do it twice.

R/github-actions.R
```diff
@@ -4,51 +4,44 @@
...
 #'   `styler::style_pkg()` and update the PR.
-#' @rdname actions
+#' @rdname github_actions
 #' @export
-use_tidy_actions <- function() {
+use_github_actions_tidy <- function() {
   check_uses_github()
 
-  new <- use_action_check_full() && use_action_pr_commands()
+  new <- use_github_action_check_full() && use_github_action_pr_commands()
```
jimhester: One issue with this is that if a package does _not_ have a README, then `use_github_action_check_full()` will return `FALSE`, so the pr commands won't be added.

I guess I decided this is rare enough not to worry about.

jennybc:
Thanks!

PR #937, opened 2019-11-14
Do not append a newline to the lines

jimhester:
rlang::inform() already appends a newline

### DIFF
--- a/R/ui.R
+++ b/R/ui.R
@@ -271,7 +271,7 @@ cat_line <- function(..., quiet = getOption("usethis.quiet", default = FALSE)) {
     return(invisible())
   }
 
-  lines <- paste0(..., "\n")
+  lines <- paste0(...)
   rlang::inform(lines)
 
   invisible()


jennybc:
Thanks!

krlmlr:
@lionel-: Does `inform()` still append newlines? I see this with most recent rlang and usethis:

``` r
usethis::create_project(tempdir())
#> ✔ Setting active project to '/tmp/RtmppOB7bf'✔ Creating 'R/'✔ Writing a sentinel
#> file '.here'● Build robust paths within your project via `here::here()`●
#> Learn more at <https://here.r-lib.org>✔ Setting active project to '<no active
#> project>'
```

<sup>Created on 2019-12-28 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Note how ✔ and ● are missing a newline before.

This only happens with knitr, presumably because something else appends a newline if it is missing in interactive sessions.

This behavior is introduced with this pull request. I haven't tested older rlang versions. Do we need to revert?

lionel-:
This is strange, looking at the code I don't see anything that could cause missing newlines in non-interactive sessions. When I run `R -e "rlang::inform('foo'); rlang::inform('bar')"`, the whitespace looks correct as well.

jimhester:
You can reproduce this outside of usethis with the following (I was using the current CRAN release of rlang)

``` r
f <- function() {
  rlang::inform("hi")
  rlang::inform("ho")
}
f()
#> hiho
#> NULL
```

<sup>Created on 2019-12-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

Since reprex uses knitr you get the concatenation behavior described.

But if you run that same code in the interactive console or via the command line the newlines are added appropriately. The rlang behavior seems equivalent to `message(appendLF = FALSE)`, but _only_ when run under knitr...

lionel-:
Ah I know what it is, newline is appended when cat'ing output but not in the condition message. knitr uses `evaluate::evaluate()` which captures conditions, hence the missing newline.

jimhester:
Yeah, that makes sense, thanks for opening the issue!

jennybc:
So I should not do anything, right? usethis just provided the demo of the problem, but it lives and must be addressed in rlang.

lionel-:
Right I will fix it in rlang.

lionel-:
Now fixed, I plan to release rlang early next week.

PR #945, opened 2019-11-22
Link GitHub Actions badge to repo URL

### LINKED ISSUE #944: Do not link to the actions tab from the actions badge.
The actions tab is only available (even for public repos) if you are logged into GitHub. So having the badge link to it generates a 404 when running R CMD check.

We need to change the badge link to something else, I guess just the repo URL...

DavisVaughan:
Closes #944 

Linking directly to the GitHub Actions page on the repo will 404 if you are not signed into GitHub, which makes CRAN check unhappy.

i.e. try going here in an incognito window:
https://github.com/r-lib/vctrs/actions?workflow=R-CMD-check

The suggestion in #944 is just to link to the repo URL, so that is what I've done here.

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -46,10 +46,9 @@ use_github_actions_badge <- function(name = "R-CMD-check") {
   check_uses_github()
 
   name <- utils::URLencode(name)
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks!

Issue #954, opened 2019-12-07
Change proj_activate so it accepts relative paths


lettucehead:
Hello,

Instead of going through the process of creating a reproducible example, can you just look at this diff?

```
diff --git a/R/proj.R b/R/proj.R
index 5ac1e5e..974b301 100644
--- a/R/proj.R
+++ b/R/proj.R
@@ -274,11 +274,11 @@ proj_activate <- function(path) {
     rstudioapi::openProject(path, newSession = TRUE)
     invisible(FALSE)
   } else {
+    proj_set(path)
     if (user_path_prep(getwd()) != path) {
       ui_done("Changing working directory to {ui_path(path, base = NA)}")
       setwd(path)
     }
-    proj_set(path)
     invisible(TRUE)
   }
 }
diff --git a/tests/testthat/test-proj.R b/tests/testthat/test-proj.R
index 5294411..22bfe4b 100644
--- a/tests/testthat/test-proj.R
+++ b/tests/testthat/test-proj.R
@@ -1,5 +1,19 @@
 context("projects")
 
+test_that("proj_activate() on relative dir", {
+  tmpdir <- file_temp()
+  tmpleaf <- basename(tmpdir)
+  tmpparent <- dirname(tmpdir)
+  on.exit(dir_delete(tmpdir))
+  dir_create(tmpdir)
+  setwd(tmpparent)
+  proj_activate(tmpleaf) # This is the important part.
+  expect_equal(
+    getwd(),
+    tmpdir
+  )
+})
+
 test_that("proj_set() errors on non-existent path", {
   expect_usethis_error(
     proj_set("abcedefgihklmnopqrstuv"),
```

Unfortunately the tests cannot be run for some reason, so if there are any regressions that is possibly going to need someone to look at the tests themselves.

Here is the output of the tests:

```
        > devtools::test()
        Loading usethis

        Attaching package: ‘testthat’

        The following object is masked from ‘package:devtools’:

                        test_file
        ✔ Setting active project to '/usr/home/r/usethis'
        Error in importIntoEnv(pkgenv, exports, nsenv, exports) :
                cannot change value of locked binding for 'proj_get'
        Calls: test ... load_all -> <Anonymous> -> export_ns -> importIntoEnv
        > traceback()
        5: importIntoEnv(pkgenv, exports, nsenv, exports)
        4: export_ns(package)
        3: pkgload::load_all(path = path, reset = reset, recompile = recompile,
[...]

PR #955, opened 2019-12-07
Change proj_activate so it accepts relative paths. Fixes issue #954

lettucehead:


jennybc:
What is the rationale for this? That would be good to add to the thread of the related issue.

Also, the test fails. It appears you are attempting to activate as project in a directory that doesn't fulfill any of the criteria for a project.

FYI the "fixes #n" syntax has to occur in a commit message or in the PR description (*not* the PR title) in order to have the desired effect.

https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords

PR #958, opened 2019-12-13
export use_github_action function

smwindecker:
Discussed with @jimhester this morning at ozunconf that this functionality to call actions from marketplace url's should be exported


jimhester:
LGTM, thanks @smwindecker!

jennybc:
@smwindecker I think you need to run `devtools::document()` to that the new export causes the requisite change in `NAMESPACE`.

I'll eventually do this myself and merge, but if you do it first, even better.

jimhester:
@jennybc the namespace already has the change in this PR unless I am missing something https://github.com/r-lib/usethis/pull/958/files#diff-7347fe5a0f184f79ef064e92e3beb297R106

jennybc:
> @jennybc the namespace already has the change in this PR unless I am missing something

Oh, that's odd, isn't it 🤔? I mean ... how did that happen? I was just looking at this PR in isolation.

jennybc:
Oh, sorry. Nevermind all.

smwindecker:
Thanks @jennybc! Yes not sure why it looked like a check had failed. I felt pretty confident leaving it though, as I knew the documentation was updated.

jennybc:
The travis failure is some other, unrelated thing. This looks like an intermittent failure, plus there are at least 2 other travis gremlins lurking this week. No worries.

PR #968, opened 2020-01-02
`pr_finish()`: Check if remote tracking branch still exists

### LINKED ISSUE #950: `pr_finish()`: Delete local branch if remote was deleted
When the remote has been deleted and I run `git pull` before `pr_finish()`, I get the following

```r
Error in git2r::fetch(repo, name = remref_remote(remref), credentials = credentials,  : 
  Error in 'git2r_remote_fetch': 'name' must be a character vector of length one with non NA value
```

1. PR gets merged on GH and the branch is deleted
2. Running `git pull` (the deleted remote branch is fetched)
3. When calling `pr_finish()` now, it errors with the msg above because the remote was removed during the `git pull` call

This happens quite often when working on multiple machines. When finishing the PR on one machine, I do not immediately run `pr_finish()` on the other one. Instead I pull first and then run into the problem described above. 
Happy to do a PR if it would be accepted.

## Proposal

`pr_finish()` should check if the remote still exists and (if not) only delete the local branch.

pat-s:
fixes #950 

Tested it on a case where the remote branch was already deleted but the local branch was still present -> worked.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,8 @@
 
 * `pr_finish()` checks that you don't have any local changes (#805).
 
+* `pr_finish()` does not error if the remote branch was already deleted (#950)
+
 * `use_lifecycle()` now adds `@importFrom lifecycle deprecate_soft` to 
   silent an R CMD check note (#896).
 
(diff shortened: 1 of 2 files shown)


pat-s:
@jennybc Should I close here because its unlikely that this approach will be merged?

hadley:
Yeah, I think better to close this and we'll consider the problems more broadly in the near future.

PR #976, opened 2020-01-09
move test helper to enable access from outside

### LINKED ISSUE #973: FR: export functions to facilitate tests
I am using the `usethis` package as a basis for [another package](https://www.github.com/aaronpeikert/repro) (it is btw a great and remarkably easy to extend `usethis`). However as I started to write tests that involve testing whether files are created etc., I had to copy the test helper from usethis [(tests/testthat/helper.R)](https://github.com/r-lib/usethis/blob/b2e894eb6d1d7f3312a783db3bb03a7cc309ba61/tests/testthat/helper.R) into my package. It would be great if you could export them as functions as they are really useful for extending `usethis`.

aaronpeikert:
Moving the test helpers `scoped_temporary_thing()` and friends to `R/` enables reuse of them by packages which build upon `usethis` to facilitate tests of these packages.
Because of their rather internal nature, they are *not* exported, which does not pose a problem for reuse by other since tests are allowed to use unexported functions.
closes #973

### DIFF
--- a/R/scope_temporary.R
+++ b/R/scope_temporary.R
@@ -1,0 +1,62 @@
+# These functions are helpful for testing usethis functions
+# and are therefore not intended for the enduser
+# but may help developers of packages based on usehis
+
+## putting `pattern` in the package or project name is part of our strategy for
+## suspending the nested project check during testing
+pattern <- "aaa"
+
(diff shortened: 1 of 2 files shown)


hadley:
You shouldn't be using these, whether or not they're exported — we reserve the right to change these at any time, without warning.

PR #977, opened 2020-01-09
allow create_package(.) to work; improve valid_package regex

MichaelChirico:
Please let me know if you'd like me to file an issue with this.

My workflow was:

 1. Navigate to my github directory `cd ~/github`
 2. Create my pkg dir: `mkdir pkg`
 3. RStudio : File > New Project... > Existing Directory > [select `pkg`]
 4. In new RStudio session within `pkg.Rproj`, try to run `usethis::create_package('.')`, but it failed.

This issue would allow this workflow which doesn't seem to crazy to me.

I kind of slapped `basename(normalizePath(.))` in there, not sure if you'd prefer that to be part of `user_path_prep` or `path_file`; please advise.

### DIFF
--- a/R/create.R
+++ b/R/create.R
@@ -36,6 +36,8 @@ create_package <- function(path,
   check_path_is_directory(path_dir(path))
 
   name <- path_file(path)
+  # e.g. create_package('.') will expand . to the current dirname
+  name = basename(normalizePath(name))
   if (check_name) {
     check_package_name(name)
   }
--- a/R/create.R
+++ b/R/create.R
@@ -37,7 +37,7 @@ create_package <- function(path,
 
   name <- path_file(path)
   # e.g. create_package('.') will expand . to the current dirname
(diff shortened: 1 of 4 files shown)


jennybc:
Already a PR about this in https://github.com/r-lib/usethis/pull/916

R/create.R
```diff
@@ -36,6 +36,8 @@ create_package <- function(path,
   check_path_is_directory(path_dir(path))
 
   name <- path_file(path)
+  # e.g. create_package('.') will expand . to the current dirname
+  name <- basename(normalizePath(name))
```
jennybc: I am almost certain we'd never proceed like this because we've gone to great lengths to use fs everywhere internally.

MichaelChirico:
thanks @jennybc, I searched the issue tracker & not open PRs (now I'm seeing the benefit of always filing an issue! 😁).

I'll close this and copy-paste my use case in support 👍

PR #986, opened 2020-01-24
Fix use_logo url converting https://* to https:/*

mitchelloharawild:
```
> usethis::use_logo("hex/icon.svg")
✔ Copied 'hex/icon.svg' to 'man/figures/logo.svg'
● Add logo to your README with the following html:
  # icon <a href='https:/pkg.mitchelloharawild.com/icon'><img src='man/figures/logo.svg' align="right" height="139" /></a>
  [Copied to clipboard]
```

This PR fixes the URL `https:/*` to be `https://*`. I don't think treating the URL as a path here is necessary, as it is from pkgdown config.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,8 @@
   `use_github_action_pr_commands()`, `use_github_actions_tidy()` to set up a GitHub Actions
   for a package (@jimhester).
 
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks!

PR #987, opened 2020-01-26
Renv compatibility

adelmofilho:
It implements an optional functionality on `use_package` function allowing it to import the package version contained on `renv.lock` file created by `renv` package. This ensures package development associated with package version management.


hadley:
Can you please file an issue so we can understand the problem better before we commit to a solution?

Issue #990, opened 2020-01-29
pr_push() with a repo with no-pr yields "Error in prs[[1]]: subscript out of bounds"


maurolepore:
I expected `pr_push()` to be able to push a new pull request to a repo with no previous pr (e.g. https://github.com/maurolepore/with-no-pr). But instead I get "Error in prs[[1]]: subscript out of bounds".

The problem seems to be in `pr_find()`. It correctly returns `character()` with a new pr to a repo with some pr (e.g. https://github.com/maurolepore/with-some-pr). But errors if the new pr is to a repo with no previous pr.

``` r
# Ok
usethis:::pr_find("maurolepore", repo = "with-some-pr", pr_branch = "new")
#> character(0)

# Fails
usethis:::pr_find("maurolepore", repo = "with-no-pr", pr_branch = "new")
#> Error in prs[[1]]: subscript out of bounds
```

<sup>Created on 2020-01-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

cc' @uni-sabia

PR #991, opened 2020-01-29
pr_push() can now push to repo with no pr (#990)

maurolepore:
See #990 

4411390 tests an internal function and uses a test repo in my personal account. Instead you may want to test an exported function and use a repo that you control.


jennybc:
Thanks! I made some updates.

PR #994, opened 2020-01-31
add tests for `use_data_table()`

### LINKED ISSUE #984: Add a test for use_data_table()
If we'd had a test for this, this PR (#983) would not have been necessary.

Related functions (that **do** have tests) to look at for cues how to do this:

  * `use_tidy_eval()`
  * `use_pipe()`
  * `use_tibble()`

malcolmbarrett:
Closes #984 

(I am using a mock roxygen check here, but `use_data_table()` doesn't currently check for roxygen despite using `@import`. Will add a separate issue for that.)

### DIFF
--- a/tests/testthat/test-use-data-table.R
+++ b/tests/testthat/test-use-data-table.R
@@ -1,0 +1,20 @@
+context("use_data_table")
+
+test_that("use_data_table() requires a package", {
+  scoped_temporary_project()
+  expect_usethis_error(use_data_table(), "not an R package")
+})
+
+test_that("use_data_table() Imports data.table", {
+  with_mock(
+    `usethis:::uses_roxygen` = function(base_path) TRUE, {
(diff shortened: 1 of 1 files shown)


tests/testthat/test-use-data-table.R
```diff
@@ -0,0 +1,20 @@
+context("use_data_table")
```
jennybc: `context()` is no longer necessary and is mildly discouraged. So this should be deleted.

jennybc: I realize we still have some pre-existing `context()` calls, but we shouldn't be adding any new ones.

malcolmbarrett:
Removed `context()` (that's what I get for following existing patterns 😉) and added mock install of data.table (very clever!)

jennybc:
Thanks!

PR #997, opened 2020-01-31
Check for roxygen in `use_data_table()`

malcolmbarrett:
`use_data_table()` doesn't currently check for roxygen, but as this implementation uses `@import`, it seems essential, as in the case of `use_tibble()`. This PR adds a check.


jennybc:
LGTM

PR #1009, opened 2020-02-04
Use better link for the GitHub Actions badge

jimhester:
GitHub now provides a /actions URL which is accessible even when users
are not logged in, so we can now link to that


jennybc:
Thanks!

Issue #1005, opened 2020-02-03
Installation of system dependencies failing after running use_github_actions_tidy


Robinlovelace:
I read with great interest about GitHub actions in these (awesome comic themed!) slides: https://speakerdeck.com/jimhester/github-actions-for-r?slide=40

Following the instructions I tried:

```r
use_github_actions_tidy()
```

For the package `stplanr`.

Good news: the outpute seems to have got the actions going.

Issue, the system requirements seem to fail on all platforms except for Windows: https://github.com/ropensci/stplanr/runs/420731417 (screenshot below)

I had a look at `R-CMD-check.yml` and guess that there is an underlying issue with the ~~[`pkgdepends`](https://github.com/r-lib/pkgdepends)~~ [`sysreqs`](https://github.com/r-hub/sysreqs) package. Happy to open an issue there, but seeing as I first found this issue while using this package, and imagining that others may have similar issues, opening here as a starting point.

On a related note, wondering if there's a way to use RStudio's `r-system-requirements` project to support automated package dependencies for things like GDAL, on which `stplanr` depends, which is mentioned in the README here: https://github.com/rstudio/r-system-requirements

Thanks.

![image](https://user-images.githubusercontent.com/1825120/73665160-8f37e600-4698-11ea-8d53-11c1ac9e58e9.png)

Robinlovelace:
Update, have just tried re-running the command after installing dev version and the resulting yaml files seem unchanged.

jennybc:
cc @jimhester 

I'm going to close this because all usethis does is write files. I think this is an issue for https://github.com/r-lib/actions or https://github.com/r-hub/sysreqs.

FWIW I setup gh actions for googlesheets4 yesterday. I believe the sysreqs only works for Linux? For my macOS builds I had to add `brew install this-and-that` commands to install system requirements. For all but the simplest package, it is quite likely that the base workflow YAML is going to need customization.

Robinlovelace:
Great, thanks for the pointers. Hope this issue, and your response, is useful to others.

Robinlovelace:
Heads-up @jennybc and @jimhester, good news: my colleague @layik fixed the issue with this commit: https://github.com/ropensci/stplanr/commit/ad6b95227a8f518119b3394f8529f9d1120682fb

I'm unsure why the dependencies do not install as per https://github.com/r-hub/sysreqs/issues/14 but happy that it's fixed and hope that others find this issue and the contents of the fix useful.

PR #1010, opened 2020-02-05
Add note about potential need to edit .yml files

Robinlovelace:
Follow-up from #1005


hadley:
I think the need to customise the yaml is somewhat implicit, but since these functions are still heavily in flux I'd prefer not to commit to specific wording at this time.

PR #1014, opened 2020-02-13
Add generated shared libraries to .gitignore for macOS

davidchall:
I saw that `usethis::use_rcpp()` adds shared library files matching `*.so` to the `src/.gitignore`. This is the file extension used on Linux. On macOS, these files are given a `*.dylib` extension.

For consistency with Linux, I thought these probably shouldn't enter source control. However, I'm very new to Rcpp, so would love to learn if I've misunderstood!

### DIFF
--- a/R/rcpp.R
+++ b/R/rcpp.R
@@ -75,7 +75,7 @@ use_c <- function(name = NULL) {
 
 use_src <- function() {
   use_directory("src")
-  use_git_ignore(c("*.o", "*.so", "*.dll"), "src")
+  use_git_ignore(c("*.o", "*.so", "*.dll", "*.dylib"), "src")
   roxygen_ns_append(glue("@useDynLib {project_name()}, .registration = TRUE")) &&
     roxygen_update()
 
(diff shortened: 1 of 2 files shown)


hadley:
This has never caused problems for me before so I'd like to understand the issue more fully before considering a solution. Could you please file an issue with reprex?

davidchall:
When I reported this issue, I was using a [conda](https://conda.io) R environment for package development. Since then, I've moved to a more standard environment (using the official binaries) and noticed that the `.dylib` file is no longer generated.

I think the underlying difference here is that [the official binaries prefer static linking on macOS](https://cran.r-project.org/doc/manuals/r-patched/R-admin.html#External-software) whereas [conda prefers dynamic linking](https://docs.conda.io/projects/conda-build/en/latest/resources/use-shared-libraries.html#shared-libraries-in-macos-and-linux). Of course, these shared libraries only appear if using dynamic linking, which is what I saw with the conda environment.

Whether you merge this PR probably comes down to whether you want to support a conda development environment. (I'm not sure a reprex would help with this ticket, but please let me know if you still want one.)

PR #1026, opened 2020-03-04
Ignore the whole .github folder

jimhester:
Rather than trying to ignore individual files. This way is simpler and
less fiddly if you later add other workflows.

### DIFF
--- a/R/github-action.R
+++ b/R/github-action.R
@@ -39,14 +39,13 @@ use_github_action <- function(name,
 
   contents <- readLines(url)
 
+
+  use_dot_github(ignore = ignore)
+
   save_as <- path(".github", "workflows", save_as)
-
   create_directory(dirname(proj_path(save_as)))
-  new <- write_over(proj_path(save_as), contents)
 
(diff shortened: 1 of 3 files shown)


jimhester:
Pretty sure this correct, but it would be good for a quick sanity check before merging.

jennybc:
I didn't run anything but, assuming you have, LGTM.

jennybc:
Is this really the "root" / main / only place to do this, though? I think there are other functions writing to `.github/` and it would be nice to rationalize all of it at once.

jimhester:
Good question, it looks like there is a `use_dot_github()` helper for this, which we should be using here.

jimhester:
Ok I now use the internal `use_dot_github()` function as of https://github.com/r-lib/usethis/pull/1026/commits/b9bdc497a7fb0f6ba0dc1063d2e1f15b7681fedb.

I had to expose the `ignore` argument for it, but it should preserve existing behavior.

jennybc:
Thanks!

PR #1027, opened 2020-03-05
Added badges for downloads and real fix for appveyor

muschellij2:
This commit does 2 things (probably should break into 2 commits), but it replaces https://github.com/r-lib/usethis/pull/875.  

BUG: Appveyor has an issue with periods, which should be replaced with `-`.  For example, `oro.nifti` (https://github.com/muschellij2/oro.nifti) maps to https://ci.appveyor.com/project/muschellij2/oro-nifti.  

ENH: Also, added badges which are from https://cranlogs.r-pkg.org/


hadley:
Can you please make two PRs for the two separate issues?

PR #1030, opened 2020-03-06
add use_data_pronoun

### LINKED ISSUE #1016: `use_*()` function for setting up rlang data pronoun
I'm not sure if this workflow is settled enough to encode it in usethis, but the [Using ggplot2 in packages](https://ggplot2.tidyverse.org/dev/articles/ggplot2-in-packages.html) vignette suggests  using the `.data` pronoun from rlang for simple cases where you know the NSE variables you're going to reference. The vignette says:

> To avoid a similar note from the CMD check about `.data`, use `#' @importFrom rlang .data` in any roxygen code block (typically this should be in the package documentation as generated by `usethis::use_package_doc()`).

I think something like `use_data_pronoun()` that operates similarly to `use_tibble()` and `use_pipe()` (e.g. adding rlang as import, checking for a package doc, providing the roxygen code) would be helpful. Happy to make a PR if agreeable!

malcolmbarrett:
Closes #1016 

This PR:

* Adds `use_data_pronoun()`, which adds rlang's `.data` pronoun to a package import following the guidelines in the [ggplot2 vignette](https://ggplot2.tidyverse.org/articles/ggplot2-in-packages.html). This lets users call `.data` easily in package code. This function is similar to `use_tidy_eval()` but oriented to much simpler use cases. This function operates similarly to `use_pipe()`.
* Adds corresponding tests.

A few notes:
* An alternative name for this function might be `use_tidy_eval_pronoun()` or something to that effect, but I thought it would be best to keep it out of the tidyverse functions.
* I did use a min version of 0.1.2 for rlang to be consistent with `use_tidy_eval()`, but I am not sure that's necessary. 
* I did not add code to export `.data`, which both `use_pipe()` and `use_tidy_eval()` do for their use cases, because it seemed less necessary to export `.data` alone. I can add this + a template, though, if desired.


R/data-pronoun.R
```diff
@@ -0,0 +1,36 @@
...
+use_data_pronoun <- function() {
+  check_is_package("use_data_pronoun()")
+  check_uses_roxygen("use_data_pronoun()")
+
+  use_dependency("rlang", "Imports", min_version = "0.1.2")
+
+  if (has_package_doc()) {
+    roxygen_ns_append("@importFrom rlang .data") && roxygen_update()
+    return(invisible(TRUE))
+  }
```
hadley: Can you please put the remaining code in an `else` branch? I think that makes it more clear there's two options here — either adding automatically or telling the user how to add manually.

PR #1031, opened 2020-03-06
`use_github_release()` now tags the latest local commit

### LINKED ISSUE #1029: use_github_release() can tag the wrong commit
The `use_github_release()` function creates a draft GitHub Release using the [corresponding GitHub API](https://developer.github.com/v3/repos/releases/#create-a-release). However, it doesn't set the `target_commitish` field and so GitHub uses the default branch (usually `master`).

There are situations where your local HEAD doesn't correspond to the remote default branch. For example, when releasing patches from a non-master branch. In this case, `use_github_release()` checks that local commits have been pushed to the remote, but then tags the wrong commit. I think this function should probably be tagging the most recent local commit.

davidchall:
Fixes #1029 

Verified this works correctly [here](https://github.com/davidchall/usethis/releases/tag/untagged-cefbae9cb62c81456b26).


hadley:
Good catch — thanks muchly!

PR #1033, opened 2020-03-06
add option `usethis.destdir`

### LINKED ISSUE #1015: Set option for `destdir`
In `create_from_github()`, `use_course()`, and `use_zip()`, usethis tries to guess a good download location if `destdir` is `NULL`. I think it would be nice to also set an option in your `.Rprofile` to specify a default location, e.g. `usethis.destdir`. I believe this would only entail a simple change to `conspicuous_place()`, which would check if such an option is set and return that location if so, otherwise going on with its usual business. Happy to make a PR if agreeable!

malcolmbarrett:
Closes #1015.

This PR adds an option `usethis.destdir` to set a default directory for use in `create_from_github()` and `use_course()`, describes it in the relevant docs and setup vignette, and adds a test.

Note that I also added a brief description about `usethis.quiet` to the setup vignette, since it was missing. If this PR is not accepted, I'd be happy to submit a smaller one that has that, since it should probably be there.

I'll admit that, while I think this would be a very useful feature, there's a little bit of an inconsistency. `use_zip()` and `use_pkgdown()` both also have `destdir` options that have defaults. `use_pkgdown()` is perhaps less of an issue but `use_zip()` is closely related to `use_course()` and even shares a help page with it.

### DIFF
--- a/R/course.R
+++ b/R/course.R
@@ -15,7 +15,10 @@
 #'   Function works well with DropBox folders and GitHub repos, but should work
 #'   for ZIP files generally. See examples and [use_course_details] for more.
 #' @param destdir The new folder is stored here. If `NULL`, defaults to user's
-#'   Desktop or some other conspicuous place.
+#'   Desktop or some other conspicuous place. You can also set a default
+#'   location using the option `usethis.destdir`, e.g. `options(usethis.destdir
+#'   = "a/good/dir")`, perhaps saved to your `.Rprofile` with
+#'   [`edit_r_profile()`]
 #' @param cleanup Whether to delete the original ZIP file after unpacking its
 #'   contents. In an interactive setting, `NA` leads to a menu where user can
 #'   approve the deletion (or decline).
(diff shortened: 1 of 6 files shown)


hadley:
This just feels a bit too specialised sorry; every configuration option is just one more thing to debug when trying to help someone else.

hadley:
Ok @jennybc wants this too so I take back my claim about it being too specialised 😄

malcolmbarrett:
Ha, I definitely agree that it's right on the line of useful and overly specialized! Perhaps I could add some output to one or both of the `*_sitrep()` functions to help make more explicit if that option is set for when you are helping someone?

R/course.R
```diff
@@ -50,14 +53,15 @@ NULL
 #' @export
 use_course <- function(url, destdir = NULL) {
   url <- normalize_url(url)
-  destdir_not_specified <- is.null(destdir)
+  destdir_not_specified <- is.null(destdir) && is.null(getOption("usethis.destdir"))
```
hadley: I think you can simplify the logic by changing the `destdir` default to `getOption("usethis.destdir")`

R/course.R
```diff
@@ -50,14 +53,15 @@ NULL
...
   destdir <- user_path_prep(destdir %||% conspicuous_place())
   check_path_is_directory(destdir)
 
   if (destdir_not_specified && interactive()) {
     ui_line(c(
       "Downloading into {ui_path(destdir)}.",
-      "Prefer a different location? Cancel, try again, and specify {ui_code('destdir')}"
+      "Prefer a different location? Cancel, try again, and specify {ui_code('destdir')}//
```
hadley: I think it's sufficient to advertise this feature in the documentation.

R/course.R
```diff
@@ -15,7 +15,10 @@
 #'   Function works well with DropBox folders and GitHub repos, but should work
 #'   for ZIP files generally. See examples and [use_course_details] for more.
 #' @param destdir The new folder is stored here. If `NULL`, defaults to user's
-#'   Desktop or some other conspicuous place.
+#'   Desktop or some other conspicuous place. You can also set a default
+#'   location using the option `usethis.destdir`, e.g. `options(usethis.destdir
```
hadley: Can you please avoid wrapping a code fragment across multiple lines?

malcolmbarrett:
Updated to reflect those changes, thanks!

hadley:
Thanks @malcolmbarrett!

PR #1036, opened 2020-03-09
Remove azure pipelines code

jimhester:
GitHub Actions replicates most of the functionality of Azure Pipelines
and is much easier to setup.

I don't think there is a ton of value in keeping the code in usethis, and it would probably just generate confusion to leave it in, so I think it is best to remove it.


jimhester:
The functions were never in a CRAN release, so I don't think we need to do a deprecation cycle.

jennybc:
Seems like a good idea.

PR #1042, opened 2020-03-11
add navigation to github secrets

maxheld83:
Might be nice to easily navigate to repo settings, especially secrets, when pasting PATs from codecov or `usethis::browse_github_pat()`.


hadley:
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Actually, given that you no longer need to routinely paste PATs I'm not sure we need this.

PR #1049, opened 2020-03-14
Separate use_git() into two steps

### LINKED ISSUE #852: use_git() should make an initial commit, if necessary
TLDR:Use_git doesn't make the initial commit if the project is started already with git from Rstudio

I started a new package following the Rstudio new project > R package >  git then I used some of the `usethis` functions to set up the package.

The history of command run is: 
```r
use_rstudio()
use_git()
use_mit_license()
use_testthat()
use_package("package")
desc::desc_normalize()
use_readme_rmd()
use_travis() # Error shown below
use_github()
use_coverage()
use_vignette("name")
?use_github
git_sitrep() # Sitrep shown below
```
However it failed:
different commands that relay on github remote
Despite having correctly setup the git configuration, see the `git_sitrep`:
Result of git_sitrep

The problem comes from `use_github`, which expects to find a branch. But the branch is not created in `use_git` because Rstudio already created the repository and then `use_git` doesn't continue and commit the "Initial commit". 

A solution would be to check if there is any committed file by checking the repository head. This involves the line https://github.com/r-lib/usethis/blob/9ffc27fd69934c74377fd3598489cebc50ef2680/R/git.R#L14.

Which could be adapted to check if there is any committed file or any branch.

hadley:
1. Initialise repo
2. Perform initial commit

Fixes #852


hadley:
@jennybc this fails because there's no git user name set on GitHub actions. Is this deliberate or an accident?

PR #1051, opened 2020-03-15
Minor overhaul of output

### LINKED ISSUE #981: Revisit the name cat_line()
Now that messaging routes through `rlang::inform()`, the function name `cat_line()` is somewhat misleading. This is an internal matter, but it's something anyone developing in usethis might encounter.

hadley:
* `cat_line()` -> `ui_inform()`, and `cat_bullet()` -> `ui_bullet()` (Fixes #981)
* `ui_line()` prints single empty line when called with no arguments 
* New `ui_silence()` for locally silencing 
* A few UI golden tests


hadley:
I _think_ fixing the windows failure will require switching from clisymbols to cli.

hadley:
Yes, and accidentally did that on master instead of in this branch 😬

hadley:
macOS-latest (3.6) failure is a false positive caused by a codecov timeout.

R/git.R
```diff
@@ -532,7 +532,7 @@ git_sitrep <- function() {
   ## TODO: rework when ui_*() functions make it possible to do better
   branch <- if (is.null(branch)) "<unset>" else branch
   tracking_branch <- if (is.null(tracking_branch)) "<unset>" else tracking_branch
-  cat_line("* ", "Local branch -> remote tracking branch: ",
+  ui_inform("* ", "Local branch -> remote tracking branch: ",
```
jennybc: Just an observation, with the caveat that I haven't recently re-read the `ui_*()` functions. When I see these hand-crafted bullets, I wonder if we should be using `ui_bullet()`?

Oh maybe it's not the same type of bullet. Maybe we need something like `ui_item()` for these sorts of bullet lists. We do this a lot.

R/ui.R
```diff
@@ -128,6 +133,19 @@ ui_warn <- function(x, .envir = parent.frame()) {
...
+#' @rdname ui
+#' @param code Code to execute with usually UI output silenced.
+#' @export
+ui_silence <- function(code) {
+  old <- options(usethis.quiet = TRUE)
+  on.exit(options(old))
+
+  code
+}
+
```
jennybc: So this is basically `with_silence()`. Do we also want `local_silence()`? 🤔

R/ui.R
```diff
@@ -187,7 +205,7 @@ ui_yeah <- function(x,
     qs <- sample(qs)
   }
 
-  cat_line(x)
+  rlang::inform(x)
```
jennybc: Can nix the `rlang::`, based on other PR.

R/ui.R
jennybc: Also, FWIW we import withr if you wanted to use `withr::with_options()`

tests/testthat/test-ui.R
```diff
@@ -1,3 +1,27 @@
...
+    ui_oops("oops")
+    ui_info("info")
+    ui_code_block(c("x <- 1", "y <- 2"))
+
+    ui_stop("an error")
+    ui_warn("a warning")
+  })
+})
+
+test_that("ui_silence() suppressed output", {
```
jennybc:
```suggestion
test_that("ui_silence() suppresses output", {
```

tests/testthat/test-ui.R
```diff
@@ -1,3 +1,27 @@
...
+  # suppress test silencing
+  withr::local_options(list(usethis.quiet = FALSE))
+
+  expect_output(ui_silence(ui_line()), NA)
```
jennybc: So many negations 🤪

R/git.R
hadley: Is that just `ui_info()`?

R/ui.R
hadley: Would you prefer `with_usethis_silence()` and `local_usethis_silence()`?

R/git.R
hadley: Yeah, it seems like sitrep could do with a UI pass a some point.

jennybc: I think we also just gained hand-crafted bullets in the PR about merge conflicts. There are probably many instances.

R/ui.R
jennybc: My comment wasn't about the name as much as having both the "with" and "local" variants, since the need for one suggests you probably also need the other (or will soon).

hadley: Yeah, but what we call it? 

We currently use `withr::local_options(list(usethis.quiet = TRUE))` in two places in the source; and I think `ui_silence()` would be better for both of them. We'd use a local silence more if the tests didn't silence by default.

PR #1053, opened 2020-03-15
Tweak use_lifecycle()

### LINKED ISSUE #1012: use_lifecycle has a confusing instruction
When I run,

```r
use_lifecycle()
```
One of the instructions is:

```
Refer to functions with `lifecycle::fun()`
```

What does this mean? Functions in my package? So instead of `\link{foo}`, I should write `\link{experimental::foo}`??

hadley:
* Don't display default package usage hints
* Don't normalize description

Fixes #1012

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -117,8 +117,7 @@ use_rd_macros <- function(package) {
 
   if (!package %in% macros) {
     macros <- c(macros, package)
-    macros <- paste0("    ", macros, collapse = ",\n")
-    desc::desc_set(RdMacros = macros, file = proj, normalize = TRUE)
+    desc::desc_set(RdMacros = macros, file = proj)
   }
 
   invisible()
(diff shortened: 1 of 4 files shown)


R/helpers.R
```diff
@@ -117,8 +117,7 @@ use_rd_macros <- function(package) {
 
   if (!package %in% macros) {
     macros <- c(macros, package)
-    macros <- paste0("    ", macros, collapse = ",\n")
-    desc::desc_set(RdMacros = macros, file = proj, normalize = TRUE)
+    desc::desc_set(RdMacros = macros, file = proj)
```
lionel-: I see

```
   [ OK: 438 | SKIPPED: 21 | WARNINGS: 1 | FAILED: 1 ]

  1. Failure: use_lifecycle() respects existing RdMacros field 
```

Why this change?

hadley: Ooops, I missed that test — but setting `normalize = TRUE` normalises the entire `DESCRIPTION` which feels a bit aggressive

PR #1056, opened 2020-03-16
Give more information about merge conflicts

hadley:


R/git-utils.R
```diff
@@ -22,12 +22,24 @@ git_pull <- function(remote_branch = git_branch_tracking_FIXME(),
...
+git_conflict_report <- function() {
+  st <- git_status(staged = FALSE, untracked = FALSE)$unstaged
+  conflicted <- unname(st[names(st) == "conflicted"])
+  conflicted_path <- purrr::map_chr(conflicted, ui_path)
+
+  ui_line(c(
+    "There are {length(conflicted)} conflicted files:",
+    paste0("* ", conflicted_path)
+  ))
+  ui_stop("Please fix, stage, and commit to continue")
```
jennybc: This PR is definitely progress. But we could also mention `git merge --abort` as a way to back out, while we're at it.

hadley: I also wondered if we should use `edit_file()` to open these files for editing?

hadley: I'm going to try it, and we can back off it seems too annoying.

PR #1059, opened 2020-03-17
Description refactoring

### LINKED ISSUE #961: Better error message for malformed usethis.description
Might be worthwhile to save to a file and attempt to parse? 

It's also particularly problematic because it causes `create_package()` to fail part way through, leaving you in a malformed state that you just have to blow away.

hadley:
Internally, I've simplified the functions so that `use_description_defaults()` does much more of the work. On top of that I've tweaked `tidy_desc()` so it no longer fails if the `Author@R` is invalid (Fixes #961), and also tweaked the defaults to process `person` objects directly so that you can now put code like this in your `.Rprofile`:

```R
options(
  usethis.description = list(
    `Authors@R` = c(
      utils::person("Hadley", "Wickham", , "hadley@rstudio.com", role = c("aut", "cre")),
      utils::person("RStudio", role = "cph")
    )
  )
)
```

I think this should substantial decrease the failure rate since you're now writing code, and will get autocomplete tips etc.

Since this is the most commonly set option, I do wonder if we should pull it out into its own option:

```R
options(
  usethis.authors = c(
    utils::person("Hadley", "Wickham", , "hadley@rstudio.com", role = c("aut", "cre")),
    utils::person("RStudio", role = "cph")
  )
)
```

Let me know if you think that's worthwhile and I'll add it to this PR.

(the `utils::` is necessary because the util package is loaded after `.Rprofile` is run).


jennybc:
Jotting things down as I go ... some of these questions might answer themselves as I go and I'll edit this.

What is the migration plan for people who already use older approach in `.Rprofile`? The one where the `Authors@R` entry is a string that is (allegedly) a valid call to `person()`.

Re: this:

> Since this is the most commonly set option, I do wonder if we should pull it out into its own option:

The move towards maintaining default `DESCRIPTION` entries as a list was very intentional, so this would represent us starting to walk in a circle. I don't care deeply but we had reasons for abandoning the individual options.

Relevant NEWS entry:

https://github.com/r-lib/usethis/blame/master/NEWS.md#L482

I could dig up more history / rationale if needed.

NEWS.md
```diff
@@ -1,5 +1,16 @@
...
+* `create_package()` no longer fails partway through if you have a malformed
+  `usethis.description` option (#961).
+  
+* In `usethis.description`, you can now set `Authors@R = person()` directly,
+  without having to wrap in additional layer of quotes. If setting this in
+  your `.Rprofile`, you'll need to use `utils::person` since the utils 
+  package isn't loaded until after your profile is executed.
+
+* `use_description_defaults()` now shows the default fields combined with
+  an options that you have set.
```
```suggestion
  any options that you have set.
```

hadley:
There's no need to migrate; the old approach will continue to work. 

If you have any hesitation about adding another option then we should not consider it futher.

R/create.R
```diff
@@ -46,7 +46,7 @@ create_package <- function(path,
   on.exit(proj_set(old_project), add = TRUE)
 
   use_directory("R")
-  use_description(fields, check_name = check_name)
+  use_description(fields, check_name = FALSE)
```
jennybc: This seems to make the `check_name` argument pointless and undoes https://github.com/r-lib/usethis/issues/528. The rationale for having an escape hatch is that people make packages that will never go to CRAN or use the package structure for a data analysis and might have some external naming convention.

R/description.R
```diff
@@ -51,63 +51,59 @@
...
-      Package = "valid.package.name.goes.here",
-      Version = "0.0.0.9000",
-      Title = "What the Package Does (One Line, Title Case)",
-      Description = "What the package does (one paragraph).",
-      "Authors@R" = 'person("First", "Last", , "first.last@example.com", c("aut", "cre"), comment = c(ORCID = "YOUR-ORCID-ID"))',
-      License = " `use_mit_license()`, `use_gpl3_license()` or friends to pick a license",
-      Encoding = "UTF-8",
-      LazyData = "true",
-      Roxygen = "list(markdown = TRUE)"
-    )
```
jennybc: I had structured it this way so people could see the defaults without reading usethis source, by calling this functions. But I think we've talked about this before and you don't find that compelling.

R/description.R
```diff
@@ -51,63 +51,59 @@
...
-    defaults$usethis,
-    defaults$usethis.description %||% defaults$devtools.desc %||% list()
-  )
-  compact(utils::modifyList(defaults, fields))
+build_description <- function(fields, package) {
+  fields <- use_description_defaults(package, fields)
+
+  desc <- desc::desc(text = glue("{names(fields)}: {fields}"))
+  tidy_desc(desc)
+  desc
```
jennybc: This looks odd. Maybe I don't remember what everything does, but worth a check.

R/create.R
hadley: Note that we're already checking the name on like 40; isn't this check redundant?

R/description.R
hadley: Yeah, it just felt like too many layers too me, and you couldn't easily see what would actually get used, which also seems like a bummer.

R/description.R
hadley: It looks weird, but it modifies the mutable desc object in place.

R/create.R
jennybc: Oh, yeah, sorry I am very off here.

R/description.R
jennybc: OK right I did not recall the R6-ness of `desc`.

PR #1061, opened 2020-03-17
Replace readLines() with read_utf8()

### LINKED ISSUE #969: Functions that edit README.Rmd break Unicode characters
Functions that edit packages' README.Rmd, like `use_lifecycle_badge()` result in garbled Unicode characters.

For example, after running `use_lifecycle_badge()`, a pronunciation guide "/ˌfoʊ.naɪˈif/" in my package description became "/ËŒfoÊŠ.naÉªËˆif/".

hadley:
Fixes #969 - we don't have a reprex, I don't remember how to test this easily, but use_badge() calls block_append() which used readLines() without encoding argument which is definitely wrong.


jimhester:
We could look at the [tests in roxygen2 for this I guess](https://github.com/r-lib/roxygen2/pull/649/files#diff-4df2ab4f7afc96c94eb705214f368683R3), but I think it is probably fine without explicit tests, since the reading happens in so many places.

I guess the only thing to remember is that this doesn't do re-encoding, so all input with non-ASCII characters (templates etc.) needs to be generated in UTF-8.

PR #1064, opened 2020-03-17
Move paired R/ and test/ files

### LINKED ISSUE #784: Function to rename R file
And it's accompanying test file

hadley:
My main question: is this the right function name? A `use_` prefix doesn't seem right, and there aren't any other existing functions to draw inspiration from.

Fixes #784


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # usethis (development version)
 
+* New `rename_file()` makes it easy to rename paired `R/` and `test/` files
```
jimhester:
```suggestion
* New `rename_files()` makes it easy to rename paired `R/` and `test/` files 
```

jimhester:
Should this even be in usethis though? Seems more like a devtools function to me.

jimhester:
Also I agree with the name feeling weird, but I don't have a good alternative.

hadley:
usethis seems natural to me because it's connected to the `use_r()` and `use_test()` functions and it's about the structure of the package on disk. But it's suspicious that it's different to every other existing function.

jennybc:
How about `use_paired_files()`?

R/r.R
```diff
@@ -66,6 +66,82 @@ use_test <- function(name = NULL, open = NULL) {
...
+  if (length(old_test) > 0) {
+    ui_done("Moving {ui_path(old_test)} to {ui_path(new_test)}")
+    file_move(old_test, new_test)
+  }
+
+  # Update test file
+  test_path <- proj_path("tests", "testthat", glue("test-{new}"), ext = "R")
+  if (!file_exists(test_path)) {
+    return(invisible())
+  }
```
jennybc: I find it a little hard to follow the thread here. Could the test for "nothing to do / no files" and the early exit be earlier? I just can't immediately see the different cases of file existence / non-existence.

R/r.R
```diff
@@ -66,6 +66,82 @@ use_test <- function(name = NULL, open = NULL) {
 }
 
 
+#' Automatically rename paired `R/` and `test/` files
+#'
+#' @description
+#' * Moves `R/{old}.R` to `R/{new}.R`
+#' * Moves `tests/testthat/test-{old}.R` to `tests/testthat/test-{new}.R`
+#' * Moves `tests/testthat/test-{old}-*.*` to `tests/testthat/test-{new}-*.*`
+#'   and updates paths in the test file.
```
jennybc: Is there a conceptual description of these files? Does it have something to do with how you use `verify_output()`? I'm not immediately sure what we're talking about.

R/r.R
```diff
@@ -66,6 +66,82 @@ use_test <- function(name = NULL, open = NULL) {
...
+#' * Also removes now unneeded `context()` calls from the test file.
```
jennybc:
```suggestion
#' * Removes `context()` calls from the test file, which are unnecessary (and discouraged) as of testthat v2.1.0.
```

Speaking of which, should this PR do anything about putting a minimum version on testthat?

R/r.R
```diff
@@ -66,6 +66,82 @@ use_test <- function(name = NULL, open = NULL) {
...
+#'
+#' This is a potentially dangerous operation, so you must be using Git in
+#' order to use this function
+#'
+#' @param old,new Old and new file names
```
jennybc: Maybe say something about extension handling? I.e. indicate that "foo" and "foo.R" both work.

R/r.R
hadley: testthat 2.1.0 is now a year old, and the version dependencies of suggested packages aren't checked (IIRC)

R/r.R
hadley: This exits if there's no test file with the given name, as opposed to no tests at all.

hadley: Ironically I now don't understand what this block of code does at all.

R/r.R
hadley: Me neither.

PR #1065, opened 2020-03-18
Ensure custom URLs are allowed in use_github_action()

coatless:
This PR fixes how URLs are setup in `use_github_action()` by moving the appending of the workflow `name` outside of the `url` check. 

**Note:** This is required as the `name` parameter must be given but isn't used with a custom `url`.


jimhester:
Actually the original intent was that you specify the full url, including the name if you are using a custom url. So I think the way to fix this is to put the [custom name logic](https://github.com/r-lib/usethis/pull/1065/files#diff-516730624e7f35900a914ca5353c4d0dR26-R29) inside the `is.null(url)` conditional.

jennybc:
@coatless Do you want to take another pass? A usethis release is planned for April 6 (#1076).

coatless:
@jennybc sure.

@jimhester So, moving the naming logic in gives:

```r
if (is.null(url)) {
  stopifnot(is_string(name))
  if (!grepl("[.]yaml$", name)) {
    name <- paste0(name, ".yaml")
  }
  url <- glue("https://raw.githubusercontent.com/r-lib/actions/master/examples/{name}")	
}
```

The function call for a custom action would then be:

```r
use_github_action(url = "https://raw.githubusercontent.com/r-lib/actions/master/examples/blah.yaml")
```

Is there a reason for not wanting the URL to point to a generic location?

jimhester:
I am not sure I understand your question, but yes the behavior you describe is the intended usage.

coatless:
@jimhester Rephrase:

Given the current function call of:

```r
use_github_action(action)
```

Wouldn't it make more sense to have it:

```r
use_github_action(action, url="generic")
```

Instead of:

```r
use_github_action(url="specific/to/action.yaml")
```

coatless:
@jimhester when you have a moment.

R/github-action.R
```diff
@@ -22,15 +22,19 @@ use_github_action <- function(name,
...
+    # Append a `.yaml` extension if needed
+    if (!grepl("[.]yaml$", name)) {
+      name <- paste0(name, ".yaml")
+    }
 
-  if (is.null(url)) {
     url <- glue("https://raw.githubusercontent.com/r-lib/actions/master/examples/{name}")
+  } else {
+    stopifnot(is_string(url))
+    stopifnot(grepl(".*[.]yaml$", url))
```
jimhester, line 37: I think is too strict, for instance `.yml` is pretty common extension for yaml files, even if `.yaml` is actually the one recommended by the [yaml faq](https://yaml.org/faq.html).

I think we should actually just not check the extension at all.

coatless: Okay, I've removed the extension check completely for a custom URL. The only enforcement is a string of some type.

NEWS.md
```diff
@@ -8,6 +8,9 @@
 
 * We now recommend GitHub Actions instead of Travis-CI or AppVeyor, and strongly
   recommend upgrading your packages.
+  
+* Fix `use_github_action()` URL parameter ensure custom URLs are allowed.
```
jennybc, line 12:
```suggestion
* Fix `use_github_action()` URL parameter to ensure custom URLs are allowed.
```

R/github-action.R
```diff
@@ -22,15 +22,18 @@ use_github_action <- function(name,
...
-  stopifnot(is_string(name))
+  # Check if a custom URL is being used.
+  if (is.null(url)) {
+    stopifnot(is_string(name))
 
-  if (!grepl("[.]yaml$", name)) {
-    name <- paste0(name, ".yaml")
-  }
+    # Append a `.yaml` extension if needed
+    if (!grepl("[.]yaml$", name)) {
```
jennybc, line 30: Just based on reading other comments in this review this makes me wonder if a `name` with pre-existing`.yml` extension will now get an additional `.yaml` extension? Which would be undesirable.

If I'm off-base, just ignore me.

If I'm not, perhaps we should add a test that hits the `.yaml` vs `.yml` thing.

coatless: @jennybc it will and also cause an error because `r-lib/actions` are labelled as `.yaml`. So, let's patch that!

coatless: Should be addressed with a `gsub` for `yml` to `yaml` in df603d0 and associated testing infrastructure in 820b76d.

R/github-action.R
```diff
@@ -22,15 +22,19 @@ use_github_action <- function(name,
...
+  # Check if a custom URL is being used.
+  if (is.null(url)) {
+    stopifnot(is_string(name))
 
-  if (!grepl("[.]yaml$", name)) {
-    name <- paste0(name, ".yaml")
-  }
+    # Append a `.yaml` extension if needed
+    name = gsub("[.]yml$", ".yaml", name)
+    if (!grepl("[.]yaml$", name)) {
```
jennybc, line 31: If you keep it, please switch to `name <-` for assignment, instead of `name =`.

But it feels like it would be simpler to, instead, use regex of `[.]ya?ml$` inside the `if()` (caveat: untested but I assume you see what I mean).

I don't see how we can just substitute `.yaml` for `.yml` if that is not the extension used by the workflow file. We can just allow both `.yaml` and `.yml` through, yeah?

tests/testthat/test-github-action.R
```diff
@@ -0,0 +1,43 @@
...
+
+  # Directly call to r-lib actions
+  use_github_action(url = "https://raw.githubusercontent.com/r-lib/actions/master/examples/check-full.yaml")
+  expect_proj_dir(".github")
+  expect_proj_dir(".github/workflows")
+  expect_proj_file(".github/workflows/check-full.yaml")
+
+})
+
+test_that("use_github_action() replaces yml with yaml in name", {
```
jennybc, line 16: If I'm right above, I think this can be added as a second `use_github_action()` call in the test above, with `.yml` instead of `.yaml`.

tests/testthat/test-github-action.R
```diff
@@ -0,0 +1,43 @@
...
+  use_git_remote(name = "origin", url = "https://github.com/fake/fake")
+
+  # Directly call to r-lib actions
+  use_github_action("check-full.yml")
+  expect_proj_dir(".github")
+  expect_proj_dir(".github/workflows")
+  expect_proj_file(".github/workflows/check-full.yaml")
+})
+
+test_that("use_github_action() replaces appends yaml in name if missing", {
```
jennybc, line 30:
```suggestion
test_that("use_github_action() appends yaml in name if missing", {
```

R/github-action.R
coatless: @jennybc with @jimhester's scoping, the file extension is a pretty specific case to what is listed here: 

https://github.com/r-lib/actions/tree/master/examples

So, the following URL would work:

https://raw.githubusercontent.com/r-lib/actions/master/examples/pkgdown.yaml

But, this wouldn't:

https://raw.githubusercontent.com/r-lib/actions/master/examples/pkgdown.yml

As a result, I think a two-pass solution is only viable, where the first pass removes replaces `.yml` with `.yaml` if present and the second pass verifies the extension is present. 

Sorry, about the `=` assignment.

tests/testthat/test-github-action.R
coatless: The `.yml` / `.yaml` components would only be enforced when there isn't a custom URL. So, I don't think this can be merged in with the above test as it is checking:

```r
use_github_action("pkgdown.yml")
```

whereas the first test is ensuring that a custom URL can be pulled and downloaded.

```r
use_github_action(url="https://my-secret/gh/actions/master/examples/pkgdown.yaml")
```

jennybc:
@jimhester can you take a quick look here? If you 👍 before Monday-ish, it will be in the release.

R/github-action.R
jimhester: I don't think this extra logic is necessary, the examples in r-lib/actions (which is the only place that the 'name' parameter will be used) are always going be `.yaml`.

jimhester: I would revert df603d0adfbcf52c4b44676a8cb3a6f4d9962fef

coatless: Okay, reverted. Thanks @jimhester !

coatless:
@jennybc should be good to go now once the CI finishes :)

coatless:
@jennybc this passes cleanly. For whatever reason, the macOS-latest build was cancelled.

jennybc:
Thanks all!

PR #1068, opened 2020-03-18
Check that the internet is on

### LINKED ISSUE #766: pr_push() with no internet gives deceiving message
Trying to do `pr_finish()` with no internet:

``` r
> pr_finish()
✔ Setting active project to 'C:/Users/jenny/Documents/reprex'
✔ Switching back to 'master' branch
Error: GitHub `auth_token` is invalid.
See `browse_github_token()` for help storing a token as an environment variable.
```

An invalid auth token is clearly not the problem in this case.

hadley:
Fixes #766

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,7 +29,6 @@ Imports:
     gh,
     git2r (>= 0.23),
     glue (>= 1.3.0),
-    pingr,
     purrr,
     rematch2,
     rlang (>= 0.4.2),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -29,6 +29,7 @@ Imports:
     gh,
     git2r (>= 0.23),
     glue (>= 1.3.0),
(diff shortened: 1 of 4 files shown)


jennybc:
Do we want to add another package for this?

FWIW we already have a homegrown `skip_if_offline()`:

https://github.com/r-lib/usethis/blob/ef2354ccf350ae98c549dc952230069704355c3c/tests/testthat/helper.R#L122-L129

Which brings up various things:

  * Hey we *can* use `testthat::skip_if_offline()` now.
  * Alternatively, we could wrap `!is.null(curl::nslookup(host, error = FALSE))` as an internal helper, use it in our skipper, and use it elsewhere (like in this PR) to check for internet.

hadley:
Oh I missed that — the curl approach isn't quite as good as the pingr approach, but since we're already using there's no need to take a second approach.

PR #1071, opened 2020-03-19
Support more remote types in use_dev_package.

### LINKED ISSUE #970: `use_dev_package` should handle more remote types
`use_dev_package` can only handle dev packages installed from github remotes. 

A simple change to the `package_remote` function would make `use_dev_package` work with any remote type that has a `Remotes` field of the form `{RemoteType}::{RemoteUsername}/{RemoteRepo}`.

In particular, this would cover package dependencies from `GitLab` and `Bitbucket`.

Something like this:

```R
package_remote <- function(package) {
  desc <- desc::desc(package = package)
  remote_info <- desc$get(c("RemoteType", "RemoteUsername", "RemoteRepo"))
  
  supported_remotes <- c("github", "gitlab", "bitbucket")
  is_valid_remote <- 
    rlang::is_string(remote_info[["RemoteType"]]) && 
    remote_info[["RemoteType"]] %in% supported_remotes
  
  if (!is_valid_remote) {
    ui_stop("{ui_value(package)} was not installed from a supported remote type.")
  }
  
  glue::glue_data(as.list(remote_info), "{RemoteType}::{RemoteUsername}/{RemoteRepo}")
}
```

I don't really see a reason to restrict the remotes to github, gitlab and bitbucket though. I would just check that `RemoteType`, `RemoteUsername` and `RemoteRepo` are not `NA`.

The code above would add an explicit `github::` prefix to github remotes but this could obviously be avoided. IMO `usethis` is github-centric enough as it is, so I would leave the prefix for clarity.

This is related to #293 to the extent that it is about extending the functionalities beyond github, but this is a simpler (and probably more common) use case.

antoine-sachet:
Any remote of the form 'type::username/repo' is now supported by `use_dev_package`. Close #970.

The internal function `package_remote` only checks that the `RemoteType`, `RemoteUsername` and `RemoteRepo` are non-NA scalar string.

This means:
* All standard remotes like bitbucket, gitlab, etc will work just fine.
* ~~Github remotes will now be prefixed by `github::` in the DESCRIPTION~~
* As requested, github remotes will *not* get the `github::` prefix, in line with the existing behaviour.


R/package.R
```diff
@@ -62,19 +62,19 @@ use_remote <- function(package) {
...
+
+  # Check Remote{Type,Username,Repo} are all non-na scalar strings
+  is_valid_remote <- all(purrr::map_lgl(remote_info,
+                                        ~ rlang::is_string(.x) && !is.na(.x)))
+  if (!is_valid_remote) {
+    ui_stop("{ui_value(package)} was not installed from a supported remote.")
   }
 
-  glue::glue_data(as.list(github_info), "{RemoteUsername}/{RemoteRepo}")
+  glue::glue_data(as.list(remote_info), "{RemoteType}::{RemoteUsername}/{RemoteRepo}")
```
hadley: Can you conditionally remove `GitHub::`? i.e. preserve the existing behaviour for the case of GitHub remotes?

hadley:
Thanks @antoine-sachet!

PR #1074, opened 2020-03-19
Allow use_test() to work outside packages

### LINKED ISSUE #1017: use_testthat() fails for Shiny apps
In the next version of Shiny we're planning on releasing a new function for running tests in the application environment, `shiny::runTests()` https://github.com/rstudio/shiny/pull/2585

This will promote the use of `testthat` in Shiny apps.

I thought to start a discussion here on `usethis::use_testthat()`'s behavior with respect to this change. Currently, running it in a Shiny app results in the following error message:
```
Error: `use_testthat()` is designed to work with packages.
```
We hope to promote unit testing in Shiny apps moving forward, and I thought it might be good for `use_testthat()` to work in this case.

Unfortunately, Shiny doesn't have a predicate function to determine whether or not the current directory is plausibly a Shiny app. However, I think the logic's pretty simple: at a minimum, there has to be either an `app.R` or both a `server.R` and `ui.R` in the current directory.

So, one way forward would be to check for package *and* Shiny app plausibility, and bail otherwise, but as a non-expert on all things `usethis` I'm looking forward to hearing how other people think it should work.

Thanks in advance for your feedback, and consider me on the hook for whatever implementation is needed 👍

hadley:
Fixes #1017


alandipert:
Worked for me, thanks 🙇

PR #1080, opened 2020-03-26
Configure a dummy git user

### LINKED ISSUE #1070: Set up dummy git user in code coverage github action
So `skip_if_no_git_config()` isn't activated

jennybc:
Fixes #1070

Lesson learned along the way: the git config has to be global, not local, if it needs to be in force inside the scoped temporary projects we used widely in the tests.


jennybc:
Well, this appears to be "working" because we seem to have a test failure in a test that was previously skipped via `skip_if_no_git_config()` 😬

PR #1081, opened 2020-03-26
Ensure we can write_utf8() even when outside a project

jennybc:
This is the root cause of what we were talking about in https://github.com/r-lib/usethis/pull/1072#pullrequestreview-380448740 (my initial comment was off-base, but this was worth sorting out).

### DIFF
--- a/R/write.R
+++ b/R/write.R
@@ -92,13 +92,21 @@ read_utf8 <- function(path, n = -1L) {
   base::readLines(path, n = n, encoding = "UTF-8", warn = FALSE)
 }
 
-write_utf8 <- function(path, lines, append = FALSE, line_ending = proj_line_ending()) {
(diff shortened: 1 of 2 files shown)


jennybc:
I think the Windows failure is unlucky around trying to delete a temporary folder too quickly? We use this exact pattern in many other tests.

```
1. Error: write_utf8() can operate outside of a project  --------------------
217
[EBUSY] Failed to remove 'C:/Users/RUNNER~1/AppData/Local/Temp/Rtmp8OJp0R/file17204a2c1187': resource busy or locked
218
Backtrace:
219
 1. fs::dir_delete(tmpdir)
220
 2. fs:::rmdir_(rev(c(old, dirs)))
```

The failure in the test coverage run looks related to the git config stuff, which I will work on separately.

PR #1083, opened 2020-03-29
Equip tidy_download() with retries (with a longer connect timeout)

### LINKED ISSUE #988: Expose curl options in use_course() and friends?
Lesson learned from rstudio::conf 2020.

If the internet is functional but slow, `use_course()` can timeout. It inherits the default timeout from curl.

But I could possibly afford some way to set a longer timeout in an individual call or to do so globally via an env var or option.

Timeout can be set on the handle, i.e. `h <- new_handle(timeout = 10)` right around here:

https://github.com/r-lib/usethis/blob/4cae785a04bc2824619d8ae39ccd574852d00e6f/R/course.R#L208

jennybc:
Closes #988

### DIFF
--- a/R/course.R
+++ b/R/course.R
@@ -51,7 +51,7 @@ NULL
 #'  highlight the download destination. Workflow:
 #' * User executes, e.g., `use_course("bit.ly/xxx-yyy-zzz")`.
 #' * User is asked to notice and confirm the location of the new folder. Specify
-#'   `destdir` to prevent this.
+#'   `destdir` or configure the `"usethis.destdir"` option to prevent this.
 #' * User is asked if they'd like to delete the ZIP file.
 #' * If new folder contains an `.Rproj` file, a new instance of RStudio is
 #'   launched. Otherwise, the folder is opened in the file manager, e.g. Finder
@@ -129,6 +129,10 @@ use_zip <- function(url,
 #' in the `Content-Disposition` header. In the absence of this header, a
 #' filename is generated from the input URL. In either case, the filename is
 #' sanitized. Returns the path to downloaded ZIP file, invisibly.
+#'
+#' `tidy_download()` is setup to retry after a download failure. In an
+#' interactive session, it asks for user's consent. All retries use a longer
+#' connect timeout.
 #'
 #' ## DropBox
 #'
@@ -236,13 +240,14 @@ use_zip <- function(url,
 NULL
 
 # 1. downloads from `url`
-# 2. determines filename from content-description header (with fallbacks)
-# 3. returned path has content-type and content-description as attributes
+# 2. calls a retry-capable helper to download the ZIP file
+# 3. determines filename from content-description header (with fallbacks)
+# 4. returned path has content-type and content-description as attributes
 tidy_download <- function(url, destdir = getwd()) {
   check_path_is_directory(destdir)
   tmp <- file_temp("tidy-download-")
-  h <- curl::new_handle(noprogress = FALSE, progressfunction = progress_fun)
-  curl::curl_download(url, tmp, quiet = FALSE, mode = "wb", handle = h)
+
+  h <- download_url(url, destfile = tmp)
   ui_line()
 
   cd <- content_disposition(h)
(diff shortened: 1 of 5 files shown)


R/course.R
```diff
@@ -236,21 +240,22 @@ use_zip <- function(url,
...
+# 2. calls a retry-capable helper to download the ZIP file
+# 3. determines filename from content-description header (with fallbacks)
+# 4. returned path has content-type and content-description as attributes
 tidy_download <- function(url, destdir = getwd()) {
   check_path_is_directory(destdir)
   tmp <- file_temp("tidy-download-")
-  h <- curl::new_handle(noprogress = FALSE, progressfunction = progress_fun)
-  curl::curl_download(url, tmp, quiet = FALSE, mode = "wb", handle = h)
+
+  h <- download_url(url, destfile = tmp)
```
jennybc: We need the handle `h` here in order to query its content-disposition header (used for naming the destination directory) and content type.

R/course.R
```diff
@@ -259,6 +264,81 @@ tidy_download <- function(url, destdir = getwd()) {
   invisible(full_path)
 }
 
+download_url <- function(url,
+                         destfile,
+                         handle = curl::new_handle(),
+                         n_tries = 3,
+                         retry_connecttimeout = 40L) {
```
jennybc: Our first try is always with the curl default. The curl package defaults to 10s, which was clearly too short at rstudio::conf. 

All retries are with `retry_connecttimeout`. I chose 40 seconds because we (and others on the internet) seem to talk about values in the range of 30 to 60 second.

There are other knobs that affect timeout (see #988 for research), but the curl package doesn't set its own default for anything but `connecttimeout` and I believe it's the one that caused us grief. I'd only consider messing with the others if we have evidence we need to.

R/course.R
```diff
@@ -259,6 +264,81 @@ tidy_download <- function(url, destdir = getwd()) {
...
+  handle_options <- list(noprogress = FALSE, progressfunction = progress_fun)
+  curl::handle_setopt(handle, .list = handle_options)
+
+  we_should_retry <- function(i, n_tries, status) {
```
jennybc: I took some cues from `httr::RETRY()`:

https://github.com/r-lib/httr/blob/master/R/retry.R

R/course.R
```diff
@@ -259,6 +264,81 @@ tidy_download <- function(url, destdir = getwd()) {
...
+      # https://ec.haxx.se/usingcurl/usingcurl-timeouts
+      # "Failing to connect within the given time will cause curl to exit with a
+      # timeout exit code (28)."
+      # (however, note that all timeouts lead to this same exit code)
+      # https://ec.haxx.se/usingcurl/usingcurl-returns
+      # "28. Operation timeout. The specified time-out period was reached
+      # according to the conditions. curl offers several timeouts, and this exit
+      # code tells one of those timeout limits were reached."
+      # https://github.com/curl/curl/blob/272282a05416e42d2cc4a847a31fd457bc6cc827/lib/strerror.c#L143-L144
+      # "Timeout was reached" <-- actual message we could potentially match
```
jennybc: Ideally we would act, knowing what sort of error we got. But I think that's not possible today.

Silver living: I'm sure there are intermittent failures for reasons other than connect timeout, so this does give us a general way to retry a couple of times, for whatever reason.

R/course.R
```diff
@@ -259,6 +264,81 @@ tidy_download <- function(url, destdir = getwd()) {
...
+    }
+  }
+
+  i <- 1
+  # invariant: we have made i download attempts
+  while (we_should_retry(i, n_tries, status)) {
+    if (i == 1) {
+      curl::handle_setopt(
+        handle,
+        .list = c(connecttimeout = retry_connecttimeout))
```
jennybc: I have no idea how I can test that this is actually having the desired effect.

jeroen: You can try to connect to a dead host a look at `system.time()` afterwards maybe?

jennybc: Do you have any ideas of a specific "dead host"? Or method for finding one? If I simply disconnect from the internet, I get an immediate error about not being able to resolve the host name.

jeroen: It's a bit tricky indeed because you need a host that exists but does not respond. You could try to connect to a port that is not used. For example:

```r
h <- new_handle(connecttimeout=7, verbose=T)
system.time(curl_fetch_memory('http://cran.r-project.org:999', handle = h))
```

But this is also not waterproof because some servers might have a firewall that explicitly rejects the connection so it fails immediately. But the example above seems to work for me.

jennybc: Thanks, that example is pretty compelling. I mostly wanted to confirm that the name `connecttimeout` is correct and that I'm using `curl::handle_setopt()` correctly. So I think we're good on those fronts.

``` r
library(curl)

h <- new_handle(connecttimeout = 3, verbose = TRUE)
system.time(curl_fetch_memory('http://cran.r-project.org:999', handle = h))
#> Error in curl_fetch_memory("http://cran.r-project.org:999", handle = h): Timeout was reached: [cran.r-project.org:999] Connection timed out after 3003 milliseconds
#> Timing stopped at: 0.017 0.007 3.005

handle_setopt(h, connecttimeout = 6)
system.time(curl_fetch_memory('http://cran.r-project.org:999', handle = h))
#> Error in curl_fetch_memory("http://cran.r-project.org:999", handle = h): Timeout was reached: [cran.r-project.org:999] Connection timed out after 6003 milliseconds
#> Timing stopped at: 0.001 0 6.003

handle_setopt(h, .list = list(connecttimeout = 5))
system.time(curl_fetch_memory('http://cran.r-project.org:999', handle = h))
#> Error in curl_fetch_memory("http://cran.r-project.org:999", handle = h): Timeout was reached: [cran.r-project.org:999] Connection timed out after 5004 milliseconds
#> Timing stopped at: 0.001 0.001 5.005
```

<sup>Created on 2020-03-30 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

PR #1084, opened 2020-03-30
Prefer rlang::is_interactive(), is_string(), and seq2() to our DIY versions

### LINKED ISSUE #1082: Switch to rlang::is_interactive()
Switch to `rlang::is_interactive()` instead of the DIY version we're using now and shim `interactive()` to catch any future use of it.

jennybc:
Closes #1082

Not a super exciting PR. Review request is more of an FYI.

### DIFF
--- a/R/addin.R
+++ b/R/addin.R
@@ -9,7 +9,7 @@
 #' @inheritParams use_template
 #'
 #' @export
-use_addin <- function(addin = "new_addin", open = interactive()) {
+use_addin <- function(addin = "new_addin", open = rlang::is_interactive()) {
   addin_dcf_path <- proj_path("inst", "rstudio", "addins.dcf")
 
   if (!file_exists(addin_dcf_path)) {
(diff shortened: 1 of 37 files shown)


R/addin.R
```diff
@@ -9,7 +9,7 @@
 #' @inheritParams use_template
 #'
 #' @export
-use_addin <- function(addin = "new_addin", open = interactive()) {
+use_addin <- function(addin = "new_addin", open = rlang::is_interactive()) {
```
jennybc: I namespace this when it appears as a user-facing default.

Otherwise I don't because we import the rlang namespace.

jennybc:
> Can we remove `is_testing()` too?

Yes, but I will do that in a separate PR. It also involves de-janking the way we give a special dispensation for creating nested projects in the tests. I already have a tab open for this literally and mentally.

> Also in that same file is `is_string()` and `seq2()`

Done.

Issue #801, opened 2019-07-03
`use_data()` should prompt you to document your data set


DavisVaughan:
Currently `usethis::use_data(internal = FALSE)` just creates a `data/` folder and places the data in an rda file there. Since a `check()` warns that you should document your data files, it might be nice for usethis to prompt the user to do so as well. 

It might also go ahead and generate a data file in `R/data.R`.

Could it even go so far as to look at the columns in your dataset and generate a `@format` roxygen block that `\describe`s all the columns? That is done in the Documenting Data section in the link below.

See this section for exporting data, which mentions `usethis::use_data()`:
https://r-pkgs.org/data.html#data-data

And this section for documenting data sets, which could be updated if this feature is added:
https://r-pkgs.org/data.html#documenting-data

jennybc:
This has been discussed before:

https://github.com/r-lib/usethis/issues/274

and it looks like any actual "doing" re: creating the docs should happen in roxygen2.

So usethis could:

  * TODAY: have better messaging about the need to document exported datasets
  * FUTURE: call this hypothetical function on roxygen2 to actually do it

ijlyttle:
FWIW, we worked on something related at Uncoast Unconf: https://uncoast-unconf.github.io/steward/articles/steward.html#create

Admittedly, this approach may be a lot heavier than what you are proposing here, but if there is something in "steward" that can help the effort, you are welcome to it.

Issue #274, opened 2018-02-20
Data doc template?


maelle:
Would it make sense for `use_data` or another fonction to also help with data documenting, like the roxygen2 function skeleton one gets from RStudio menu?

I asked in https://community.rstudio.com/t/template-for-documenting-a-dataset/5365/2 whether such functionality already existed somewhere and it doesn't seem so from the (1!) answer I received.

jennybc:
Yes, it is interesting to think out loud whether this advice could be automated:

Documenting datasets section in <http://r-pkgs.had.co.nz/data.html>

If we can agree to an implementation, then you (or someone else) could implement in a PR!

Are you game to propose a sketch?

maelle:
:tada: I am happy to discuss, but couldn't implement it myself in the short term 😢 

I've thought a bit, not much, about it and here what came to mind:

* if the template creating becomes part of `use_data` then there should be an overwrite parameter, FALSE by default (because I could run the data preparation script again if I update the data, without wanting to overwrite the existing documentation).

* the template would be loaded in a brand new (if non existing) .R file called nameofthedata-docs.R?

* if the data in say `usethis::use_data_doc` is a data.frame the items could be filled with the names and types of the variables (a bit like what the roxygen2 skeleton does with function parameters), so the function wouldn't "just" load a template. Not sure about other data types.

* btw should the script where data is prepared when there's one be linked in the data documentation in details? not that this could be automated.

jimhester:
This would basically be a roxygen equivalent to the confusingly named [utils::prompt](https://www.rdocumentation.org/packages/utils/versions/3.4.3/topics/prompt) function. I agree this would be useful!

``` r
prompt(mtcars)
#> Created file named 'mtcars.Rd'.
#> Edit the file and move it to the appropriate directory.
readLines("mtcars.Rd")
#>  [1] "\\name{mtcars}"                                                               
#>  [2] "\\alias{mtcars}"                                                              
#>  [3] "\\docType{data}"                                                              
#>  [4] "\\title{"                                                                     
#>  [5] "%%   ~~ data name/kind ... ~~"                                                
#>  [6] "}"                                                                            
#>  [7] "\\description{"                                                               
#>  [8] "%%  ~~ A concise (1-5 lines) description of the dataset. ~~"                  
#>  [9] "}"                                                                            
#> [10] "\\usage{data(\"mtcars\")}"                                                    
#> [11] "\\format{"                                                                    
[...]

PR #1086, opened 2020-04-02
Proposal of a new feature, automatic documentation for data `add_data_doc()`

Cervangirard:
Hi, 
here is a function that allows to create automatically the documentation of a dataset :
```
use_data_doc('my_data', description = "My desc", source = "My source")
```

It creates doc_example.R in the R folder

Example:

```
#' my_data
#'
#' My desc.
#'
#' @format A data frame with 10 rows and 3 variables:
#' \describe{
#'   \item{ name }{  character }
#'   \item{ age }{  numeric }
#'   \item{ weight }{  numeric }
#' }
#' @source My source
"my_data"
```


jennybc:
This has come up a couple of times before, so I'll just link to those threads:

  * #801 `use_data()` should prompt you to document your data set
  * #274 Data doc template?

The TL;DR is that if any such things happens in must happen or start in roxygen2.

dcnorris:
FYI package 'sinew' https://stackoverflow.com/a/63782246/3338147

Issue #1087, opened 2020-04-03
use_tidy_labels() issues with preexisting labels


DavisVaughan:
yardstick has pre-existing issues labels (like `enhancement`). I went to run `use_tidy_labels()` on it and got back:

```r
> usethis::use_tidy_labels()
✓ Renaming labels: 'enhancement' -> 'feature'
  'good first issue' -> 'good first issue :heart:'
  'help wanted' -> 'help wanted :heart:'
  'question' -> 'reprex'
Error in out[[1]] : subscript out of bounds
```

I think this is related to the following issue, which has been fixed in the cran version of gh (the fix there is what causes the issue here)
https://github.com/r-lib/gh/issues/66

This map is what fails
https://github.com/r-lib/usethis/blob/54659c36c5391396535098209dd1e7ee6f25840a/R/github-labels.R#L110-L113

For yardstick, the `enhancement` label needs to be renamed but there are currently no issues with that label, so an empty response comes back. That means this branch of the usethis `gh()` function is run

https://github.com/r-lib/usethis/blob/54659c36c5391396535098209dd1e7ee6f25840a/R/github-labels.R#L84-L88

But since empty responses have changed in gh, the `out[[1]]` fails

jimhester:
I just ran into this myself

jennybc:
I think this was also fixed by #1088. Someone correct me if I'm wrong.

PR #1088, opened 2020-04-03
State minimum version for gh and remove conditional code

hadley:


DavisVaughan:
Closes #1087

jennybc:
There is at least one other place where we wrapped a gh call due to versions, in the `pr_*()` family. Let me find that and undo it here as well, if we're going to increase gh min version.

jennybc:
Yeah it's #991

PR #1089, opened 2020-04-03
Empty `gh()` responses are now correctly returned as classed empty lists

### LINKED ISSUE #1087: use_tidy_labels() issues with preexisting labels

DavisVaughan:
Closes #1087

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -30,7 +30,7 @@ Imports:
     curl (>= 2.7),
     desc,
     fs (>= 1.3.0),
-    gh,
+    gh (>= 1.1.0),
     git2r (>= 0.23),
     glue (>= 1.3.0),
     purrr,
--- a/R/github-labels.R
+++ b/R/github-labels.R
@@ -70,7 +70,7 @@ use_github_labels <- function(repo_spec = github_repo_spec(),
   check_github_token(auth_token)
 
   gh <- function(endpoint, ...) {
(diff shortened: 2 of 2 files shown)


DavisVaughan:
@hadley beat me to it!

DavisVaughan:
Closed in favor of #1088

PR #1093, opened 2020-04-07
Use separate test-coverage action to run test coverage

jimhester:
Rather than running test coverage after the check finishes, this action
runs it as a separate workflow

This can be useful when a package takes a while to compile and check, as
covr needs to recompile it anyway and measuring the coverage does not
really depend on a successful check

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -40,6 +40,7 @@ use_tidy_github_actions <- function() {
   full_status    <- use_github_action_check_full()
   pr_status      <- use_github_action_pr_commands()
   pkgdown_status <- use_github_action("pkgdown")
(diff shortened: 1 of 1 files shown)


jimhester:
I wrote this a few weeks ago, but apparently never sent the PR?

jennybc:
No, we've already been here! I remember too.

https://github.com/r-lib/usethis/commit/602acc7435c64013c3d9ce15c5e7c295eb06cd9f

jennybc:
These lines appear in current HEAD on `master`:

https://github.com/r-lib/usethis/blob/master/R/github-actions.R

jimhester:
Ok, I don't understand how this PR is mergeable then... I will close it.

PR #1100, opened 2020-04-13
More updates for create_tidy_package()

### LINKED ISSUE #1098: What to do about use_tidy_ci()?
While reviewing #1096, I noticed that `use_tidy_ci()` has not kept up with our transition away from Travis and AppVeyor and towards Github Actions:

https://github.com/r-lib/usethis/blob/a1321959b315f15f500cfb89fad550636df4a7d9/R/tidyverse.R#L98-L121

Do we want to just ignore it? Or do we want to make it do different things? As in, it could just be an alias for `use_tidy_github_actions()`:

https://github.com/r-lib/usethis/blob/0b5e8b363ea7e56cb2150110b767f5b17d701e6d/R/github-actions.R#L35-L58

cc @Hadley @jimhester

hadley:
Fixes #1098

### DIFF
--- a/R/tidyverse.R
+++ b/R/tidyverse.R
@@ -10,12 +10,6 @@
 #' of the tidyverse conventions as possible, issues a few reminders, and
 #' activates the new package.
 #'
-#' * `use_tidy_ci()`: sets up [Travis CI](https://travis-ci.com) and
-#' [Codecov](https://codecov.io), ensuring that the package is actively tested
-#' on the versions of R officially supported by the Tidyverse (current release,
-#' devel, and four previous versions). It also ignores `compat-` and `deprec-`
-#' files from code coverage.
-#'
 #' * `use_tidy_description()`: puts fields in standard order and alphabetises
 #'   dependencies.
 #'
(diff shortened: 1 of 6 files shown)


R/tidyverse.R
```diff
@@ -71,15 +65,16 @@ create_tidy_package <- function(path,
   old_project <- proj_set(path)
   on.exit(proj_set(old_project), add = TRUE)
 
-  use_roxygen_md()
```
hadley, line 74: This is now the default in `create_package()`

R/tidyverse.R
```diff
@@ -71,15 +65,16 @@ create_tidy_package <- function(path,
...
   use_testthat()
   use_mit_license(name)
   use_tidy_description()
 
   use_readme_rmd(open = FALSE)
   use_lifecycle_badge("experimental")
   use_cran_badge()
+
   use_cran_comments(open = FALSE)
+  use_tidy_release_test_env()
```
hadley, line 77: See results of applying this below.

PR #1109, opened 2020-04-17
Fix typo

salim-b:
and be more precise about Git terminology 😇


jennybc:
The problem is: this wording is too specific to always be correct. Depending on whether we're working in a fork or not, we might pull from `upstream` or `origin`.

https://github.com/r-lib/usethis/blob/a1321959b315f15f500cfb89fad550636df4a7d9/R/pr.R#L267-L280

salim-b:
> The problem is: this wording is too specific to always be correct. Depending on whether we're working in a fork or not, we might pull from `upstream` or `origin`.

I've rephrased it in https://github.com/r-lib/usethis/pull/1109/commits/fa2aa25879f975ca46d1a70ccfe9d45c6ec8027f to be more precise. Is this fine?

jennybc:
Thanks for this. I used it as a nudge to go over this whole article, which actually needed many other changes in light of the transition to gert.

PR #1110, opened 2020-04-17
Fix description

salim-b:


R/latest-dependencies.R
```diff
@@ -1,7 +1,7 @@
 #' Use "latest" versions of all dependencies
 #'
-#' Pins minimum version of dependencies to latest (as determined by "source)"
-#' Useful for tidyverse package, but should otherwise be used with extreme
+#' Pins minimum versions of dependencies to latest ones (as determined by `source`).
+#' Useful for tidyverse packages, but should otherwise be used with extreme
```
jennybc, line 4: We really do mean **the** tidyverse package here. As in, the meta-package.

This is internal documentation, btw.

salim-b: ok, fixed

PR #1114, opened 2020-04-23
Fixes for reverted behavior of fs::path_real

jimhester:
As of https://github.com/r-lib/fs/commit/2c3024f2a02510610f454f2e4dbf607a1a340ad2 `fs::path_real()` again works only on file paths which actually exist on the file system. Unfortunately trying to support paths which only partially exist was very error prone across operating systems, particularly Windows.

This change unfortunately breaks the current version of usethis. The changes in this PR will fix the issues. I also checked the usethis reverse dependencies and verified they work with the devel versions of usethis and changes in this PR.


jennybc:
What sort of release timeline do you have in mind for fs?

jimhester:
Soon, in the next week at the latest. If we can't do a usethis release that soon I can keep this out of the fs release until the next one.

jennybc:
I think we can do a quick usethis release, since there was a "real" usethis release so recently. In some sense, the sooner the better.

@Hadley Do you want me to just make this happen? Meaning: a quick usethis patch release.

hadley:
@jennybc yeah, sounds good

jennybc:
Notes to myself:

Clean `R CMD check` locally with usethis `HEAD` and fs 1.4.1 (current CRAN version).

Widespread test failure locally with usethis `HEAD` and dev fs. >100 instances of this:

```
[ENOENT] Failed to realize '/var/folders/yx/3p5dt4jj1019st0x90vhm9rr0000gn/T/RtmpbztDd7/testpkg72b66a00a980': no such file or directory
```

Clean `R CMD check` locally with usethis from this PR and dev fs.

Clean `R CMD check` locally with usethis from this PR and fs 1.4.1 (current CRAN version).

So, yes, everything looks forward and backwards compatible.

jennybc:
Thanks for this -- especially the revdep checks!

PR #1116, opened 2020-04-28
Test with added PPA

jeroen:


jennybc:
Thanks, nice to see concrete proof of the root cause and a way to fix.

jennybc:
I assume this is superseded by later developments.

PR #1119, opened 2020-04-28
Test the new setup-r action with PPA support

jeroen:


jennybc:
Implemented this change directly in master because I can't merge a draft PR and I really wanted to start seeing legit results on ubuntu. Thanks!

jeroen:
OK! The idea was actually to just test it here, and wait until jim had merged https://github.com/r-lib/actions/pull/97 so that the problem would be solved by itself, without any changes in usethis.

jennybc:
Yeah I know I sort of jumped the gun, but I really wanted to push usethis out the door and wanted good data from ubuntu. The workflows are easy for us to change and don't go to CRAN anyway. I'll revert this later.

PR #1131, opened 2020-05-06
Pkgdown cleanup

jennybc:
I looked into the pkgdown helpers, reflecting on some recent pre-release checks in usethis and devtools.

And discovered @Hadley, @jimhester, and I had all added various pieces of redundant and partially implemented functionality. So I cleaned it up.

The review request and tagging is as much about raising awareness than anything else.

### DIFF
--- a/R/code-of-conduct.R
+++ b/R/code-of-conduct.R
@@ -31,7 +31,8 @@ use_code_of_conduct <- function(path = NULL) {
     ignore = is_package() && is.null(path)
   )
 
-  href <- pkgdown_url() %||% "https://contributor-covenant.org/version/2/0"
+  href <- pkgdown_url(pedantic = TRUE) %||%
+    "https://contributor-covenant.org/version/2/0"
   href <- paste0(href, "/CODE_OF_CONDUCT.html")
 
   ui_todo("Don't forget to describe the code of conduct in your README:")
--- a/R/code-of-conduct.R
+++ b/R/code-of-conduct.R
@@ -10,8 +10,8 @@
 #' If your package is going to CRAN, the link to the CoC in your README must
 #' be an absolute link to a rendered website as `CODE_OF_CONDUCT.md` is not
 #' included in the package sent to CRAN. `use_code_of_conduct()` will
-#' automatically generate this link if you use pkgdown and have set the
-#' `url` field in `pkgdown.yml`; otherwise it'll link to
+#' automatically generate this link if (1) you use pkgdown and (2) have set the
+#' `url` field in `_pkgdown.yml`; otherwise it'll link to
 #' <https://contributor-covenant.org/version/2/0>.
 #'
 #' @param path Path of the directory to put `CODE_OF_CONDUCT.md` in, relative to
@@ -31,7 +31,7 @@ use_code_of_conduct <- function(path = NULL) {
     ignore = is_package() && is.null(path)
   )
 
-  href <- project_pkgdown_url() %||% "https://contributor-covenant.org/version/2/0"
+  href <- pkgdown_url() %||% "https://contributor-covenant.org/version/2/0"
   href <- paste0(href, "/CODE_OF_CONDUCT.html")
 
   ui_todo("Don't forget to describe the code of conduct in your README:")
--- a/R/pkgdown.R
+++ b/R/pkgdown.R
@@ -113,6 +113,28 @@ uses_pkgdown <- function() {
     file_exists(proj_path("pkgdown", "_pkgdown.yml"))
 }
 
+
+project_pkgdown <- function(base_path = proj_get()) {
+  path <- path_first_existing(
+    base_path,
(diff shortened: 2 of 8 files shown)


R/code-of-conduct.R
```diff
@@ -10,8 +10,8 @@
 #' If your package is going to CRAN, the link to the CoC in your README must
 #' be an absolute link to a rendered website as `CODE_OF_CONDUCT.md` is not
 #' included in the package sent to CRAN. `use_code_of_conduct()` will
-#' automatically generate this link if you use pkgdown and have set the
-#' `url` field in `pkgdown.yml`; otherwise it'll link to
+#' automatically generate this link if (1) you use pkgdown and (2) have set the
+#' `url` field in `_pkgdown.yml`; otherwise it'll link to
```
jennybc, line 14: If we're going to be specific, it should at least be the most common config filename.

R/edit.R
```diff
@@ -15,21 +15,22 @@
 #' edit_file("DESCRIPTION")
 #' edit_file("~/.gitconfig")
 #' }
-edit_file <- function(path, open = rlang::is_interactive()) {
+edit_file <- function(path, open = is_interactive()) {
```
jennybc, line 18: This change is actually already in `master` but I forgot to push that first.

R/release.R
```diff
@@ -40,7 +40,7 @@ release_checklist <- function(version, on_cran) {
   cran_results <- cran_results_url()
   has_src <- dir_exists(proj_path("src"))
   has_news <- file_exists(proj_path("NEWS.md"))
-  has_pkgdown <- file_exists(proj_path("_pkgdown.yml"))
+  has_pkgdown <- uses_pkgdown()
```
jennybc, line 43: This is what originally came here to fix.

tests/testthat/test-pkgdown.R
```diff
@@ -0,0 +1,60 @@
...
+  create_local_package()
+  use_pkgdown()
+  expect_true(uses_pkgdown())
+  expect_true(is_build_ignored("^_pkgdown\\.yml$"))
+  expect_true(is_build_ignored("^docs$"))
+})
+
+# pkgdown helpers ----
+test_that("pkgdown helpers behave in the absence of pkgdown", {
+  local_mock(`usethis:::is_installed` = function(pkg) TRUE)
```
hadley, line 17: Shouldn't that be `FALSE`?

R/pkgdown.R
```diff
@@ -28,6 +28,11 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
   use_build_ignore("pkgdown")
   use_git_ignore(destdir)
 
+  ui_todo(
```
hadley, line 31: I wonder if this is actually sufficiently important we should force people to supply it?

R/pkgdown.R
```diff
@@ -41,6 +46,52 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
...
+  } else {
+    gsub("/$", "", url)
+  }
+}
+
+# travis ----
+
 #' @export
 #' @rdname use_pkgdown
 use_pkgdown_travis <- function() {
```
hadley, line 97: Since you're in here, maybe it's good idea to seriously de-promote `use_pkgdown_travis()` since we no longer recommend it?

R/pkgdown.R
```diff
@@ -28,6 +28,11 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
   use_build_ignore("pkgdown")
```
hadley, line 28: Should be (build + git) ignoring `docs` here too? (I think so, given that the github actions flow is quite straighforward)

tests/testthat/test-pkgdown.R
jennybc: No, this is so that the test can be run even if pkgdown is not installed. You don't really need pkgdown to check the pkgdown helpers. We have a similar pattern in many tests, when we do setup related to package X, but package X is not in usethis's Imports (or, often, not even in Suggests).

R/pkgdown.R
jennybc: I think you often don't know it at this point? I wouldn't want it to be a hard blocker.

R/pkgdown.R
jennybc: I already did some of that in one of your PRs before I merged it. Do you think I should do more / different?

https://github.com/r-lib/usethis/pull/1106/commits/90dfa97c75c642c6b31119967f03ff93cef930ef

R/pkgdown.R
jennybc: We do that already:

https://github.com/r-lib/usethis/blob/94a389144d661d392ccf9636b209e04285704b62/R/pkgdown.R#L27-L29

PR #1141, opened 2020-05-26
Don't remind to check pkgdown index for patch release

hadley:


jennybc:
I don't care deeply but I would lean towards leaving this reminder in. I feel like there's often a straggler and it doesn't feel correlated with release type (i.e. they are are often left over from not checking this in prior minor/major releases).

hadley:
It doesn't feel right to me to do in a patch release, since you shouldn't generally be adding new functions in a patch release.

PR #1160, opened 2020-06-18
Add use_cpp11 function

jimhester:


inst/templates/code-cpp11.cpp
```diff
@@ -0,0 +1,5 @@
+#include <cpp11.h>
+using namespace cpp11;
+
+[[cpp11::export]]
+void fun() {}
```
jimhester, line 5: It seemed useful to me to have an example of the exporting syntax in the template file, let me know if you don't agree, currently the Rcpp example does not have this.

jennybc:
Thanks again for letting me use this PR as a target for manual tests of `pr_*` functionality in the gert branch. I continue to dole out little changes, but so far I think they all remain legit. Please weigh in if you beg to differ.

I won't be merging anything until the gert PR lands.

jimhester:
I realized this should also add a `C++11` system requirement, so I wrote a `use_system_requirement()` helper function. It would be useful to review that if you have the time.

R/helpers.R
```diff
@@ -90,6 +90,26 @@ use_dependency <- function(package, type, min_version = NULL) {
   invisible()
 }
 
+use_system_requirement <- function(requirement) {
+  stopifnot(is_string(requirement))
+  existing_requirements <- desc::desc_get_field("SystemRequirements", default = character(), file = proj_get())
+  existing_requirements <- utils::head(strsplit(existing_requirements, ", ?"), n = 1)
```
jimhester, line 96: I used head here rather than `[[1]]` to deal with the 0 length case, which is common if the package does not already have a SystemRequirements field. e.g.

``` r
utils::head(character(), n = 1)
#> character(0)
character()[[1]]
#> Error in character()[[1]]: subscript out of bounds
```

<sup>Created on 2020-07-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

jennybc:
Thanks!

PR #1192, opened 2020-07-22
Fix use_tutorial

### LINKED ISSUE #1178: Tutorials created with use_tutorial can't be found by learnr::run_tutorial
I'm currently working on a "tutorial package" for tidymodels called [learntidymodels](https://github.com/tidymodels/learntidymodels). 
We expect users to install this package and open their tutorial of choice locally by using `learnr::run_tutorial()` function.

When I use `usethis::use_tutorial` to add tutorials to the package, it adds the newly created tutorial `.Rmd` to `.Rbuildignore`. Consequently, `learnr::run_tutorial()` or `learnr::available_tutorials()` functions can't find the tutorial(s) and throw an error. 

For example:
```
library(usethis)
path <- file.path(tempdir(), "mypkg")
create_package(path)
use_tutorial("tutorial", "Some tutorial")
devtools::install()

learnr::run_tutorial("tutorial", "mypkg")
#> Error: No tutorials found for package: "mypkg"
```

I am not sure if there's a specific reason why these tutorial Rmds should be included in `.Rbuildignore`. 
But when these `.Rmd` files **are not** added to `.Rbuildignore` everything works fine and package still seems to pass all R CMD checks. On the other hand `use_tutoria()` adds the rendered html files to `.gitignore` and it may make sense to add these to `.Rbuildignore` as well?

karaesmen:
Fixes #1178 

It changes the `use_tutorial` function so the tutorial `Rmd` is no longer added in `.Rbuildignore`, which now makes the tutorial discoverable to `learnr::run_tutorial`. 

On the other hand, it now build ignores  the `html` file generated from rendering the tutorial.

### DIFF
--- a/R/tutorial.R
+++ b/R/tutorial.R
@@ -33,9 +33,6 @@ use_tutorial <- function(name, title, open = rlang::is_interactive()) {
   use_git_ignore("*.html", directory = dir_path)
   use_dependency("learnr", "Suggests")
 
-  html_path <- path(dir_path, asciify(name), ext = ".html")
-  use_build_ignore(html_path)
-
   path <- path(dir_path, asciify(name), ext = "Rmd")
 
   data <- project_data()
--- a/R/tutorial.R
+++ b/R/tutorial.R
@@ -33,6 +33,9 @@ use_tutorial <- function(name, title, open = rlang::is_interactive()) {
   use_git_ignore("*.html", directory = dir_path)
   use_dependency("learnr", "Suggests")
 
+  html_path <- path(dir_path, asciify(name), ext = ".html")
+  use_build_ignore(html_path)
+
   path <- path(dir_path, asciify(name), ext = "Rmd")
 
   data <- project_data()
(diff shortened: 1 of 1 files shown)


R/tutorial.R
```diff
@@ -33,6 +33,9 @@ use_tutorial <- function(name, title, open = rlang::is_interactive()) {
   use_git_ignore("*.html", directory = dir_path)
   use_dependency("learnr", "Suggests")
 
+  html_path <- path(dir_path, asciify(name), ext = ".html")
+  use_build_ignore(html_path)
+
```
jennybc, line 38: Is this not already handled by the `use_git_ignore()` call 2 lines above?

karaesmen: It's my bad! I thought not including the`.html` in `.Rbuildignore` would fail the local check, but that doesn't seem to be the case. Fixing it now.

jennybc: I prefer the smallest change possible, so let's merge this. If we have more ignoring to do, we'll find out! But I think this is good to go.

jennybc:
Thanks!

PR #1193, opened 2020-07-23
Testthat 3e

hadley:
Remaining issues:

* Need to switch to mockery/mockr
* glue/fs_path no longer compares equal to character causing a bunch of spurious failures


hadley:
I cherry-picked out the removal of `context()`; along with the re-colouring, I think this is all we need to do now.

PR #1195, opened 2020-07-29
Manually install libgit2-dev

jimhester:
It is not included in the system requirements database because it
conflicts with libcurl4-openssl-dev. But the `setup-r` action include
the travis ppa which has a backport for libgit2-dev which does work
without conflicts


jennybc:
Thanks!

Issue #1198, opened 2020-08-02
use_gpl3_license() modifies GPL-3 license text


fweber144:
Calling `usethis::use_gpl3_license("My name")` produces a modified `LICENSE.md` file, with "modified" denoting modifications compared to the [original license text](https://www.gnu.org/licenses/gpl-3.0.md). In particular, the name of the R package and the text passed to `usethis::use_gpl3_license()`'s argument `name` (`"My name"` in the example above) are used in section "How to Apply These Terms to Your New Programs" of the license text. However, it seems like modifying the GPL-3 license text is not allowed (not even for section "How to Apply These Terms to Your New Programs"), see for example [here](https://softwareengineering.stackexchange.com/questions/281949/how-to-properly-apply-a-license). I'm not an expert for legal affairs (and I also don't know if the person who answered in that Stack Exchange thread is one), but it makes sense to me that this section "How to Apply These Terms to Your New Programs" is intentioned to be a template and therefore must be included unmodified.

jennybc:
Would you like to make a PR showing how you think it should be? I don't think we're intentionally doing anything creative here.

fweber144:
Done: [https://github.com/r-lib/usethis/pull/1199](https://github.com/r-lib/usethis/pull/1199). However, in that PR, I just demonstrated my suggestion for `use_gpl3_license()`. You will probably have to adopt it for the other GPL-like licenses, too.

PR #1199, opened 2020-08-04
Fix issue #1198 ("use_gpl3_license() modifies GPL-3 license text")

fweber144:
This PR is intended to fix [issue #1198](https://github.com/r-lib/usethis/issues/1198) ("use_gpl3_license() modifies GPL-3 license text"). However, I just demonstrated my suggestion for `use_gpl3_license()`. You will probably have to adopt it for the other GPL-like licenses, too.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -44,6 +44,8 @@ There is increasing interest in making the name of a repo's default branch confi
 
 ## Other changes
 
+* `use_gpl3_license()` now uses the original (unmodified) GPL-3 license text available at [https://www.gnu.org/licenses](https://www.gnu.org/licenses), so it does not fill out the templates provided in the license's section "How to Apply These Terms to Your New Programs" anymore.
+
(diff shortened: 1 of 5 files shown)


fweber144:
I forgot to update the documentation; I will solve that asap.

fweber144:
Documentation is fixed. I also updated the NEWS file and updated the corresponding testthat test. I'm sorry for any inconvenience caused by creating a premature PR. I hope this PR is OK now. But keep in mind that this PR still only applies to `use_gpl3_license()` and probably needs to be adopted for the other GPL-like licenses.

jennybc:
@Hadley Do you have any objection to this? It means our template does less. It becomes a verbatim file.

In two places, we'd leave placeholders as placeholders, instead of populating them via whisker:

https://github.com/r-lib/usethis/blame/00e060017b5e18c6ba5e122d153173419a7f35bb/inst/templates/license-GPL-3.md#L556

https://github.com/r-lib/usethis/blame/00e060017b5e18c6ba5e122d153173419a7f35bb/inst/templates/license-GPL-3.md#L576

hadley:
I think this is correct, although I'd prefer to maintain the existing markdown formatting, just removing the existing whisker templates.

hadley:
Thanks for the PR and for brining this problem to our attention! I've wrapped the fix up into a more comprehensive set of changes at #1208

PR #1208, opened 2020-09-10
Licensing improvements

### LINKED ISSUE #1198: use_gpl3_license() modifies GPL-3 license text

### LINKED ISSUE #1206: GPL license tweaks
GPLv2 or greater is a little friendlier than GPLv3 or greater since it's clearly directly compatible with R's license. We should provide this as a default, possibly deprecating `use_gpl3_license()` (maybe with new `version` argument).

In either case, we should default to using "or greater" as otherwise packages that use GPL v2 are not automatically compatible with GPLv3. See (e.g) https://opensource.stackexchange.com/questions/1777/why-is-gplv2-incompatible-with-gplv3

### LINKED ISSUE #1207: MIT license tweaks
Should default copyright holder to "{package} authors" since that's a more accurate statement.

hadley:
* Make it easier to license non-packages
* Don't change licenses that aren't templates
* Change default copyright holder for MIT license
* Clarify argument names
* GPL and Apache licenses get version arguments
* Update/simplify tests

Fixes #1198. Fixes #1207. Fixes #1206.


R/license.R
```diff
@@ -4,150 +4,190 @@
...
-
+check_license_version <- function(version, possible) {
+  version <- as.double(version)
 
-find_name <- function() {
-  name <- getOption("usethis.full_name")
-  if (!is.null(name)) {
-    return(name)
+  if (!version %in% possible) {
+    possible <- glue_collapse(possible, sep = "", last = "or, ")
```
jennybc, line 174:
```suggestion
    possible <- glue_collapse(possible, sep = ", ", last = " or ")
```
Not sure exactly what inputs are anticipated but this is more what I expected to see.

tests/testthat/test-license.R
```diff
@@ -0,0 +1,40 @@
...
+  expect_true(is_build_ignored("^LICENSE\\.md$"))
+
+  expect_proj_file("LICENSE")
+  expect_false(is_build_ignored("^LICENSE$"))
+})
+
+test_that("other licenses work without error", {
+  create_local_package()
+
+  expect_error(use_agpl_license(3), NA)
```
jennybc, line 17: We have `expect_error_free()` if you want to use it.

https://github.com/r-lib/usethis/blob/master/tests/testthat/helper.R#L110-L112

jennybc (approved):

NEWS.md
```diff
@@ -42,6 +42,35 @@ There is increasing interest in making the name of a repo's default branch confi
...
+   
+* `use_mit_license()` now sets the default copyright holder to 
+  "{package} authors". This makes it more clear that the copyright holders
+  are the contributors to the package; unless you are using a CLA there is no
+  one copyright holder of a package (#1207).
+  
+* New `use_gpl_license()` and  `use_agpl_license()` make it easier to pick 
+  specific versions of the three GPL license, and to choose whether or not you 
+  include future versions of the license. `use_gpl_license()` defaults to 
+  version 2 (and above) so that your code is compatible with R itself;
```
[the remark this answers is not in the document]
hadley: Oh I missed that R is GPLv2 or GPLv3; that simplifies things a lot.

R/license.R
hadley: Oops!

tests/testthat/test-license.R
hadley: I don't really like that helper 😐

hadley:
(I'm looking into the 3.3 failure — the root cause appears to be systemfonts)

PR #1211, opened 2020-09-11
write_utf8()'s path

### LINKED ISSUE #1125: `proj_line_ending()` breaks `write_utf8()` when executed outside of the target project
If you use `write_utf8()` (possibly through the `write_over()` exported function) to write programmatically from outside of any project (eg opening RStudio and selecting "close project", ie "Project: (none)" appear in the top-right corner) to a file which is into a project, the function possibly fails with the following error:

```r
Error: Path 'C:/Users/<user name>/Documents/' does not appear to be inside a project or package.
```
Note: if the call was run from inside a project which is not the one of the target file, it (re-)activate the current project.

The error is caused because of [here](https://github.com/r-lib/usethis/blob/master/R/write.R#L104) `write_utf8()` checks for `possibly_in_proj(path)` but in the following line ([here](https://github.com/r-lib/usethis/blob/master/R/write.R#L105)) it calls `proj_line_ending()` without passing that `path` to it.
In fact, `proj_line_ending()` does not have any arguments, but when [here](https://github.com/r-lib/usethis/blob/f5eb0e3240a50a9d757e1825b860475decae6dbf/R/line-ending.R#L3) it calls `project_name()` without any arguments, `project_name()` assign to its argument `base_path` the default value `proj_get()` ([here](https://github.com/r-lib/usethis/blob/1f4de552c8bd5238c6ec325f901ab9b546afe3d8/R/proj.R#L248)). Now, if we have called `write_utf8()` from inside a project, the returned name is that project name (which is the wrong one in this case), while if we are not in a project at all, it returns `proj_set(".")` ([here](https://github.com/r-lib/usethis/blob/1f4de552c8bd5238c6ec325f901ab9b546afe3d8/R/proj.R#L57)), which calls `proj_set(".")` ([here](https://github.com/r-lib/usethis/blob/1f4de552c8bd5238c6ec325f901ab9b546afe3d8/R/proj.R#L57)) because outside of a project, and possibly returns the mentioned error because (normally) outside of any project `"."` is `"~"`.

My purposed solution is to add the same default argument of `proj_name()` (ie `base_path = proj_get()`) to `proj_line_ending()`. Next, pass `base_path` into its call to `proj_name()`. Ie updating `proj_line_ending()` to

```r
proj_line_ending <- function(base_path = proj_get()) {
  # First look in .Rproj file
  proj_path <- proj_path(paste0(project_name(base_path), ".Rproj"))
  if (file_exists(proj_path)) {
    config <- read_utf8(proj_path)

    if (any(grepl("^LineEndingConversion: Posix", config))) {
      return("\n")
    } else if (any(grepl("^LineEndingConversion: Windows", config))) {
      return("\r\n")
    }
  }
```
This way every call to `proj_line_ending()` (ie without explicit arguments) should not change it previous behaviour.

Next, update `write_utf8` [here](https://github.com/r-lib/usethis/blob/400306aabedd01710559aa5f4984754191c28dde/R/write.R#L105) to

```r
write_utf8 <- function(path, lines, append = FALSE, line_ending = NULL) {
...
   if (possibly_in_proj(path)) {
      line_ending <- proj_line_ending(path)
    } else {
...
}
```

CorradoLanera:
This fix #1125

(Sorry for the previous PR I made from the wrong branch)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,13 @@
 # usethis (development version)
 
+
+`write_utf8()` now works even if `path` is a project's path which is 
+not the "current" project's path. (@CorradoLanera, #1125)
+
+`proj_line_ending()` now has an argument `base_path` (default set
+to `proj_get()`) make it able to work with different path from the 
+default. (@CorradoLanera, #1125)
+
 ## Adoption of gert and credentials
 
 Usethis now uses the gert package for Git operations (<https://docs.ropensci.org/gert>), where previously we used git2r. The main motivation for this is to provide a smoother user experience by discovering and using the same credentials as command line Git (and, therefore, the same as RStudio). Under the hood, the hero is the credentials package (<https://docs.ropensci.org/credentials/>). `use_git_credentials()` and `git_credentials()` are now deprecated, since usethis no longer needs a way for a user to provide explicit credentials. The switch to gert + credentials should eliminate most credential-finding fiascos, but if you want to learn more, see the [introductory vignette](https://cran.r-project.org/web/packages/credentials/vignettes/intro.html) for the credentials package. 
--- a/R/line-ending.R
+++ b/R/line-ending.R
@@ -1,6 +1,6 @@
-proj_line_ending <- function() {
+proj_line_ending <- function(base_path = proj_get()) {
   # First look in .Rproj file
-  proj_path <- proj_path(paste0(project_name(), ".Rproj"))
+  proj_path <- proj_path(paste0(project_name(base_path), ".Rproj"))
   if (file_exists(proj_path)) {
     config <- read_utf8(proj_path)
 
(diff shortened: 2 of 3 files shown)


jennybc:
This is an enormous diff.

Screen Shot 2020-09-11 at 10 18 55 AM


Something seems to have gone wrong in the PR initiation process, such as not basing it on current `master`. If you are willing to use `usethis::pr_init()`, it should get things rolling in the right way.

CorradoLanera:
@jennybc now everything should be ok, sorry for the inconvenience.

jennybc:
When we review this PR, we should refer to the notes in "principles.md":

https://github.com/r-lib/usethis/blob/master/principles.md#helper-functions

It provides some context for figuring out what to do and this example may provide some basis for clarifying policies around helpers and the active project.

PR #1213, opened 2020-09-14
Experimenting with gitcreds

jennybc:
This is not going to get merged in this form, but is an experiment.


jennybc:
Concluding this experiment. We've now merged a PR into gh that has `gh::gh_token()` wired up to gitcreds, so the next pass here will be based on that.

PR #1214, opened 2020-09-14
Sync usethis with current cpp11 conventions

### LINKED ISSUE #1212: Problems with use_cpp11()
@jimhester - I wanted to try out the cpp11 package in a new project, and so I used the `use_cpp11()` function from the development version of usethis.

This function creates a source file with the contents:

https://github.com/r-lib/usethis/blob/fdba11320b96853affc1ea9605f13a0171fc319f/inst/templates/code-cpp11.cpp#L1-L5

But to successfully compile, I needed to change lines 1 & 4:

```c++
 #include <cpp11.hpp> 
 using namespace cpp11; 
  
 [[cpp11::register]] 
 void fun() {} 
```

Perhaps the cpp11 interface has changed slightly since `use_cpp11()` was created?

Also, it took me a little while to realize that [all these packages](https://github.com/r-lib/cpp11/blob/0e0442d54d1b24ba1890e5893bada428c2a82a8d/R/register.R#L35) (brio, cli, decor, desc, glue, tibble, vctrs) must be installed, in order to develop with cpp11. They're only listed in the `Suggests` list, which makes total sense from the end user's perspective. But perhaps `use_cpp11()` could check if these are installed?

jimhester:
Fixes #1212

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,7 +46,6 @@ Imports:
     yaml
 Suggests: 
     covr,
-    cpp11,
     knitr,
     magick,
     pkgdown (>= 1.4.0),
@@ -56,8 +55,7 @@ Suggests:
     styler (>= 1.2.0),
     testthat (>= 2.1.0)
 Remotes: 
-    r-lib/gert,
-    r-lib/cpp11
+    r-lib/gert
 Encoding: UTF-8
 Language: en-US
 LazyData: true
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -55,7 +55,8 @@ Suggests:
     styler (>= 1.2.0),
     testthat (>= 2.1.0)
 Remotes: 
-    r-lib/gert
+    r-lib/gert,
+    r-lib/cpp11
 Encoding: UTF-8
 Language: en-US
 LazyData: true
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -46,6 +46,7 @@ Imports:
     yaml
 Suggests: 
     covr,
+    cpp11,
     knitr,
     magick,
     pkgdown (>= 1.4.0),
--- a/R/cpp11.R
+++ b/R/cpp11.R
@@ -11,7 +11,6 @@
 use_cpp11 <- function() {
   check_is_package("use_cpp11()")
   check_uses_roxygen("use_cpp11()")
-
   use_src()
 
   use_dependency("cpp11", "LinkingTo")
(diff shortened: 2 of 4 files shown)


R/cpp11.R
```diff
@@ -23,5 +22,21 @@ use_cpp11 <- function() {
     open = is_interactive()
   )
 
+  check_cpp_register_deps()
+
   invisible()
 }
+
+get_cpp_register_deps <- function() {
+  res <- desc::desc(package = "cpp11")$get_field("Config/Needs/cpp11/cpp_register")
```
jimhester, line 31: I added the packages needed for `cpp11::cpp_register()` in https://github.com/r-lib/cpp11/commit/bb6745fbc6ee4662c8fec28e6f9632c6b9855d69, not yet on CRAN, so I added cpp11 to the Remotes for now.

davidchall: That's a fantastic solution!

DESCRIPTION
```diff
@@ -55,7 +55,8 @@ Suggests:
     styler (>= 1.2.0),
     testthat (>= 2.1.0)
 Remotes: 
-    r-lib/gert
+    r-lib/gert,
+    r-lib/cpp11
```
jennybc, line 59: usethis itself does not Import cpp11, nor does it need to, I think.

This is relevant to the package `use_cpp11()` is acting on, yeah?

jimhester: Sorry this was left over from running testing this locally, I will remove it.

jimhester: Actually it is for usethis and cpp11 itself, `desc::desc(package = "cpp11")` needs a installed version of cpp11, so I should add cpp11 to Suggests I guess.

jennybc: Well, this is guarded by `use_dependency("cpp11", "LinkingTo")` which will error if cpp11 is not installed. We have a lot of precedence for this with, e.g., Rcpp, RcppArmadillo, data.table, and more. Otherwise, any package that usethis helps to setup becomes a dependency of usethis itself.

jennybc: Am I missing how cpp11 can't be handled in a similar way?

jimhester: Sure, it can, but how are you going to run the tests if it is not in suggests?

jimhester: Oh I guess we are using a mock in the tests, I can do the same for this.

jennybc:
Thanks!

The windows failure looks either like a random glitch or a ragg problem.

(I realize cpp11 probably is or eventually will be an indirect dependency of usethis, but it's nice to not make it an explicit one if we can help it.)

PR #1215, opened 2020-09-15
Use a try block to catch any errors in withr::defer

jimhester:
withr 2.3.0 now throws errors if they occur in `withr::defer()` expressions.

In previous versions of withr these errors were silently ignored.

There are cases in the usethis tests where `setwd()` and `fs::dir_delete()` throw errors. For example if `old_project` is `NULL` then `setwd()` will fail.

This PR restores the previous behavior, but you may want to revisit this function, as it stands if `old_project` is `NULL` then `dir` is not deleted, I am not sure if that is the intent or not.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -44,9 +44,11 @@ create_local_thing <- function(dir = file_temp(pattern = pattern),
       ui_silence({
         proj_set(old_project, force = TRUE)
       })
-      setwd(old_project)
-      ui_done("Deleting temporary project: {ui_path(dir)}")
-      fs::dir_delete(dir)
+      try(silent = TRUE, {
+        setwd(old_project)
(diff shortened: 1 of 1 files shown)


jennybc:
OK thanks, I'll sort it out.

Is this semi-urgent for you/us because of CRAN?

jimhester:
I submitted withr 2.3.0 on friday and this was caught by CRAN'sr revdep checks.

Strangely it is not caught by the cloud checks, nor if I run `R CMD check` from the command line, which is why I did catch this before submission, nor when running `devtools::test()` locally on usethis.

But it _is_ caught if I run `rcmdcheck::rcmdcheck()`, which is how I was able to reproduce it and figure out a fix.

jennybc:
Closing in favour of #1218

PR #1218, opened 2020-09-16
Patch release for upcoming withr

jennybc:
The only thing that *had* to change was this aspect of `create_local_thing()`:

* We should capture and restore original working directory explicitly and not assume original working directory == original project. Sometimes there is no active usethis project to start with (so, project is `NULL`), but obviously there is still a working directory. If you conflate the two, you might try to restore the working directory to `NULL`. This used to "work" because withr was swallowing the error, but going forward it will not work. This change improves usethis, in any case.

I used this chance to do a complete audit of active project before and after each test, which I believe should always be the same. I discovered 2 tests that violated this and fixed them. I also refactored `create_local_thing()` to reflect the structure recommended in [testthat's vignette about test fixtures](https://testthat.r-lib.org/articles/test-fixtures.html). The changes described in this paragraph were optional for the patch release, but made sense to do at the same time.

I backported a 3rd test fix that we made since v1.6.1 (https://github.com/r-lib/usethis/commit/e0064aa87321830312ae79f2816a5c76fe0b9ca3#diff-ad31bf57b6e9b29bcce272bc388ca1aa).

Lesson learned about checking: we discovered this via CRAN reverse dependency checks, which clearly run somewhere that does not fulfill any of rprojroot's criteria for a project. So, it hits the "no active project" scenario described above. We were able to replicate using `rcmdcheck::rcmdcheck()`. But we did not see these failures with `devtools::test()` or `devtools::check()` or our cloud revdepchecks, presumably because all of those run in such a way that there is some sort of active project at check time. Possible insight for the future: the safest thing to do is to check from a temp directory, completely decoupled from any project, package, or git repo.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -21,3 +21,4 @@
 ^inst/templates/.*\\.html$
 ^man/figures/usethis-full-logo\.png$
 ^man/figures/usethis-og-1280x640\.png$
+^internal$
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .RData
 inst/doc
 docs
+internal
--- a/CRAN-RELEASE
+++ b/CRAN-RELEASE
@@ -1,2 +1,2 @@
 This package was submitted to CRAN on 2020-09-17.
-Once it is accepted, delete this file and tag the release (commit c8ab540cb).
+Once it is accepted, delete this file and tag the release (commit 23a2cecef).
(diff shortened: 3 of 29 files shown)


jennybc:
No intention to merge this. But the changes that should move over (mostly) came in b1c8d7fd5f7450a8b73b1d5c5d449523416c01c2.

PR #1224, opened 2020-09-20
Use `name` instead of "R build status" in GHA badge

malcolmbarrett:
This PR updates `use_github_actions_badge()` to more generally support GHA badges, using `name` instead of "R build status". This essentially maintains the default behavior (the Rmd link and `ui_done()` message now say "R-CMD-check" instead of "R build Status", but the appearance of the badge itself does not change) while giving more informative messages when you're adding a badge for a different type of workflow besides "R-CMD-check".


jennybc:
Looks reasonable. Thanks!

Issue #1217, opened 2020-09-16
Eliminate on.exit()


jennybc:
Eliminate all uses of `on.exit()` and consider shimming, like we've done with `interactive()` (steering usage towards `is_interactive()`). Context: when usage of `on.exit()` slips in, it clobbers cleanup scheduled with `withr::local_*()` functions, often at some distance. Using `on.exit(add = TRUE, after = FALSE)` "rescues" things, but it's even easier and more consistent to stay within the withr toolkit.

PR #1226, opened 2020-09-23
Only disable crayon for individual tests

hadley:
Turned out to be simpler than I expected, and I think the payoff (nicely coloured tests) is worth it. The only minor head scratcher was finding the `on.exit()` with missing `add = TRUE`.


jennybc:
When I merge the giant PR I'm working on, I'm going to systematically eliminate all `on.exit()` calls (#1217)!

PR #1233, opened 2020-10-01
Improve use_testthat()

hadley:
* Better edition detection
* Always override existing values
* Make it easy to turn on parallel testing
* Error if try to set edition outside of package

### DIFF
--- a/R/test.R
+++ b/R/test.R
@@ -6,7 +6,7 @@
 #' @param edition testthat edition to use. Defaults to the latest edition,
 #'   i.e. the major version number of the currently installed testthat.
 #' @param parallel Should tests be run in parallel? Only supported in testthat
-#'   3.0.0; see <https://testthat.r-lib.org/articles/parallel.html> for
+#'   >= 3.0.0; see <https://testthat.r-lib.org/articles/parallel.html> for
 #'   details and caveats.
 #' @seealso [use_test()] to create individual test files
 #' @export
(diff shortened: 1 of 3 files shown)


R/test.R
```diff
@@ -5,6 +5,9 @@
 #'
 #' @param edition testthat edition to use. Defaults to the latest edition,
 #'   i.e. the major version number of the currently installed testthat.
+#' @param parallel Should tests be run in parallel? Only supported in testthat
+#'   3.0.0; see <https://testthat.r-lib.org/articles/parallel.html> for
```
jennybc, line 9:
```suggestion
#'   >= 3.0.0; see <https://testthat.r-lib.org/articles/parallel.html> for
```

R/test.R
```diff
@@ -46,14 +59,20 @@ use_testthat_impl <- function(edition = NULL) {
 }
 
 check_edition <- function(edition = NULL) {
-  testthat_version <- utils::packageVersion("testthat")[[1, 1]]
+  version <- utils::packageVersion("testthat")[[1, c(1, 2)]]
+  if (version[[2]] == "99") {
```
jennybc, line 63: Should we insert a TODO to simplify this once testthat releases?

jennybc (commented):
LGTM. Note that I did not run any code, just read and commented.

R/test.R
hadley: I think it would be slightly better to introduce a `major_version()` helper that wrapped this up, as it's a pattern we might need elsewhere.

PR #1236, opened 2020-10-03
Test on multiple versions of Linux

jeroen:
I know this is against tidyverse practices, but because you are building on libgit2 it is important to test on non-legacy versions of Linux. These Linux versions ship with other versions of libgit2, which could affect behavior of gert/usethis.


hadley:
I don't think it's against our practices, just something we don't normally believe is necessary (but clearly is here)

jennybc:
Shall we use either 20.04 or 18.04 as the main linux platform and then include 1 run with remaining one and with 16.04? So, basically, redistribute the CI efforts towards more recent ubuntu and away from 16.04?

jennybc:
Thanks @jeroen for the nudge. I rolled this into some larger updates I did to the workflow in 02667631fd5c86dd948be9a896a3b0e4ed7d6d80 and 791e2ad2e8c33db0a487e7f61a2ff880a0acf151.

PR #1239, opened 2020-10-05
Fix for git_branch_set_upstream

jeroen:
This function has gained an argument.


jennybc:
I saw some notifications from gert and ended up seeing this need and made same change (plus another, close by) already in a61e49167324d5878ff4fd79177d812fe9063ab5.

PR #1258, opened 2020-10-26
Also change working directory for temporary projects

### LINKED ISSUE #1244: Add `setwd` argument to `with_project()` and `local_project()`
I find that you usually want to change working directory to the temporarily active project. I think there should be an argument `setwd = TRUE` that does that, by default. And when it's `TRUE`, then cleanup would obviously restore original working directory, as well as the original active project.

@jimhester I know I talked you into using one of these functions recently in devtools. Would this have any downside for you?

jennybc:
Closes #1244

### DIFF
--- a/R/proj.R
+++ b/R/proj.R
@@ -98,48 +98,46 @@ proj_path <- function(..., ext = "") {
   path_norm(path(proj_get(), ..., ext = ext))
 }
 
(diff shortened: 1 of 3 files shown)


jennybc:
@jimhester I'm going to merge this but if you look it over and have any comment, let me know.

PR #1271, opened 2020-11-11
Introduce use_github_pages() and use_tidy_pkgdown()

### LINKED ISSUE #224: Could use_github_pages() turn on GitHub Pages?
There is already support for `use_github_labels` in `usethis`, it would be great to programmatically turn on GitHub Pages for a repo without having to visit the settings page.

```r
use_github_pages(source.branch = "master")
# Your site is published at https://username.github.io/repo
```
If the provided `source.branch` cannot be found then return the following error (a la http://style.tidyverse.org/error-messages.html#hints):

```r
use_github_pages(source.branch = "not-a-real-branch")
#> Can't find branch 'not-a-real-branch' in the list of branches for GitHub repo 'my-repo'.
#> Refer to https://github.com/username/my-repo/settings for further settings.
```

jennybc:
Closes #224


jennybc:
@Hadley Is this what you had in mind?

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+#'   be only (or at least primarily) a remote branch.
+#'   - Typically the `gh-pages` branch is managed by an automatic "build and
+#'     deploy" job, such as the one configured by
+#'     [`use_github_action("pkgdown")`][use_github_action()].
+#' * Publish from the `"/docs"` directory of a "regular" branch, probably the
+#'   repo's default branch.
+#'   - The user is assumed to have a plan for how they will manage the content
+#'     below `"/docs"`.
+#'
+
```
hadley, line 15:
```suggestion
#'
```

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
+#' Configure a GitHub Pages site
+#'
+#' Activates or reconfigures a GitHub Pages site for a project hosted on GitHub.
+#' This function anticipates two specific usage modes:
+#' * Publish from the root directory of a `gh-pages` branch, which is assumed to
+#'   be only (or at least primarily) a remote branch.
+#'   - Typically the `gh-pages` branch is managed by an automatic "build and
```
hadley, line 7: I don't love this nested list style. I think you could just run them into the top-level list.

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+#'     deploy" job, such as the one configured by
+#'     [`use_github_action("pkgdown")`][use_github_action()].
+#' * Publish from the `"/docs"` directory of a "regular" branch, probably the
+#'   repo's default branch.
+#'   - The user is assumed to have a plan for how they will manage the content
+#'     below `"/docs"`.
+#'
+
+#' @param branch Branch for the site source. Default is `gh-pages`, coupled with
```
hadley, line 16: Maybe `@param branch,path` since they're so intertwined?

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' use_github_pages()
+#' }
+use_github_pages <- function(branch = "gh-pages", path = "/", cname = NA) {
+  stopifnot(is_string(branch), is_string(path))
+  stopifnot(is.na(cname) || is.null(cname) || is_string(cname))
+  path <- match.arg(path, choices = c("/", "/docs"))
```
hadley, line 48: I'd be tempted not to validate this on the R side, so that if GitHub changes behaviour in the future, we don't need to touch this code.

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+  }
+  safe_gh <- purrr::safely(gh)
+
+  if (branch == "gh-pages") {
+    new_branch <- create_gh_pages_branch(tr, branch = "gh-pages")
+    if (new_branch) {
+      # merely creating gh-pages branch automatically activates publishing
+      # BUT we need to give the servers time to sync up before a new GET
+      # retrieves accurate info... ask me how I know
+      Sys.sleep(2)
```
hadley, line 71: BTW I've started noticing this behaviour even in the web UI — e.g. sometimes after I create an issue, I get a broken url which is resolved when refreshing.

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+    site$source$path != path ||
+    (is.null(cname) && !is.null(site$cname)) ||
+    (is_string(cname) && cname != site$cname)
+
+  if (need_update) {
+    args <- list(
+      endpoint = "PUT /repos/{owner}/{repo}/pages",
+      source= list(branch = branch, path = path)
+    )
+    if (is.null(cname) && !is.null(site$cname)) {
```
hadley, line 98: I'm not sure we should touch `cname` because it's confusing that it creates a fill (`docs/CNAME`) that will appear next time you pull.

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+  need_update <-
+    site$source$branch != branch ||
+    site$source$path != path ||
+    (is.null(cname) && !is.null(site$cname)) ||
+    (is_string(cname) && cname != site$cname)
+
+  if (need_update) {
+    args <- list(
+      endpoint = "PUT /repos/{owner}/{repo}/pages",
+      source= list(branch = branch, path = path)
```
hadley, line 96:
```suggestion
      source = list(branch = branch, path = path)
```

R/pkgdown.R
```diff
@@ -47,6 +47,93 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
...
+  pkgdown_url <- paste0("https://", cname)
+
+  config <- pkgdown_config_path()
+  ui_done("
+    Recording pkgdown URL ({ui_value(pkgdown_url)}) in {ui_path(config)}")
+  config_lines <- read_utf8(config)
+  config_lines <- config_lines[!grepl("^url:", config_lines)]
+  write_over(config, c(
+    paste("url:", pkgdown_url),
+    if (nzchar(config_lines[1])) "",
```
hadley, line 91:
```suggestion
    if (nzchar(config_lines[[1]])) "",
```

R/github-pages.R
```diff
@@ -0,0 +1,164 @@
...
+  stopifnot(is.na(cname) || is.null(cname) || is_string(cname))
+  path <- match.arg(path, choices = c("/", "/docs"))
+  tr <- target_repo(github_get = TRUE)
+  if (!isTRUE(tr$can_push)) {
+    ui_stop("
+      You don't seem to have push access for {ui_value(tr$repo_spec)}, which \\
+      is required to turn on GitHub Pages.")
+  }
+
+  gh <- function(endpoint, ...) {
```
hadley, line 56: Given that you've used it 3 times, feels like should be it's own function?

hadley (commented):
This looks great! It's above and beyond what I was expecting, and I can see that it's going to save a bunch time and protect us against many minor mistakes when creating websites.

R/pkgdown.R
```diff
@@ -47,6 +47,93 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
...
+#'     GitHub Pages
+#'   - The pkgdown website URL (if specified or inferred) is added to the
+#'     pkgdown configuration file and to the URL field of DESCRIPTION
+#'
+#' @param cname Optional, custom domain name. If the target repo belongs to the
+#'   tidyverse or r-lib GitHub organizations, this is formed automatically.
+#'
+#' @rdname tidyverse
+#' @export
+use_tidy_pkgdown <- function(cname = NULL) {
```
hadley, line 69: Just occurred to me: should we say something about tidytemplate for sites tidyverse org? Or maybe it's not worth it, given adding a new tidyverse package is a relatively rare occurrence.

R/github-pages.R
jennybc: It has to be defined internally this way, because we're referencing `tr`. But we use this pattern so much that, after this PR, I am going to do something to make it a bit nicer everywhere.

R/github-pages.R
jennybc: I have to leave these blank lanes or else my reflexive use of "reflow comment" in RStudio mangles all bullet lists.

R/github-pages.R
jennybc: I lean towards keeping it. Our primary use case is with an orphan `gh-pages` branch which most of us never pull or inspect in any way.

The only properties one can set about GitHub Pages are branch, path, cname, so it just feels odd to offer a programmatic way to set/reset 2 of them but not the third. Plus we really do need this to do complete setup with `use_tidy_pkgdown()`.

jennybc: The thing I am currently mulling over is how best to align `cname` here and `cname` in `use_tidy_pkgdown()`.

hadley: This file is usually generated by pkgdown from the `url` field and it makes me nervous to have two ways to achieving the same goal.

I don't think we need this for `use_tidy_pkgdown()`?

R/pkgdown.R
jennybc: I think it's possibly we should offer ourselves better templating for pkgdown config, but I definitely would not get into that here in this PR or, probably for v2.0.0.

R/github-pages.R
jennybc: So you propose leaving the `cname` arg/feature in `use_github_pages()` bug, but NOT using it when we call `use_github_pages()` inside `use_tidy_pkdown()`?

jennybc: The problem is ... someone other than us might want to use `use_tidy_pkgdown()` and ideally they'd have a way to make sure `url:` gets set in `_pkgdown.yaml`, in DESCRIPTION, in repo homepage.

hadley: What about adding a url parameter to `use_pkgdown()`? I think we can set it up with a default, so even if you don't know the final url when you create the pkgdown site, you're at least left with a placeholder in your yaml file that you can later edit.

jennybc: I think I got to a good place here. `use_tidy_pkgdown()` no longer as any arguments, specifically, no `cname` argument.

The default pkgdown URL is now determined by what GitHub Pages tells me it is.
If the repo is owned by tidyverse or r-lib, in an interactive session, I propose that we go ahead and configure the correct, eventual URL.

All the pkgdown URL business is now also in an (unexported) helper.

DESCRIPTION
```diff
@@ -57,7 +57,7 @@ RdMacros:
     lifecycle
 Remotes: 
     r-lib/gert,
-    r-lib/gh
+    r-lib/gh#139
```
krlmlr, line 60: This is going to fail as soon as the PR is closed. It's a bit safer (but not as descriptive) to refer to a branch with the `@branch` notation instead.

jennybc: I don't plan for this situation to last very long, but yes you're right.

jennybc: Both are going to break when we merge and delete the branch. I'll update this as soon as we merge in gh.

PR #1276, opened 2020-11-18
When bumping testthat edition, also set minimum version in DESCRIPTION

hadley:


jennybc (approved):
Just from reading, LGTM.

PR #1280, opened 2020-11-18
Complete the conversion to testthat 3e

hadley:
I did the bare minimum to get the tests running cleanly. You may want to do more cleanup, especially around glue.


jennybc:
@Hadley I finished this off and am merging. But after Thanksgiving, if you look it over and have comments, we can refine.

Overview of what I changed:

* Moved mockr from Imports to Suggests
* Created a wrapper around `mockr::with_mock()` that hardwires `.env = "usethis"` so tests can be developed interactively and so that `devtools::test_file()` can be used
* Created `glue_chr()`and `glue_data_chr()` wrappers
* Continued test cleanup, i.e. minimizing the code inside `with_mock()`

PR #1281, opened 2020-11-21
Transfer cfg object in check_ours_or_fork()

vinhtantran:
cfg is a required argument in both check_for_bad_config()and check_for_maybe_config(). Not transferring into these creates a bug that keeps emitting error.


jennybc:
Thanks!

PR #1286, opened 2020-11-26
Update `version.c` in `use_dev_version()`

lionel-:
With this change, we check for existence of a src/version.c file with a line that matches `package_version = "VERSION"`. If there is a match, it is updated to the new version.

This is useful for projects like rlang or vctrs which check at load time that the version stored in the dynlib corresponds to the installed version. https://github.com/r-lib/rlang/blob/v0.4.9/src/version.c


jennybc:
Thanks! Good addition.

PR #1287, opened 2020-11-26
Add leading newline before news header

lionel-:
Having the news header in the first line feels a bit wrong to me. It seems at least some style guides recommend spacing around headers, e.g. https://google.github.io/styleguide/docguide/style.html

### DIFF
--- a/R/news.R
+++ b/R/news.R
@@ -39,5 +39,5 @@ use_news_heading <- function(version) {
   }
 
   ui_done("Adding new heading to NEWS.md")
-  write_utf8(news_path, c(title, "", news))
+  write_utf8(news_path, c("", title, "", news))
 }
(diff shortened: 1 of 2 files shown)


jennybc:
I find empty lines at the top a file unsettling and always remove them when I find them. Maybe we come from different editor traditions?

I suppose you're referring to this:
https://google.github.io/styleguide/docguide/style.html#add-spacing-to-headings

> Prefer spacing after # and newlines before and after

But I interpret that as applying to a heading in the middle of a file (as the example then shows).

So I politely decline this one 😅

lionel-:
Looking around a bit I now realise that I'm the only one to maintain a new line at the top of files 😲
I'll try and change this habit for consistency.

PR #1291, opened 2020-11-28
tweak gh_token_help() for GHE

ijlyttle:
applies some straightforward fixes:

- surrounds host name with quotes
- adapts argument names to functions

PR #1292, opened 2020-11-28
apply host_hint/code_hint to create_github_token()

ijlyttle:
Brings host-specific feedback to `create_github_token()`.

This is very similar to #1291 - I waffled on attaching this change to that PR.


jennybc:
Also good.

PR #1295, opened 2020-11-30
reworking host_hint and code_hint

### LINKED ISSUE #1294: consolidate host_hint/code_hint
I found another place in the code where it would make sense to provide host-specific feedback, then I heard a voice saying something about "copy-pasting three times" 🤔 

I'd get it if you think this is overkill, and would be happy to make an alternative implementation to this PR.

ijlyttle:
fix #1294 

Given this is a more-substantial change than the previous PR's, I am mindful that I might not have put things in the "right place".


jennybc:
Thanks, seems like a good idea. I reworked it a bit, but am merging because I'll be offline for a couple of days and want to resolve this. Let me know if you have concerns and we can revise.

ijlyttle:
Looks great to me!

PR #1301, opened 2020-12-04
Don't import lifecycle package in `use_lifecycle()`

lionel-:
The main purpose of this function is to import badges in `man/figures` and using lifecycle badges only requires a build time dependency.

Instead, we suggest using `usethis::use_dependency()` in case lifecycle functions are needed.

### DIFF
--- a/R/lifecycle.R
+++ b/R/lifecycle.R
@@ -2,9 +2,9 @@
 #'
 #' @description
 #'
-#' Call this to copy the lifecycle badges into your package (`man/figures`) and
-#' remind you of the syntax to use them in the documentation of individual
-#' functions or arguments.
+#' This helper imports the lifecycle badges in the `man/figures`
+#' folder of your package. It also reminds you of the syntax to use
+#' them in the documentation of individual functions or arguments.
 #'
 #' See the [getting started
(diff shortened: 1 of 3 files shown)


R/lifecycle.R
```diff
@@ -2,9 +2,9 @@
 #'
 #' @description
 #'
-#' Call this to copy the lifecycle badges into your package (`man/figures`) and
-#' remind you of the syntax to use them in the documentation of individual
-#' functions or arguments.
+#' This helper imports the lifecycle badges in the `man/figures`
```
hadley, line 5:
```suggestion
#' This helper copies the lifecycle badges in to the `man/figures`
```

R/lifecycle.R
```diff
@@ -48,5 +44,10 @@ use_lifecycle <- function() {
     "- `r lifecycle::badge('archived')`"
   ))
 
+  ui_todo(c(
+    "If you want to use functions like `lifecycle::deprecate_soft()` in your package:",
+    "- `usethis::use_depency(\"lifecycle\", \"import\")`"
```
hadley, line 49:
```suggestion
    "- `usethis::use_package(\"lifecycle\")`"
```

hadley (approved):

PR #1302, opened 2020-12-04
Use markdown syntax for lifecycle badges

lionel-:
And remove RdMacros: from DESCRIPTION


R/ci.R
```diff
@@ -2,7 +2,7 @@
 #'
 #' @description
 #'
-#' \lifecycle{soft-deprecated}
+#' `r lifecycle::badge("soft-deprecated")`
```
jennybc, line 5: It's not clear to me from #1303 that, if we used usethis this itself for lifecycle setup, we would have access to a "soft-deprecated" badge (although we seem to have the svg in `man/figures`).

R/github-labels.R
```diff
@@ -22,10 +22,10 @@
...
 #' * `help wanted` indicates that a maintainer wants help on an issue.
 #'
-#' @param repo_spec,host,auth_token \lifecycle{defunct}: These arguments are
-#'   now deprecated and will be removed in the future. Any input provided via
-#'   these arguments is not used. The target repo, host, and auth token are all
-#'   now determined from the current project's Git remotes.
+#' @param repo_spec,host,auth_token `r lifecycle::badge("defunct")`: These
+#'   arguments are now deprecated and will be removed in the future. Any input
+#'   provided via these arguments is not used. The target repo, host, and auth
+#'   token are all now determined from the current project's Git remotes.
```
jennybc, line 28: Ditto. #1303 removes the "defunct" badge from lifecycle setup. Applies here and for many more below.

R/ci.R
```diff
@@ -2,7 +2,7 @@
 #'
 #' @description
 #'
-#' \lifecycle{soft-deprecated}
+#' `r lifecycle::badge("deprecated")`
 #'
 #' Some of these functions are now soft-deprecated since the tidyverse team has
```
jennybc, line 7:
```suggestion
#' Some of these functions are now deprecated since the tidyverse team has
```

jennybc: Should we actually deprecate these, i.e. in the code? Right now the deprecation is just a state of mind.

R/pkgdown.R
```diff
@@ -10,13 +10,14 @@
...
-#'   * `use_pkgdown_travis()` \lifecycle{soft-deprecated} helps you set up
-#'     pkgdown for automatic deployment on Travis-CI. This is soft-deprecated,
-#'     as the tidyverse team has shifted away from Travis-CI and towards GitHub
-#'     Actions. `use_pkgdown_travis()` creates an empty `gh-pages` branch for
-#'     the site and prompts about next steps regarding deployment keys and
-#'     updating your `.travis.yml`. Requires that the current user can push to
-#'     the primary repo, which must be configured as the `origin` remote.
+#'   * `use_pkgdown_travis()` `r lifecycle::badge("deprecated")` helps you
+#'     set up pkgdown for automatic deployment on Travis-CI. This is
+#'     soft-deprecated, as the tidyverse team has shifted away from Travis-CI
```
jennybc, line 15:
```suggestion
#'     deprecated, as the tidyverse team has shifted away from Travis-CI
```

jennybc (approved):

R/ci.R
hadley: Soft deprecation is still a thing; it just doesn't get it's own badge; and it looks like `use_travis()` is already calling `deprecate_soft()`?

hadley:
Merged this to avoid future merge conflicts. I can do a separate review of calls to deprecation functions if you want.

PR #1303, opened 2020-12-04
Remove badges we rarely use for functions

hadley:


lionel- (approved):

jennybc:
This creates some self-contradiction within usethis itself. I give some examples:

The CI functions for travis and appveyor are labelled "soft-deprecated":

https://github.com/r-lib/usethis/blob/61730df6044567cb5a46eeae996d981f0bcdf338/R/ci.R#L5-L11

I guess these become "questioning"?


We have been referring to certain functions and function arguments as "defunct":

https://github.com/r-lib/usethis/blob/12993dbe855736932d464ad320c76f81b7f232b1/R/github.R#L37

https://github.com/r-lib/usethis/blob/12993dbe855736932d464ad320c76f81b7f232b1/R/release.R#L119

(and many more)

I guess these become "deprecated"? Do "deprecated" and "defunct" mean the same thing?

jennybc:
Yeah, reading here, it seems like we just sort of invalidated or reduced access to several stages, some of which are being used in usethis itself:

https://lifecycle.r-lib.org/articles/lifecycle.html#stages

Will these stages also be reworked / removed in the lifecycle package?

lionel-:
Given latest discussions, there is a single "deprecation" stage with three modes of communications: soft warnings, warnings, and errors. This corresponds to the `deprecate_` functions in lifecycle.

From the user point of view, soft-deprecation and deprecation are the same thing. They signal a function that should no longer be used. Soft-deprecated functions should call `lifecycle::deprecate_soft()` but use the "deprecated" badge. I guess I should make that change in #1302.

hadley:
@jennybc we're working on more details here: https://github.com/r-lib/lifecycle/pull/63

PR #1313, opened 2020-12-18
Make UI calls more consistent in PR functions

malcolmbarrett:
There are several `ui_*()` calls that feel not-quite-right to me in `pr.R`; this PR swaps out several existing calls with other UI functions that feel more appropriate given the message.


malcolmbarrett:
These R CMD check errors seem to have nothing to do with my innocent little PR 😇

jennybc:
I feel like I had some weak reason for a couple of these, but I don't feel strongly. And most of these do seem to be clear improvements. Happy to merge. Thanks.

PR #1319, opened 2020-12-23
Edit template

malcolmbarrett:
This PR adds `edit_template()` for creating and editing files in `inst/templates`. I've resisted this PR because it's such a simple function, but I find myself constantly wanting to have it available. I'm conscious of the fact that, as the package matures, these little helper functions may not find their forever home in usethis. Nevertheless, here's why I think usethis should have this function:

1. `use_template()` introduces a structure, much in the way that R Markdown templates do. That structure should be automated away because if you misspecify it, templates will not work. `use_rmarkdown_template()` is important not just because of the complexity of that structure but the consistency. usethis templates use a considerably simpler structure, yet--and I'm not kidding--I can *never* remember if it's `inst/template` or `inst/templates`. I literally have to look it up *every single time*. 
2. Templates are a key asset of usethis when you're using the package to develop internal tools. Templates are, in fact, probably the most useful feature of the ecosystem when you're building a usethis-like package. Iterating templates should thus have good support.


jennybc:
I recently added a feature to `pr_fetch()` and `pr_resume()` that, when called argument-less, it gives you some choices.

I suspect something similar might be useful here, because who is going to remember the exact names of their existing templates? So if `template` is not provided, could say something like "looking up existing templates in `inst/templates ...` and provide a list.

malcolmbarrett:
Ooo that's a great idea, thank you! I'll try to add that today. I've greatly enjoyed that functionality for the pr functions so far for exactly the same reason

malcolmbarrett:
Ok, `edit_template()` now uses the same `menu()` approach as the PR functions, more or less. I'm excited about that!

PS I noticed that Hadley's change to the tests resulted in just calling `edit_template()` without `capture.output()`. I modeled that test after those for `edit_file()`; happy to do a small (separate) PR to make those tests consistent if `capture.output()` is not needed, if you like.

R/edit.R
```diff
@@ -40,12 +40,43 @@ edit_file <- function(path, open = rlang::is_interactive()) {
...
+  }
+
+  choice <- utils::menu(
+    choices = templates,
+    title = "Which template do you want to edit? (0 to exit)"
+  )
+
+  if (choice == 0) {
+    return(invisible())
+  }
```
jennybc, line 75: I think this case sort of takes care of itself, i.e. `templates[0]` will be `character()`, which seems OK, yeah?

R/edit.R
```diff
@@ -40,12 +40,43 @@ edit_file <- function(path, open = rlang::is_interactive()) {
...
   path <- proj_path("inst", "templates", template)
   edit_file(path, open)
 }
 
+choose_template <- function() {
+  if (!is_interactive()) {
+    return(character())
+  }
+  templates <- path_file(dir_ls("inst/templates", type = "file"))
+  if (purrr::is_empty(templates)) {
```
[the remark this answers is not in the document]
malcolmbarrett: oh, I forgot its true home is in rlang. Will do

R/edit.R
malcolmbarrett: Good point!

jennybc:
Thanks!

malcolmbarrett:
Awesome, thanks to you both!

Issue #1322, opened 2021-01-07
urlchecker should come earlier in release issue


hadley:
You probably want to run it before any check runs

PR #1324, opened 2021-01-08
Improve `use_release_issue()`

malcolmbarrett:
I recently had a confusing encounter with `use_release_issue()`. Firstly, I had already used `use_version()` to increment, and the menu option did not provide me a way to say "use the current version". Secondly, the menu prompt made it sounds like `use_release_issue()` was going to increment my version (which it didn't, of course; it just told me to use `use_version()` in the checklist), resulting in me changing my version then changing it back when I realized nothing happened. 

This PR makes a few improvements to the menu encounter and resulting actions:

1. It adds an option to use the current version. This is only a menu option for `use_release_issue()` and not `use_version()`, since the latter doesn't make sense.
2. It clarifies the menu message a bit to ask for what the release version *will* be. 
3. In cases where you already have the release version set, it does not tell you to use `use_version()` in the resulting checklist.
4. Some bookkeeping with the related tests + adding a test for when you are using the same version as the current version.

If you think this PR merits a merge, I'd also be happy to address #1322, if you like, since it's related to this function.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 `use_readme_rmd()`, `use_readme_md()`, `use_tidy_contributing()`, and `use_tidy_support()` use updated logic for determining the `OWNER/REPO` spec of the target repo (#1312).
 
+`use_release_issue()` now allows you to specify the current release and, if so, does not instruct you to `use_version()` in the resulting checklist (@malcolmbarrett, #1324)
+
 # usethis 2.0.0
 
 ## Adoption of gert and changes to Git/GitHub credential handling
--- a/R/release.R
+++ b/R/release.R
@@ -25,7 +25,7 @@ use_release_issue <- function(version = NULL) {
     }
   }
 
-  version <- version %||% choose_version()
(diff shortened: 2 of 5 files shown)


hadley:
Hmmmm, I'm not sure sure about allowing a release with the current version, because I'm not sure it's a good idea — I think you want to minimise any discrepancies between the release CRAN version and any github commits with the same version number, and setting the package version early on increases the risk of pushing the version to github.

hadley:
I think this is mostly just uncertainty you'll face using `use_release_issue()` for the first time, so for the reason described I'd prefer not to add to usethis. Thanks for thinking about this little weird corner of usethis 😄

malcolmbarrett:
Thanks! I get that perspective. FWIW, the reason I've tended to bump the version earlier is because of the note generated by winbuilder and r-hub re: large version number. I've found that I've missed things a couple of times thinking it was the same ol' note. But that's my problem, not usethis's 😅 

Anyway, I may submit a different PR that just clarifies the messages (since I found them a bit confusing) without touching the version stuff

PR #1334, opened 2021-01-20
use_pipe() now documents arguments and return value

### LINKED ISSUE #1333: use_pipe() doesn't include \arguments and  \value in Rd page
With a newly submitted package that utilizes `usethis::use_pipe()`, I got this from CRAN:

```
Please add \value to .Rd files regarding exported methods and explain
the functions results in the documentation. Please write about the
structure of the output (class) and also what the output means. (If a
function does not return a value, please document that too, e.g.
\value{No return value, called for side effects} or similar)
Missing Rd-tags:
      pipe.Rd: \arguments,  \value
```

cpsievert:
Closes #1333

### DIFF
--- a/inst/templates/pipe.R
+++ b/inst/templates/pipe.R
@@ -10,5 +10,5 @@
 #' @usage lhs \%>\% rhs
 #' @param lhs A value or the magrittr placeholder.
 #' @param rhs A function call using the magrittr semantics.
-#' @returns the result of calling `rhs(lhs)`.
+#' @return The result of calling `rhs(lhs)`.
 NULL
--- a/inst/templates/pipe.R
+++ b/inst/templates/pipe.R
@@ -8,4 +8,7 @@
 #' @export
 #' @importFrom magrittr %>%
 #' @usage lhs \%>\% rhs
+#' @param lhs A value or the magrittr placeholder.
+#' @param rhs A function call using the magrittr semantics.
+#' @returns the result of calling `rhs(lhs)`.
 NULL


inst/templates/pipe.R
```diff
@@ -8,4 +8,7 @@
 #' @export
 #' @importFrom magrittr %>%
 #' @usage lhs \%>\% rhs
+#' @param lhs A value or the magrittr placeholder.
+#' @param rhs A function call using the magrittr semantics.
+#' @returns the result of calling `rhs(lhs)`.
```
jennybc, line 13: Morally, it feels like this should be:

```
#' @inheritParams magrittr::`%>%`
#' @inherit magrittr::`%>%` return
```

and yet I believe it can't really be so, because here we are just documenting `NULL`, so at `document()`-time we see:

```
Warning message:
Topic '%>%': no parameters to inherit with @inheritParams 
```

And no return value is documented because `` magrittr::`%>%` `` itself does not technically document a return value (https://magrittr.tidyverse.org/reference/pipe.html).

So maybe this is the best plan? @Hadley do you have an opinion?

jennybc, line 13:
```suggestion
#' @param lhs A value or the magrittr placeholder.
#' @param rhs A function call using the magrittr semantics.
#' @return The result of calling `rhs(lhs)`.
```

jennybc:
Just for ambience, here's the current documentation for the base pipe in R-devel:

https://stat.ethz.ch/R-manual/R-devel/library/base/html/pipeOp.html

inst/templates/pipe.R
hadley: Yeah, I think this approach is fine. I don't think many people will actually read this documentation, so this is just about satisfying CRAN.

jennybc:
Thanks for fixing it for everyone @cpsievert!

PR #1346, opened 2021-02-04
Use the pak based actions for tidy github actions

jimhester:


jennybc:
Shall we switch usethis itself over at the same time? Right now it looks like our workflows are a mix of pak and remotes.

.github/workflows/pkgdown.yaml
```diff
@@ -8,46 +8,50 @@ name: pkgdown
 
 jobs:
   pkgdown:
-    runs-on: macOS-latest
+    runs-on: ubuntu-18.04
```
jimhester, line 11: The new workflows use ubuntu for the pkgdown and test coverage workflows, hopefully that is Ok for the usethis case.

jennybc:
Thanks!

PR #1348, opened 2021-02-05
Minor improvement to installed package checks

hadley:


NEWS.md
```diff
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* All functions that require a package now ask you if you'd liked to install it.
```
jennybc, line 3:
```suggestion
* All functions that require a package now ask you if you'd like to install it.
```

R/roxygen.R
```diff
@@ -21,6 +21,8 @@ use_roxygen_md <- function() {
     if (!uses_git()) {
       ui_todo("Use git to ensure that you don't lose any data")
     }
+
+    check_installed("roxygen2md")
```
jennybc, line 25: Just below ... does "existing Rd markdown" even make sense? Feels like we could word that better.

jennybc (approved):

R/roxygen.R
hadley: Yeah, good catch.

PR #1350, opened 2021-02-05
Rename use_tidy_pkgdown() to use_pkgdown_github_pages()

### LINKED ISSUE #1308: Does use_tidy_pkgdown() have the wrong name?
Discussion from review of usethis v2.0.0 blog post:

@hadley 
Probably shouldn't promote `use_tidy_pkgdown()` because it's our workflow?
 
 @jennybc 
In this case, `use_tidy_pkgdown()` is both our workflow AND what I would recommend for other maintainers as well. `use_tidy_pkgdown()` is `use_pkgdown() + use_github_pages() + use_github_action("pkgdown") + use_pkgdown_url()`.

Not sure what to do about that, if anything. I see your point, i.e. where other "tidy" functions often do something that is overkill or inappropriate for others. But it's not the case here.
 
 @jennybc
In any case, it's not really a blog post problem but potentially a function naming problem.
 
 @hadley 
Yeah, feels like we might've given the wrong name. Maybe should be `use_github_pkgdown()` or similar?

hadley:
Fixes #1308


R/github-pages.R
```diff
@@ -46,7 +46,7 @@
 #' @examples
 #' \dontrun{
 #' use_github_pages()
-#' use_github_pages(branch = git_branch_default(), path = "/docs")
+#' use_github_pages(branch = git_branch_default(), path = "/docs")p
```
jennybc, line 49:
```suggestion
#' use_github_pages(branch = git_branch_default(), path = "/docs")
```

R/pkgdown.R
```diff
@@ -1,22 +1,26 @@
...
-#'     updating your `.travis.yml`. Requires that the current user can push to
-#'     the primary repo, which must be configured as the `origin` remote.
+#' a beautiful website. usethis provides two functions help you use pkgdown:
+#'
+#' * `use_pkgdown()`: creates a pkgdown config file, adds relevant files or
+#'   directories to `.Rbuildignore` and `.gitignore`, and builds favicons if
+#'   your package has a logo.
+#'
+#' * `use_pkgdown_github_pages()`: implements the GitHub setup needed to
+#'   automatically publishes you package to GitHub pages:
```
jennybc, line 12:
```suggestion
#'   automatically publish your pkgdown site to GitHub pages:
```

jennybc (approved):

PR #1351, opened 2021-02-05
Provide pointer to GHA docs

### LINKED ISSUE #1307: feature request: message with link to documentation when using a `use_github_action()` example
It would be very helpful to get a message with a link to this README and encouragement to read it when using one of the `use_github_action()` examples:

https://github.com/r-lib/actions/blob/master/examples/README.md

It's easy to miss this info with **usethis** since you bypass the https://github.com/r-lib/actions repo (though I'll confess that I missed it anyway since it didn't occur to me to look for an additional README in a subdirectory). This one has some crucial information, in particular, that it's necessary to run `renv::snapshot()` locally and commit the `renv.lock` in order for the workflow to work.

hadley:
Fixes #1307

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -143,8 +143,10 @@ use_github_action <- function(name,
     url <- glue(
       "https://raw.githubusercontent.com/r-lib/actions/master/examples/{name}"
     )
+    readme <- "https://github.com/r-lib/actions/blob/master/examples/README.md"
   } else {
     stopifnot(is_string(url))
+    readme <- NULL
   }
 
   if (is.null(save_as)) {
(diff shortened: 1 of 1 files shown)


R/github-actions.R
```diff
@@ -163,6 +165,9 @@ use_github_action <- function(name,
   if (open && new) {
     edit_file(proj_path(save_as))
   }
+  if (!is.null(readme)) {
+    ui_todo("Learn more at <{readme}>")
```
jennybc, line 169: I feel like we might have some other convention going with URLs but I can't search around for that atm.

jennybc (approved):

R/github-actions.R
hadley: I looked for `ui_url()` but didn't see anything. We can do a thorough pass if/when the IDE gets support for turning these into "real" links.

PR #1354, opened 2021-02-07
Make create_local_thing() quieter

hadley:
I don't know how you feel about this, but I think this makes running individual tests substantially more pleasant. I'm happy to trust that `create_local_package()` does the right thing; I don't need to be informed of its every action.

### DIFF
--- a/tests/testthat/helper.R
+++ b/tests/testthat/helper.R
@@ -34,35 +34,23 @@ create_local_thing <- function(dir = file_temp(pattern = pattern),
     ui_stop("Target {ui_code('dir')} {ui_path(dir)} already exists.")
   }
 
+  withr::local_options(usethis.quiet = TRUE)
+
   old_project <- proj_get_() # this could be `NULL`, i.e. no active project
   old_wd <- getwd()          # not necessarily same as `old_project`
 
-  withr::defer(
-    {
-      ui_done("Deleting temporary project: {ui_path(dir)}")
-      fs::dir_delete(dir)
-    },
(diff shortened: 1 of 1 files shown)


tests/testthat/helper.R
```diff
@@ -34,35 +34,23 @@ create_local_thing <- function(dir = file_temp(pattern = pattern),
...
+  switch(
+    thing,
+    package = create_package(dir, rstudio = rstudio, open = FALSE, check_name = FALSE),
+    project = create_project(dir, rstudio = rstudio, open = FALSE)
   )
+  withr::defer(fs::dir_delete(dir), envir = env)
 
-  withr::defer(proj_set(old_project, force = TRUE), envir = env)
   proj_set(dir)
+  withr::defer(ui_silence(proj_set(old_project, force = TRUE)), envir = env)
```
hadley, line 50: This still needs `ui_silence()` since it's not run inside this function.

jennybc:
Ha! I had added this fairly recently, with intention (e.g. 5475c461421). I think in some dev contexts, I've had experiences where I end up in a weird situation (maybe when I'm context switching a lot) and I appreciated more noisiness.

How much does it bother you?

hadley:
A mild to moderate amount. It makes running the tests individually very noisy, which made me think that was something wrong.

jennybc:
Maybe it should detect whose machine we're on (Jenny or Hadley) and be noisy vs. quiet accordingly 😁

PR #1356, opened 2021-02-08
Add use_stopwords()

kbenoit:
Add `use_stopwords()`, so that this function can be relocated from the [**stopwords** package](https://github.com/quanteda/stopwords).

We added `use_stopwords()` in v2.0, but this introduced too many downstream dependencies via importing **usethis**. It's a very popular package and it would be great to keep the `use_stopwords()`, but it is more logical and efficient to locate this function in the **stopwords** package.


hadley:
I think this is a bit too specific for usethis. Can we instead help you keep it in stopwords, but with fewer dependencies? (i.e. by moving it to suggests).

jennybc:
For more context, we are also now questioning `use_tibble()` and `use_data_table()`, because once you start adding these sort of helpers, there is no end of it.

kbenoit:
No problem. I'd added it this way https://github.com/quanteda/stopwords/blob/master/R/use_stopwords.R which would not work with usethis in Suggests, but I'm all ears to alternatives. Especially if you are going to issue recommendations for `use_tibble()` and `use_data_table()`, this would be a very useful set of guidelines to wean the package from these and future helpers of this sort.

hadley:
The technique you use there is potential very fragile because the copying happens at package build time, not load time. That means there’s the potential for a mismatch because the installed version of use this and the version of use this that the embedded functions expect.

kbenoit:
@Hadley good point, but what alternative to that sort of code smuggling exists when the source package has banned exports?

PR #1359, opened 2021-02-09
Clarify `choose_version()` prompt for `use_release_issue()`

malcolmbarrett:
This is a drastically simplified version of #1324. As mentioned there, I was confused by the prompt in `use_release_issue()`, identical to the prompt in `use_version()`, which made it sound like it was going to increment for me then and there.

This PR simply clarifies the message and doesn't touch any of the other issues mentioned in 1324. Unfortunately, I couldn't think of wording that worked for both cases that didn't make it sound like `use_release_issue()` was going to increment then and there, so I had to introduce a few more lines of code than I would have liked for such a simple PR.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -25,7 +25,7 @@ use_release_issue <- function(version = NULL) {
     }
   }
 
-  version <- version %||% choose_version()
+  version <- version %||% choose_version(release = TRUE)
   if (is.null(version)) {
     return(invisible(FALSE))
   }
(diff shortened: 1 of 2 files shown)


hadley:
Thanks! I think it's slightly better to provide a `message` argument, and because I want to submit to CRAN momentarily, I just made the fix.

malcolmbarrett:
Thanks! I agree that's better so np

Issue #1360, opened 2021-02-09
Lifecycle badge URL rejected by CRAN


MattCowgill:
I am the maintainer of a CRAN package (`{readabs}`) that uses a Tidyverse lifecycle badge in README.md. 

On submitting an update to CRAN I received the following: 

```
Thanks, we see:

   Found the following (possibly) invalid URLs:
     URL: https://www.tidyverse.org/lifecycle/#stable (moved to 
https://lifecycle.r-lib.org/articles/stages.html)
       From: README.md
       Status: 200
       Message: OK

Please change http --> https, add trailing slashes, or follow moved 
content as appropriate.

Please fix and resubmit.
```

The tidyverse.org URL is added when users invoke `usethis::use_lifecycle_badge()`. It would be good to ensure that the badge points to the lifecycle.r-lib.org address without redirects, to avoid CRAN rejections.

MattCowgill:
PS thanks for a great package!

jennybc:
I believe this to be addressed in the dev version already:

23300fcb854a14d50007c20ef6f9dc5038dd33af

30bc851a103dfec03ecac0cb5e419090b4c4038f

MattCowgill:
Ah, thank you! Sorry, I didn't realise.

jennybc:
A small usethis release is coming soon

PR #1349, opened 2021-02-05
Updates for latest lifecycle workflow

hadley:
Fixes #1323

PR #1370, opened 2021-02-18
Change badge URL to ease CRAN submissions

jdtrat:
Hello! For a recent package submission, CRAN requested I change the url "https://www.tidyverse.org/lifecycle/#stable" to "https://lifecycle.r-lib.org/articles/stages.html". If this is common to others, maybe it should be the default?


jennybc:
Thanks, this has already been fixed (recently). I imagine you need to update to the released version of usethis.

https://github.com/r-lib/usethis/pull/1349

https://github.com/r-lib/usethis/issues/1360

PR #1376, opened 2021-02-24
Make use_tidy_github_actions() more pushy

### LINKED ISSUE #1367: use_tidy_github_actions() should be more forceful
I don't think it needs to prompt to overwrite because (a) we know that the user must be using git + github and (b) the user are mostly us. Should also avoid the repeated advice to see the readme.

hadley:
Since we (presumably) know what we're doing. Fixes #1367.

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -85,10 +85,10 @@ use_tidy_github_actions <- function() {
 
   use_coverage(repo_spec = repo_spec)
 
-  full_status <- use_github_action_check_full(repo_spec = repo_spec)
-  pr_status   <- use_github_action_pr_commands()
-  pkgdown_status <- use_github_action("pkgdown-pak", save_as = "pkgdown.yaml")
-  test_coverage_status <- use_github_action("test-coverage-pak", save_as = "test-coverage.yaml")
+  full_status <- use_github_action_check_full(repo_spec = repo_spec, overwrite = TRUE)
+  pr_status   <- use_github_action_pr_commands(overwrite = TRUE)
+  pkgdown_status <- use_github_action("pkgdown-pak", save_as = "pkgdown.yaml", overwrite = TRUE)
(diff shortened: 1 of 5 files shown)


R/github-actions.R
```diff
@@ -160,12 +161,12 @@ use_github_action <- function(name,
...
   create_directory(path_dir(proj_path(save_as)))
 
-  new <- write_over(proj_path(save_as), contents)
+  new <- write_over(proj_path(save_as), contents, overwrite = overwrite)
 
   if (open && new) {
     edit_file(proj_path(save_as))
   }
-  if (!is.null(readme)) {
+  if (!is.null(readme) && !overwrite) {
```
hadley, line 169: This is not 100% correct but is expedient, and I don't think another argument is worth it.

jimhester (approved):

R/write.R
```diff
@@ -69,16 +69,17 @@ write_union <- function(path, lines, quiet = FALSE) {
 #'   necessary or overwriting existing, if proposed contents are not identical
 #'   and user is available to give permission.
 #' @param contents Character vector of lines.
+#' @param overwrite Unconditionally an existing file.
```
jennybc, line 72: There seem to be some missing words here

jennybc (approved):
Feels like this suggests a tweak or addition to the `write_over()` tests

So: when `overwrite = FALSE`, user will be asked if they are OK with replacing the existing workflow?

I suspect we could be pushy more globally, i.e. whenever we see the project `uses_git()`.

hadley:
Hmmmm, I like the idea of suppressing this question if you already use git. What do you think about just doing that globally?

jennybc:
> Hmmmm, I like the idea of suppressing this question if you already use git. What do you think about just doing that globally?

I'm on board.

hadley:
Closing in favour of adjusting `write_over()`

PR #1377, opened 2021-02-26
Add `use_import_from()`

### LINKED ISSUE #1361: Generalizing `use_tibble()` and `use_data_table()`
`use_tibble()` and `use_data_table()` were marked as questioning in #1046 because:

> [they] don't do that much, and [they suggest] hundreds of other similar functions that might be added to usethis

Which is think is reasonable. I'm opening this issue to discuss if a generalized version of these functions is a good fit for usethis. I'm thinking something like `use_import(package, function = NULL)`. So, these functions would simply be replaced by these actions: `use_import("tibble", "tibble")` and `use_import("data.table")` or something to that effect. This generalization would primarily address the second reason above, but with respect to the first reason, I think usethis brings a consistency to roxygen imports that is very useful.

The reason I'm interested in this workflow is because, even though usethis is just adding a little roxygen, I like that it adds it to a consistent place and automates away that little bit of tedium, much like `use_package()`. Putting it in the package doc if it exists or using `ui_code_block()` otherwise is very nice to me. It has the added benefit of making such imports more consistent; they can be hard to find without digging around the docs or namespace, and if usethis has an opinion about where to put them, then I can start checking in one location to start.

malcolmbarrett:
This PR adds a new function, `use_import_from()`, that adds `@importFrom` to the package doc, potentially as a general replacement for for functions like `use_tibble()`. It also adds pkgload to Suggests.

I made the judgment call to rename the proposed function to `use_import_from()` so that it was clearer that the function would not add `@import`, only `@importFrom`.

While I did not touch `use_pipe()` or `use_tibble()`, this function (or some of its internal functionality) could replace some of the boilerplate in those functions. That might be desirable if you want to start pushing `use_package_doc()` a bit more.  Notably, since this function does not facilitate `@import` uses, it cannot replicate `use_data_table()`.

Closes #1361

cc: @hadley


R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
+#' Import a function from another package
+#'
+#' `use_import_from()` imports a function from another package by adding the
+#' Roxygen tag `@importFrom` to the package-level documentation (possibly
+#' created with [`use_package_doc()`]). Importing a function from another
+#' package allows you to refer to it without a namespace (e.g., `fun()` instead
+#' of `package::fun()`). Use such imports judiciously, as they make it hard to
+#' tell where a function comes from.
```
hadley, line 8:
```suggestion
#' `use_import_from()` imports a function from another package by adding the
#' roxygen2 `@importFrom` tag to the package-level documentation (which can
#' be created with [`use_package_doc()`]). Importing a function from another
#' package allows you to refer to it without a namespace (e.g., `fun()` instead
#' of `package::fun()`). 
```

R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
...
+#'
+#' @param package Name of package to import `fun` from
+#' @param fun Name of function to import
```
hadley, line 11:
```suggestion
#' @param package Package name
#' @param fun Function name
```

Do these (should these?) allow vectors?

R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
...
+#' \dontrun{
+#' use_import_from("usethis", "ui_todo")
+#' }
+#'
+use_import_from <- function(package, fun, load = is_interactive()) {
+  check_is_package("use_import_from()")
+  check_uses_roxygen("use_import_from()")
+
+  if (!check_has_package_doc()) {
+    return(FALSE)
```
hadley, line 29:
```suggestion
    return(invisible(FALSE))
```

R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
...
+
+  if (changed && load) {
+    check_installed("pkgload")
+    pkgload::load_all()
+  }
+
+  invisible(changed)
+}
+
+import_from <- function(package, fun) {
```
hadley, line 44: I don't think this needs to be a separate function. It's neither complicated nor used in multiple places.

R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
...
+use_import_from <- function(package, fun, load = is_interactive()) {
+  check_is_package("use_import_from()")
+  check_uses_roxygen("use_import_from()")
+
+  if (!check_has_package_doc()) {
+    return(FALSE)
+  }
+
+  use_dependency(package, "Imports")
+  changed <- roxygen_ns_append(import_from(package, fun)) &&
```
hadley, line 33: `roxygen_update()` doesn't do anything; it just reminds the user. I think it's probably worth fixing the name so would mind renaming `roxygen_update()` to `roxygen_remind()` and then creating an `roxygen_update_ns()` function that does `roxygen2::roxygenise(proj_get(), "namespace")`?

R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
...
+    roxygen_update()
+
+  if (changed && load) {
+    check_installed("pkgload")
```
hadley, line 37: I think you can drop this check since roxygen2 imports pkgload so we know it's installed.

R/use_import_from.R
```diff
@@ -0,0 +1,69 @@
+#' Import a function from another package
+#'
+#' `use_import_from()` imports a function from another package by adding the
+#' Roxygen tag `@importFrom` to the package-level documentation (possibly
+#' created with [`use_package_doc()`]). Importing a function from another
+#' package allows you to refer to it without a namespace (e.g., `fun()` instead
+#' of `package::fun()`). Use such imports judiciously, as they make it hard to
```
hadley, line 7: Should mention that it re-loads the package so the newly imported function is immediately available.

jennybc:
usethis does several things that are usually followed by a call to `load_all()` (e.g. `use_data()`, `use_namespace()`).

Implicitly I've sort of thought this: "modify the source package = usethis's job, other package workflow stuff = devtools's job". So devtools is the storefront for the most common workflow functions from roxygen2, pkgload, etc.

It feels like the most self-consistent thing for usethis to do here would be to recommend running `devtools::load_all()`, the same way we recommend running `devtools::document()` in ~8 different places, versus calling it ourselves.

Obviously I don't care much about us adding one call to `load_all()`. It's more that the one-off nature of it feels odd.

R/use_import_from.R
malcolmbarrett: They don't currently (sort of modeled after use_package()) but I think `fun` should since it's common to have `@importFrom pkg x y z`

malcolmbarrett:
I think usethis has really trained me to expect that slight separation in workflows, as well. Whether or not that is intentional or the best approach, it's definitely lodged in my brain at this point!

malcolmbarrett:
Oh, `roxygen_update_ns()` adds another load. Maybe I should wait until that issue is resolved before proceeding. I do like the `roxygen_remind()` rename, tho 🙃

hadley:
@jennybc I was thinking that this would be a place for an experiment.

jennybc:
OK

malcolmbarrett:
Ok, experiment commenced! One thing of note: I tried to make the communication about updating `NAMESPACE` and loading more usethis-like because it feels weird to have other styles of messages pop in (much like it did with gert originally). It also made the tests noisier.

NAMESPACE
```diff
@@ -137,6 +137,7 @@ export(use_github_release)
 export(use_gitlab_ci)
 export(use_gpl3_license)
 export(use_gpl_license)
+export(use_import_from)
```
jennybc, line 140: Not sure where I fall on this, but it seem we should also consider `use_importfrom()`

R/use_import_from.R
```diff
@@ -0,0 +1,67 @@
...
+
+  fun <- gsub("\\(.*\\)", "", fun)
+  fun <- glue_collapse(fun, sep = " ")
+  changed <- roxygen_ns_append(glue("@importFrom {package} {fun}")) &&
+    roxygen_update_ns()
+
+  if (load) {
+    ui_done("Loading {project_name()}")
+    pkgload::load_all(quiet = TRUE)
+  }
```
hadley, line 42:
```suggestion
  changed <- roxygen_ns_append(glue("@importFrom {package} {fun}")) &&
    roxygen_update_ns()

  if (changed && load) {
    ui_done("Loading {project_name()}")
    pkgload::load_all(quiet = TRUE)
  }
```

R/use_import_from.R
```diff
@@ -0,0 +1,67 @@
...
+check_has_package_doc <- function() {
+  if (has_package_doc()) {
+    return(invisible(TRUE))
+  }
+
+  if (!is_interactive()) {
+    return(invisible(FALSE))
+  }
+
+  add_pkg_doc <- ui_yeah("{ui_code('use_import_from()')} requires \\
```
hadley, line 56: Can you please follow https://style.tidyverse.org/syntax.html#long-lines here?

hadley: @jennybc might have other suggestions on how we usually format long strings in usethis

hadley (commented):
Looking good — thanks for your work on this!

R/use_import_from.R
jennybc: I have developed a certain internal style for the `ui_*()` functions (re: line breaks in source vs. line breaks on the Console) that probably horrifies @Hadley.

But I have actually reached this convention through some thought and experimentation. So you can look around in general, but here are some `ui_yeah()` examples:

https://github.com/r-lib/usethis/blob/db18e70c718b476bd074b5893e26aeb6e7427a0a/R/package.R#L96-L101

https://github.com/r-lib/usethis/blob/db18e70c718b476bd074b5893e26aeb6e7427a0a/R/pkgdown.R#L121-L127

jennybc:
This should get a NEWS bullet.

R/use_import_from.R
malcolmbarrett: :o

malcolmbarrett:
Ok, all updated! Thanks for iterating with me. Let me know if there's anything else!

hadley:
I started playing around with this. I started by adding a check to see if the function actually existed, and then accidentally removed the line pasting `funs` together.  But that made me wonder if one line per `@importFrom` is actually a better solution?

hadley:
I've left a failing test to remind me to come back to this — `block_append()` probably needs to `unique()` what it writes out and should at least have an option to sort the output.

malcolmbarrett:
Hm! I noticed that my original code did one line per element (unintentionally), and it wasn't too bad. I like that roxygen can put many imports on the same line, but in the wild it *does* get unwieldy quite quickly. 

Now that you're saying that, it made me think that I ought to add a check that `package` is length 1 to avoid an unclear error message from glue.

hadley:
I think one line per element is the right approach to avoid accidental duplication.

malcolmbarrett:
Ok, I updated the last test to match that output

hadley:
@jennybc I think it's worth you having another look — I've boosted the abilities of `block_append()` so that we create a more pleasant writing experience. I also removed the warnings from the template; I don't think they're that important since there's no real harm if you remove the comments, and it's certainly fine to edit the `@importFrom` statements.

@malcolmbarrett thanks for working on this!

R/use_import_from.R
```diff
@@ -0,0 +1,79 @@
...
+#' @return
+#' Invisibly, `TRUE` if the package document has changed, `FALSE` if not.
+#' @export
+#' @examples
+#' \dontrun{
+#' use_import_from("usethis", "ui_todo")
+#' }
+use_import_from <- function(package, fun, load = is_interactive()) {
+  if (!is_string(package)) {
+    ui_stop("`package` must be a single string")
```
jennybc, line 26:
```suggestion
    ui_stop("{ui_code('package')} must be a single string")
```

tests/testthat/test-block.R
```diff
@@ -0,0 +1,17 @@
+test_that("block_append() only writes unique lines", {
+  path <- tempfile()
```
jennybc, line 2: Elsewhere in the tests, we use `withr::local_tempfile()` for this purpose, which kills 2 or 3 birds with 1 stone.

tests/testthat/test-block.R
```diff
@@ -0,0 +1,17 @@
...
+  writeLines(block_create(), path)
+
+  block_append("---", c("x", "y"), path)
+  block_append("---", c("y", "x"), path)
+  expect_equal(block_show(path), c("x", "y"))
+})
+
+test_that("block_append() can sort, if requested", {
+  path <- tempfile()
```
jennybc, line 11: Ditto the above

malcolmbarrett:
Really like the addition of sort + unique, esp that it's the default for this use case.

hadley:
I don't think we need to make `sort` an argument — if you don't want it sorted, you can use some other approach 😄

malcolmbarrett:
💯 

Thanks for all the helpful feedback and contributions to this PR @hadley and @jennybc!

jennybc:
Should we now open an issue re: revisiting existing functions like `use_pipe()` or `use_tibble()` and route them through this code or deprecate them in favour of a call to this code, as appropriate?

hadley:
Looks like we could use it in `use_tibble()` but not in `use_pipe()` or `use_data_table()` which work somewhat differently.

malcolmbarrett:
I think it can supplement `use_pipe()` here: https://github.com/r-lib/usethis/blob/4819d19a1975aaee193018e4713c305dfb7bfb1e/R/pipe.R#L32-L35

I can file an issue and work on a corresponding PR

PR #1384, opened 2021-03-05
Don't instruct usage if no changes in `use_package()`

malcolmbarrett:
If I do `use_package("pre.existing.import", "Suggests")`, `use_dependency()` warns me that no change will be made. However, `how_to_use()` doesn't know about this, so it still tells you how to use the given type. 

This PR prevents `use_package()` and `use_dev_package()` from calling `how_to_use()` if no change has been made.

I thought to add a test for this little bug, but I don't know how to express the expectation "A warning, followed by no messages". Maybe just a snapshot test?

PS I actually don't understand the logic here, TBH. Why does usethis care if I'm shuffling dependencies around in the first place? And why does it let me move up the dependency ladder but not down?


R/package.R
```diff
@@ -33,8 +33,8 @@ use_package <- function(package, type = "Imports", min_version = NULL) {
     refuse_package(package, verboten = c("tidyverse", "tidymodels"))
   }
 
-  use_dependency(package, type, min_version = min_version)
-  how_to_use(package, type)
+  changed <- use_dependency(package, type, min_version = min_version)
+  if (changed) how_to_use(package, type)
```
hadley, line 37: Can you please follow tidyverse style here and use `{}`?

hadley (commented):
A snapshot test would be perfect I think. And add a news bullet?

R/package.R
malcolmbarrett: Oh, oops, I thought that was just for `return()` and friends! Updating now

malcolmbarrett:
Ok, style updated, test added, and news updated!

tests/testthat/test-package.R
```diff
@@ -4,6 +4,13 @@ test_that("use_package() won't facilitate dependency on tidyverse/tidymodels", {
   expect_usethis_error(use_package("tidymodels"), "rarely a good idea")
 })
 
+test_that("use_package() guides new packages but not pre-existing ones", {
+  create_local_package()
+  withr::local_options(usethis.quiet = FALSE)
+  expect_snapshot(use_package("withr"))
+  expect_snapshot(use_package("withr", "Suggests"))
```
hadley, line 11:
```suggestion
  expect_snapshot({
    use_package("withr")
    use_package("withr")
    use_package("withr", "Suggests")
  })
```

Just so that you see all three paths

jennybc:
> PS I actually don't understand the logic here, TBH. Why does usethis care if I'm shuffling dependencies around in the first place? And why does it let me move up the dependency ladder but not down?

Just to provide some history, `use_dependency()` is sort of a nightmare to work on, i.e. deserving of a comment "you think you will be able to make this much better, but that's what the last person thought" 😅

malcolmbarrett:
> "you think you will be able to make this much better, but that's what the last person thought"

🤣 ok, I'll always use `ten_foot_pole()` when dealing with `use_dependency()`

hadley:
Thanks!

Issue #1223, opened 2020-09-17
Don't import rematch2


jennybc:
This is only worth doing (inlining rematch2) if this is the only reason usethis depends on tibble.

jennybc:
Apparently I've removed rematch2 once before. Recording for posterity, so hopefully I don't bring it back in a 3rd time.

https://github.com/r-lib/usethis/commit/a07c957e3f9245a2f6f8a1c4fbb57c454fc3f1aa

Issue #1046, opened 2020-03-14
Mark use_tibble() as questioning


hadley:
Because it doesn't do that much, and it suggests hundreds of other similar functions that might be added to usethis.

charliejhadley:
I made the initial PR for `use_tibble()` as I was working on data packages where I wanted the package to guarantee a tibble was returned instead of a plain data.frame, and I think it's just as important as its sibling `use_data_table()`.

Unfortunately, but unsurprisingly folks have not often mentioned `use_tibble()` directly in their commit messages so its difficult for me to make a case for it being used widely. However, here are a few places where I can show it had value to users:

https://sharla.party/post/usethis-for-reporting/
https://rstudio4edu.github.io/rstudio4edu-book/data-pkg.html
https://www.hvitfeldt.me/blog/usethis-workflow-for-package-development/

The current workings of `use_tibble()` are far from my initial PR and if it is polluting the namespace that's horrible, but I feel there is a need for this functionality in `{usethis}`

hadley:
I'm not saying it's not useful, more that I don't think it's a good fit for usethis. It also goes against our usual recommendation to use `tibble::as_tibble()`.  `use_data_table()` was added primarily to match `use_tibble()` so I don't think its existence is a good argument for `use_tibble()`.

`use_pipe()` is a different beast because it's primarily about _re-exporting_ the pipe, not about simply using it.

charliejhadley:
Ah, I did employ reverse causality for the data.table function. Thanks as always for explaining your thought process, much appreciated 🙂

malcolmbarrett:
Do you see a more generalized form as an option? Something like `use_import(pkg, function = NULL)` (not a great name but you get the drift). I completely see the argument that it's not doing much (roxygen is doing the heavy lifting here and is not that complicated). **But** I feel like usethis does bring something to the table here: a standardized opinion on where to put these types of imports. Putting them in the package doc (if it exists) is very nice to me, and I often wish it was easier to find import statements like these when looking at package code (the namespace file usually being so crowded). I like that this automation results in a sensible place to check. 

That all said, it wouldn't be a huge loss. I like both `use_tibble()` and `use_data_table()` but your arguments make sense (in fact, I think I may have prompted this issue after proposing a new `use_*()` function in this style!). Just using roxygen instead is not too rough

jennybc:
cc / fyi @Michael Chirico

I'm marking both `use_tibble()` and `use_data_table()` as "questioning".

[...]

PR #1396, opened 2021-03-17
Use `use_import_from()` in other importing functions

### LINKED ISSUE #1382: Add `use_import_from()` to other functions
Definitely `use_tibble()`, maybe `use_pipe()`, but not `use_data_table()`

malcolmbarrett:
This PR modifies `use_pipe()`, `use_tibble()`, and `use_data_table()` to call `use_import_from()` internally. This extends `use_import_from()`'s stronger requirement that package docs are used for imports and the idea that we can rewrite the `NAMESPACE` while we're at it.

This PR also overhauls `use_data_table()` to simplify the import declaration, as discussed in #1382, as well as to address a few UI and testing elements of that function. ~~Because writing `NAMESPACE` requires loading which requires data.table, I also decided just to skip the tests if data.table is not installed. I tried to hack away at it via mocking, but it didn't seem worth it (or particularly robust in this case). I'm happy to hear other suggestions, though.~~ I ended up being able to mock this, so it is not skipped after all.

Closes #1382


tests/testthat/test-data-table.R
```diff
@@ -5,11 +5,20 @@ test_that("use_data_table() requires a package", {
...
+  package_doc <- read_utf8(proj_path(package_doc_path()))
+
+  purrr::walk(
+    c("data.table", ":=", ".SD", ".BY", ".N", ".I", ".GRP", ".NGRP", ".EACHI"),
+    ~ expect_match(
+      package_doc,
+      glue("#' @importFrom data.table {.x}"),
+      all = FALSE
+    )
+  )
```
malcolmbarrett, line 23: Is there a better way to write a test like this?

R/use_import_from.R
```diff
@@ -49,6 +49,7 @@ use_import_from <- function(package, fun, load = is_interactive()) {
 }
 
 check_fun_exists <- function(package, fun) {
+  check_installed(package)
```
malcolmbarrett, line 52: maybe this should go prior to using `check_fun_exists()`, since it will unnecessarily call `check_installed()` multiple times if `length(fun) > 1`

R/pipe.R
```diff
@@ -22,26 +22,11 @@ use_pipe <- function(export = TRUE) {
   check_is_package("use_pipe()")
   check_uses_roxygen("use_pipe()")
 
-  use_dependency("magrittr", "Imports")
-
   if (export) {
+    use_dependency("magrittr", "Imports")
```
hadley, line 26: Shouldn't magrittr always be added to imports?

R/tibble.R
```diff
@@ -31,9 +31,10 @@ use_tibble <- function() {
   check_is_package("use_tibble()")
   check_uses_roxygen("use_tibble()")
 
-  use_dependency("tibble", "Imports")
-  roxygen_ns_append("@importFrom tibble tibble") && roxygen_remind()
+  created <- use_import_from("tibble", "tibble")
```
hadley, line 34: Should we rebuild the NAMESPACE?

tests/testthat/test-data-table.R
```diff
@@ -1,15 +1,50 @@
+skip_if_not_installed("data.table")
```
hadley, line 1: No need for this; we assume that suggested packges are installed.

tests/testthat/test-data-table.R
hadley: Maybe just `expect_snapshot(roxygen_ns_show())`?

R/pipe.R
malcolmbarrett: Yes, but `use_import_from()` will run it if needed.

R/tibble.R
malcolmbarrett: I think `use_import_from()` does what you're thinking already

tests/testthat/test-data-table.R
malcolmbarrett: data.table is currently under suggests. Would it be better to add it?

tests/testthat/test-data-table.R
malcolmbarrett: ohh that's a good idea

tests/testthat/test-data-table.R
jennybc: I think you mean "data.table is currently NOT under suggests".

And, no, please do not add it! We have an ongoing battle in usethis: just because we provide some niceties for a package we do NOT take it on as a dependency. We've got various measures in place re: this re: many packages usethis has convenience functions for.

jennybc: So, to be clear, I think this line should not be added, but it has nothing to do with suggests.

jennybc: This is what mocking `check_installed()` is about in tests like these.

malcolmbarrett: Oops, yes, that's what I meant! I had some issues mocking this one, but I can't exactly remember what at the moment. I'll revisit.

malcolmbarrett: Oh, I included in the PR description. Nice, Past Malcolm!

> Because writing NAMESPACE requires loading which requires data.table, I also decided just to skip the tests if data.table is not installed. I tried to hack away at it via mocking, but it didn't seem worth it (or particularly robust in this case). I'm happy to hear other suggestions, though.

Any suggestions?

tests/testthat/test-tibble.R
```diff
@@ -6,6 +6,7 @@ test_that("use_tibble() requires a package", {
 test_that("use_tibble() Imports tibble", {
   create_local_package(path_temp("mypackage"))
   withr::local_options(list(usethis.quiet = FALSE))
+  ui_silence(use_package_doc())
   expect_snapshot(use_tibble())
```
jennybc, line 10: The `use_tibble()` test should probably gain the `with_mock(check_installed = function(pkg) TRUE`, ...)` treatment, since you're here @malcolmbarrett .

Mostly, for consistency. But also because my recent-ish re-removal of rematch2 ?I think? means tibble is no longer an indirect dependency. tibble is still present for testing, just because styler (suggested) depends on it, I think. So it's sort of coincidence that this test is currently passing. https://github.com/r-lib/usethis/issues/1223

tests/testthat/test-data-table.R
malcolmbarrett: Oh, I could mock `roxygen_update_ns()` not to do anything. I think that should be fine for this test, right? Let me try it

jennybc: 😕
So I think this is a concrete downside to roxygenising from usethis.
Kind of related to things we discussed in https://github.com/r-lib/usethis/pull/1377

But yeah, if we keep doing that, you are going to have to skip. I would just bring it inside the tests. And we're going to have to do that in all similar situations.

jennybc: This re-raises the issue of whether we even want functions like `use_tibble()` and `use_data_table()` at all.

https://github.com/r-lib/usethis/issues/1046

R/tibble.R
hadley: Ooooh yeah. I was just confused by the `created`

tests/testthat/test-data-table.R
hadley: Add what?

hadley: I'm missing something here. data.table is listed in suggests, so why wouldn't it be available when you run the tests?

hadley: Oh it's _not_ listed in suggests. I think the easiest fix is to just list it.

malcolmbarrett: data.table is not listed in suggests. I made a typo earlier. HOWEVER, I have been able to mock this successfully, so I think this is resolved

tests/testthat/test-tibble.R
malcolmbarrett: Fair enough. Updated!

R/use_import_from.R
```diff
@@ -29,8 +29,7 @@ use_import_from <- function(package, fun, load = is_interactive()) {
   check_uses_roxygen("use_import_from()")
   check_installed(package)
   check_has_package_doc("use_import_from()")
-
-  purrr::walk2(package, fun, check_fun_exists)
+  check_functions_exist(package, fun)
```
malcolmbarrett, line 32: This isn't strictly necessary, but it makes the mock a little simpler and it's consistent with the other checks

malcolmbarrett:
Ok, I think these comments are all addressed now. In particular, the data.table tests are mocked in a reasonable and safe way without needing to skip or add to suggests, and I extended those mocks to the tibble tests to be extra cautious.

tests/testthat/test-data-table.R
```diff
@@ -5,11 +5,34 @@ test_that("use_data_table() requires a package", {
...
   with_mock(
     check_installed = function(pkg) TRUE,
+    roxygen_update_ns = function(...) NULL,
+    check_functions_exist = function(...) TRUE,
     use_data_table()
   )
   expect_match(desc::desc_get("Imports"), "data.table")
-  datatable_doc <- read_utf8(proj_path("R", "utils-data-table.R"))
-  expect_match(datatable_doc, "#' @import data.table", all = FALSE)
+  package_doc <- read_utf8(proj_path(package_doc_path()))
```
jennybc, line 16: Is this serving a purpose at the moment?

tests/testthat/test-data-table.R
```diff
@@ -5,11 +5,34 @@ test_that("use_data_table() requires a package", {
...
+    check_installed = function(pkg) TRUE,
+    roxygen_update_ns = function(...) NULL,
+    check_functions_exist = function(...) TRUE,
+    expect_warning(
+      use_data_table(),
+      "data.table should be in Imports or Suggests, not Depends"
+    )
+  )
+  expect_match(desc::desc_get("Imports"), "data.table")
+  package_doc <- read_utf8(proj_path(package_doc_path()))
```
jennybc, line 35: Same question as above

tests/testthat/test-data-table.R
malcolmbarrett: You're right, this is ⚰️ code RIP 🙏 🌹

jennybc:
Thanks for bringing this across the line!

I think the CI errors are unrelated.

PR #1398, opened 2021-03-19
Check pkg installed before adding to rprofile

malcolmbarrett:
This PR adds `check_installed()` to `use_rprofile_package()`, something that I bumped into recently while iterating through a few different clean environments. `rlang::check_installed()` is just so dang helpful that I think it makes sense to check!


hadley:
Looks good — can you please add a NEWS bullet?

hadley:
Thanks @malcolmbarrett!

PR #1399, opened 2021-03-23
Update fs linter

malcolmbarrett:
In working on https://github.com/r-lib/devtools/pull/2331, I noticed that this linter (which I was using locally) was missing `system.file()`. It also doesn't seem like there is an fs equivalent of `tempdir()` (and fs itself seems to use `tempdir()`). I also updated a test using `system.file()`

### DIFF
--- a/.github/workflows/lint-undesirable-functions.yaml
+++ b/.github/workflows/lint-undesirable-functions.yaml
@@ -74,7 +74,7 @@ jobs:
             "list.dirs",
             "normalizePath",
             "path.expand",
-            "tempdir",
+            "system.file",
             "tempfile",
             "unlink",
             "basename",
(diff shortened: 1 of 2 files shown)


jennybc:
Seems good re: `system.file()`.

Re: `tempdir()`, I think the fs substitute is `fs::path_temp()`, agree? (But yeah it honestly doesn't matter much.)

https://github.com/r-lib/fs/blob/e10022f2df4571a8598a4caec4929d4b133566d9/R/temp.R#L81-L83

I'm kind of surprised to not see a call to `path_tidy()` in `path_temp()` 🤔
cc @jimhester

jimhester:
`fs::path_temp()` is the fs equivalent of `tempdir()`

malcolmbarrett:
ah, ok! I misinterpreted the use of `tempdir()` in fs as a lack of equivalent, but I see how `path_temp()` is a useful replacement. Will update.

jennybc:
> `path()` calls `path_tidy()`, so there is no need for an explicit call to it

Oh, right, duh.

Thanks @malcolmbarrett!

PR #1404, opened 2021-03-27
Fix LazyData to match new CRAN check

### LINKED ISSUE #1403: LazyData should only be set when data directory is present
It looks like CRAN has implemented a new check that gives a NOTE when LazyData is set and a data directory is not present.

```
checking LazyData ... NOTE
  'LazyData' is specified without a 'data' directory
```

Perhaps LazyData should only be set in the DESCRIPTION after calling `use_data()`, rather than being set in `create_package()`.

malcolmbarrett:
I'm also seeing #1403. Assuming this needs to be dealt with, this PR moves setting `LazyData: true` from `use_description_defaults()` to `use_data()`

Fixes #1403


hadley:
Looks good, thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

malcolmbarrett:
Done, thanks! I also realized the output in the `README` needed to be updated to match this change, so I knit that, as well.

PR #1405, opened 2021-03-29
Use `path_ext_set()` to modify extension

malcolmbarrett:
I noticed a slight inconsistency in `use_github_action()` that used its own solution to managing extensions rather than fs. This PR updates that bit of code to use `path_ext_set()`. I don't think it needs to use the more surgical `slug()` since it's referring to an r-lib file that follows a predictable pattern


hadley:
LGTM. Thanks!

Issue #857, opened 2019-07-29
Ensure alphabetical order when adding a dependency with usethis


lorenzwalthert:
Problem: New dependency added with `use_package()` and `use_dev_package()` is added as last item in the respecive Field in DESCRIPTION. 

Easy fix:  should run `use_tidy_description()` on exit to make sure the added entry in DESCRIPTION is put in the right place. This will in general have unintended side effects with the formatting of DESCRIPTION.

Clean fix: Make sure you instert in the right place in the first place. 

What do you think?

jennybc:
Kind of a duplicate of #443

jennybc:
To be clear, my personal taste is also to alphabetize them and rigorously maintain that style.

hadley:
It's easiest to do this in desc, I think: https://github.com/r-lib/desc/pull/85

PR #1409, opened 2021-03-29
Require desc 1.3.0

malcolmbarrett:
This PR bumps the desc dep to require 1.3.0. It also removes the code in `tidy_desc()` that sorts dependencies and remotes, instead letting `desc_normalize()` do it. Additionally, it improves the test for `use_tidy_description()` to check the order of remotes.

Closes #1394

PR #1414, opened 2021-04-09
Require latest {desc} for smooth experience with usethis::use_package()

lorenzwalthert:
The ordering in {desc} was discussed in https://github.com/r-lib/usethis/issues/857 and fixed in {desc}, then again reported in  https://github.com/r-lib/desc/issues/95 and hopefully fixed now with release of latest CRAN version. I propose to require that version of {desc} at least to make sure `usethis::use_package()` is not disrupting the ordering.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -28,7 +28,7 @@ Imports:
     clipr (>= 0.3.0),
     crayon,
     curl (>= 2.7),
-    desc,
+    desc (>= 1.3.0),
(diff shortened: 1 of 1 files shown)


malcolmbarrett:
I think this is a duplicate of #1409

lorenzwalthert:
Seems so. Thanks. Closing in favour of #1409. Quite surprised that this issue caused so much attention (and too lazy to search open issues, sorry).

Issue #1415, opened 2021-04-10
use_readme_rmd() uses devtools, which is overkill


msberends:
Thanks for this package!

When running `use_readme_rmd()`, it writes this to `README.Rmd`:

> And the development version from [GitHub](https://github.com/) with:
> 
> ``` r
> # install.packages("devtools")
> devtools::install_github("user/project")
> ```

I believe the `devtools` package is a tremendous overkill for just this purpose (given its numerous dependencies and other goals) and it should be advised that users only install `remotes` if they are not developers. So I would change it to:

> And the development version from [GitHub](https://github.com/) with:
> 
> ``` r
> # install.packages("remotes")
> remotes::install_github("user/project")
> ```

jennybc:
This is a topic that has come up many times, in many places.

https://r-pkgs.org/setup.html#setup-usage

https://twitter.com/JennyBryan/status/1370052151444082688

msberends:
Noted, thanks! Still, in our experience most `install_github()`-ers just users wanting the latest dev version for its newest functions or bug fixes, they aren’t devs. So kindly disagree on your advice, but I understand your point that people installing from GitHub might require devtools since they are devs. My last take on this: _those_ users will find devtools anyway. I’d just advise remotes for just `install_github()` 🙂

PR #1417, opened 2021-04-14
Update README.md

prdm0:
Change `devtools::install_github` to `remotes::install_github`.


malcolmbarrett:
The explanation for using devtools just came up in #1415

PR #1420, opened 2021-04-16
Correct casing in `rstudio_config_path()`

### LINKED ISSUE #1425: Change `rstudio_config_path()`
What we currently have is wrong. According to @jmcphers, the IDE folks didn't realize rappdirs has(and official XDG docs suggest) distinct behavior for Mac and Linux. In hindsight, RStudio should probably be more like rappdirs.

But all we need to worry about in usethis is matching what the IDE actually does.

Official docs:

https://support.rstudio.com/hc/en-us/articles/200534577-Resetting-RStudio-Desktop-s-State
https://docs.rstudio.com/ide/desktop-pro/settings.html

The real solution is for RStudio to just tell us where they look:

https://github.com/rstudio/rstudioapi/issues/176

Research inspired by reviewing #1420, which should probably be superseded by the fix for this.

charliejhadley:
Presently `rappdirs::user_config_dir("RStudio", os = "unix")` fails in Linux due to case sensitive file paths (see https://github.com/charliejhadley/rstudioprefswitcher/issues/1).

Closes #1425

### DIFF
--- a/R/rstudio.R
+++ b/R/rstudio.R
@@ -175,7 +175,7 @@ rstudio_config_path <- function(...) {
     base <- rappdirs::user_config_dir("RStudio", appauthor = NULL)
   } else {
     # RStudio only uses windows/unix conventions, not mac
-    base <- rappdirs::user_config_dir("RStudio", os = "unix")
+    base <- rappdirs::user_config_dir("rstudio", os = "unix")
   }
   path(base, ...)
 }


jennybc:
I am looking into our general strategy for finding RStudio preferences because my initial look suggests we might need a larger change.

jennybc:
Thanks! Sorry it took me so long to read all of this carefully. Will merge momentarily.

PR #1423, opened 2021-04-16
Draft use_tidy_dependencies()

### LINKED ISSUE #1418: use_tidy_dependencies()
```R
use_tidy_dependencies <- function() {
  use_package("rlang")
  roxygen_ns_append("@import rlang")
  
  use_package("ellipsis") # for now
  
  # should lifecycle be included? 
  use_package("lifecycle")
  roxygen_ns_append("@importFrom lifecycle deprecated")
  
  use_package("cli")
  use_package("glue")
  roxygen_ns_append("@importFrom glue glue")
  
  use_package("withr")
  
  # compat files?
  # copy in rlang::compat-purrr.R 
  # compat-purrr.R can assume rlang is available and use as_function()
  # 
  # will need to consider conflicts
}
```

hadley:
And apply it to usethis. Fixes #1418.


R/roxygen.R
```diff
@@ -78,11 +78,17 @@ roxygen_remind <- function() {
   TRUE
 }
 
-roxygen_update_ns <- function() {
+roxygen_update_ns <- function(load = TRUE) {
```
jennybc, line 81:
```suggestion
roxygen_update_ns <- function(load = is_interactive()) {
```

R/tidyverse.R
```diff
@@ -101,6 +104,37 @@ use_tidy_description <- function() {
...
+  imports <- roxygen_ns_append("@import rlang") ||
+    roxygen_ns_append("@importFrom glue glue")
+  if (imports) {
+    roxygen_update_ns(TRUE)
+  }
+
+  # If needed, copy in lightweight purrr compatibility layer
+  if (!desc::desc(proj_get())$has_dep("purrr")) {
+    url <- "https://raw.githubusercontent.com/r-lib/rlang/master/R/compat-purrr.R"
+    path <- tempfile()
```
jennybc, line 130:
```suggestion
    path <- path_temp()
```

R/tidyverse.R
```diff
@@ -101,6 +104,37 @@ use_tidy_description <- function() {
...
+    utils::download.file(url, path, quiet = TRUE)
+    write_over("R/compat-purrr.R", readLines(path))
```
jennybc, line 132: Just for internal consistency?
```suggestion
    write_over("R/compat-purrr.R", read_utf8(path))
```

hadley: This is why we need a linter 🤪

R/tidyverse.R
```diff
@@ -107,6 +107,10 @@ use_tidy_description <- function() {
 #' @export
 #' @rdname tidyverse
 use_tidy_dependencies <- function() {
+  if (!check_has_package_doc()) {
```
jennybc, line 110: Just noting for the sake of creating shared understanding:

I think `check_*()` functions should not have meaningful return values, i.e. they exit if the to-be-checked condition doesn't hold.

Do we agree? I think we are awfully close to that being true already.

hadley: Yeah, I'll change this function. It confused me in the tests so I should've made the change already.

(I think the only exception to this rule is sometimes an argument checking function might do some light massaging (like `tolower()` or `as.integer()`) so `x <- check_x(x)` is ok)

R/tidyverse.R
```diff
@@ -10,6 +10,9 @@
 #' of the tidyverse conventions as possible, issues a few reminders, and
 #' activates the new package.
 #'
+#' * `use_tidy_dependencies()`: sets up standard dependencies used by all
+#'   tidyverse packages (except packages that are designed to dependency free).
```
krlmlr, line 14:
```suggestion
#'   tidyverse packages (except packages that are designed to be dependency free).
```

R/tidyverse.R
```diff
@@ -101,6 +104,47 @@ use_tidy_description <- function() {
...
+  use_dependency("cli", "Imports")
+  use_dependency("glue", "Imports")
+  use_dependency("withr", "Imports")
+
+
+  # standard imports
+  imports <- any(
+    roxygen_ns_append("@import rlang"),
+    roxygen_ns_append("@importFrom glue glue"),
+    roxygen_ns_append("@importFrom lifecycle deprecated")
```
krlmlr, line 124: Is there a reason we wouldn't import all of lifecycle?

R/tidyverse.R
jennybc: Also re: shared understanding:

In the GitHub areas of usethis, I have a few `challenge_*()` functions for when I want to make sure the user understands they're doing something possibly weird / dangerous, but if they accept the risk, we keep going.

R/tidyverse.R
hadley: The reason for importing `deprecated()` is that you use it in arguments; for everything else I think it's better to be explicit.

PR #1424, opened 2021-04-16
Assume git users can undo

hadley:
Fixes #1376

This is a very simple change but has big consequences. I think it's a big UI improvement for us, and at most only a minor inconvenience for others. What do you think?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # usethis (development version)
 
+* usethis will now overwrite files without prompting if you are using git
+  because you can undo if it's a problematic change (#1376).
+
 * `use_code_of_conduct()` now requires a `contact` argument to supply contact
   details for reporting CoC violations (#1269).
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -3,6 +3,10 @@ can_overwrite <- function(path) {
     return(TRUE)
   }
 
+  if (uses_git()) {
+    return(TRUE)
+  }
+
   if (is_interactive()) {
     ui_yeah("Overwrite pre-existing file {ui_path(path)}?")
   } else {


jimhester:
Should we have some way to opt-in e.g. if you wanted a safe-mode version of usethis that would prompt about overwriting?

Or alternatively could we leave the defaults as-is, and just provide a way to globally opt out, e.g. an envvar or option we could put in our .Rprofile.

jennybc:
> very simple change but has big consequences

No joke!

I have some concern because we project an attitude of "just get started with your first package, it's easy!" and that could easily happen around the same time as "just get started with Git and do very basic stuff" (which probably doesn't include resets).

So there could be people new to both Git and package development, who feel like we sent them down that path. And now we do this!

If we do it, we should provide a function or other support for the relevant "git undo".

hadley:
Ok, since you both have concerns that this is too much (and I agree), I'll make this opt-in so that you also have to set some option.

hadley:
Ok, take a look now and let me know if you have any better ideas for the option name.

NEWS.md
```diff
@@ -1,7 +1,7 @@
 # usethis (development version)
 
-* usethis will now overwrite files without prompting if you are using git
-  because you can undo if it's a problematic change (#1376).
+* usethis will now overwrite files without prompting if you set
+  `option(usethis.auto_write = TRUE)` and are using git (#1376).
```
jennybc, line 4: Other ideas: `usethis.auto_overwrite` or `usethis.auto_write_over`.

jennybc (commented):
This needs to be noted in:

https://usethis.r-lib.org/articles/articles/usethis-setup.html

probably specifically in https://usethis.r-lib.org/articles/articles/usethis-setup.html#store-default-values-for-description-fields-and-other-preferences

NEWS.md
hadley: `usethis.safety = "off"` ?

jennybc: I prefer the specificity of the other alternatives. `usethis.safety = "off"` would need more explanation. Of course, all of them would need to be documented, but I think the need to re-read those docs would be much higher with `usethis.safety`.

jimhester: What about `usethis.overwrite = TRUE`, that seems relatively self documenting to me.

hadley: Yeah, I like that.

jimhester (approved):

vignettes/articles/usethis-setup.Rmd
```diff
@@ -70,6 +70,8 @@ Certain options are consulted by usethis and allow you to set personal defaults:
     console.
   * `usethis.destdir`: a default directory to use in `create_from_github()` and
     `use_course()`.
+  * `usethis.overwrite`: if `TRUE` and you're using git in the current project,
+    usethis will overwrite files without prompting
```
jennybc, line 74:
```suggestion
  * `usethis.overwrite`: if `TRUE` and you're using Git in the current project,
    usethis overwrites existing files without asking. You will see and can selectively accept/discard any proposed changes as part of the normal Git workflow.
```

jennybc (approved):
I proposed some minor wording changes.

It feels like we should add a test. If you've moved on to something else @Hadley, tell me and add that to finish this.

hadley:
Sure, that'd be great, thanks!

jennybc:
Observation about this PR now that I'm writing tests:

We have been trying to keep low-level helpers like `write_over()` (exported) and `can_overwrite()`(not exported) "project-agnostic". I.e. the caller is responsible for providing the path and these functions don't know about projects.

Main principle: we don't want calls to these low-level helpers to activate or change the active project.

This PR currently violates this because it has `write_over()` -> `uses_git()` -> `proj_get()`. We've run up against this before with `write_utf8()` and `write_union()`. These dilemmas are discussed in the principles document:

https://github.com/r-lib/usethis/blob/master/principles.md#helpers-and-the-active-project

I think this PR need to be adjusted similar to those other cases, such that `write_over()` can't trigger project activation.

What should we do if the active project uses Git, but I `write_over(path, ...)` a `path` outside the project? This is an edge case for usethis, because we're so project-focused, but still.

hadley:
Can we just base the git check on the file path vs. the active project?

jennybc:
> Can we just base the git check on the file path vs. the active project?

I will explore this, but yeah sounds like a good solution.

jennybc:
It's not too hard to check if `path` is in a Git repo, without tickling the active project.

If no one has anything else to say, I'll merge this in a bit.

jennybc:
Merged.

I'd urge anyone who's listening to consider setting `usethis.overwrite = TRUE` in `.Rprofile`, so we accumulate some real-world usage. I have just done so myself.

PR #1427, opened 2021-04-16
Document dangling typo

malcolmbarrett:
Looks like a typo got fixed without rendering the matching doc. Let's get this little feller fixed up


jennybc:
Thanks!

The CI error is unrelated. Either it's intermittent or something has changed about the API and `github_remotes()` needs to adapt. But as I said, totally unrelated.

PR #1433, opened 2021-04-22
prefer no spaces between arguments in chunk options

kevinushey:
A style tweak for the default vignette.

### DIFF
--- a/inst/templates/vignette.Rmd
+++ b/inst/templates/vignette.Rmd
@@ -7,7 +7,7 @@ vignette: >
   %\VignetteEncoding{UTF-8}
 ---
 
-```{r, include = FALSE}
+```{r include=FALSE}
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>"


jennybc:
Problem is, I really dislike that style 😬

Can we discuss context more? Presumably the current template isn't actively harmful, but you must have some reason for preferringnospaces.

kevinushey:
This is mainly because it's the "default" style used in the knitr documentation, e.g.

https://yihui.org/knitr/options/
https://bookdown.org/yihui/rmarkdown/r-code.html

I believe the motivation for this style is mainly to avoid overly-long chunk headers -- saving 2 characters on option does add up.

jennybc:
My inclination is to pass on this one, as I would then need to edit this in new vignettes each time I create one!

I see that @malcolmbarrett seems to agree with me (?). @Hadley do you have an opinion?

hadley:
I really dislike the no-space formatting.

malcolmbarrett:
Yes, TBH I'm always fighting the IDE's autocomplete on this one

kevinushey:
I mean, I prefer spaces as well, but the canonical form for chunk options is no spaces around `=`. This is the form we've chosen to adopt and use in other parts of the IDE as well, e.g. in our existing templates:

https://github.com/rstudio/rstudio/blob/master/src/cpp/session/resources/templates/document.Rmd

I think it's more important to be consistent with existing convention, especially if that convention is already well-established in existing documentation and tools.

jennybc:
I find that I still feel the same way: I really dislike the no space style.

Maybe the new ability (in knitr 1.35) to specify chunk options in special comments inside the chunk heralds the dawn of a new era, where it's cool to make chunk options more readable 🤓

PR #1437, opened 2021-04-25
Don't depend on suggested packages 'testthat' and 'mockr'

dmurdoch:
Cascades of errors in the winUCRT tests on CRAN are caused by tests using `testthat` unconditionally.  This changes the template so that `testthat` is only used if present, and a warning is given if not.

The second commit skips tests that need `mockr` if it is not found.

### DIFF
--- a/inst/templates/testthat.R
+++ b/inst/templates/testthat.R
@@ -1,4 +1,6 @@
-library(testthat)
-library({{{ name }}})
+if (require(testthat)) {
+  library({{{ name }}})
 
-test_check("{{{ name }}}")
+  test_check("{{{ name }}}")
+} else
+  warning("'{{{ name }}}' requires 'testthat' for tests")
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -1,4 +1,6 @@
-library(testthat)
-library(usethis)
+if (require(testthat)) {
+  library(usethis)
 
-test_check("usethis")
+  test_check("usethis")
+} else
+  warning("'usethis' requires 'testthat' for tests")
--- a/tests/testthat/test-ci.R
+++ b/tests/testthat/test-ci.R
@@ -1,5 +1,6 @@
 test_that("use_circleci() configures CircleCI", {
   skip_if_no_git_user()
+  skip_if_not_installed("mockr")
 
   local_interactive(FALSE)
   create_local_package()
--- a/tests/testthat/test-course.R
+++ b/tests/testthat/test-course.R
@@ -1,6 +1,7 @@
 ## download_url ----
 
 test_that("download_url() retry logic works as advertised", {
+  skip_if_not_installed("mockr")
   faux_download <- function(n_failures) {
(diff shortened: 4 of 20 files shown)


jennybc:
Tagging you in here @Hadley as this essentially represents a policy change.

I have often used `skip_if_not_installed()` for suggested packages elsewhere, although obviously not (much?) here in usethis. And I'm pretty sure it's not your usual M.O.

But the more substantive change is in the template for `testthat.R`:

``` r
if (require(testthat)) {
  library({{{ name }}})

  test_check("{{{ name }}}")
} else
  warning("'{{{ name }}}' requires 'testthat' for tests")
```

dmurdoch:
For context, see this thread:  https://hypatia.math.ethz.ch/pipermail/r-package-devel/2021q2/006881.html .  The basic problem is that CRAN won't install things with errors in their checks, and packages that suggest `testthat` then use it unconditionally generate errors, so CRAN won't install them.  This is causing real problems.

hadley:
This would be a major policy change; we (and basically every other package that uses testthat) uses it unconditionally because we assume that suggested packages are installed for checking.

hadley:
Hmmmm, isn't this is rather dangerous? If testthat is not available none of your tests will be run, and you'll think all is well because there are zero test failures.

dmurdoch:
I agree that everyone uses it that way, and that's a problem.  This doesn't fix that problem, but it prevents a similar issue from arising in the future with new packages.

I've also submitted PRs to change a few packages to match this template with the faint hope that the subset I found will be sufficient to get testthat working on winUCRT (so that I can get past the missing dependency issue, and see if rgl actually works there or not), but I suspect more such changes will be needed.  I don't know how to find the minimal set.

Regarding the danger:  If you set your script to fail on a warning, you'll still see get a failure with this template.  I believe a warning is not sufficient to prevent a package from being built on CRAN (though it would prevent a new submission from being accepted), but errors prevent the package from being installed at all.  

So currently rgl won't install, because of missing builds of 'htmlwidgets', 'htmltools', 'knitr', 'jsonlite', 'shiny', 'magrittr', 'crosstalk', 'manipulateWidget'.  'htmlwidgets' also needs 'yaml'.  'htmltools' also needs 'rlang', which needs 'glue', which needs 'magrittr'.

I don't know any way out of this logjam other than you or CRAN being a little flexible.  You don't need to change the template to fix it, but since copying the template causes the logjam, it makes sense to me to make the change.  

And currently the template violates the contract that listing something as a Suggested package implies (i.e. that you won't try to use the package if it's not there).  That's not your fault:  CRAN won't let most packages list 'testthat' as a stronger dependency, but it's the way things are.

hadley:
I am happy to be flexible to figure some way out of this logjam, but I'd prefer to try a few other approaches before we commit to this one.

jennybc:
Re-skimming the above, it sounds like I can close this.

PR #1441, opened 2021-05-03
Update `use_tidy_dependencies()`

lionel-:
- Remove ellipsis now that it's folded in rlang (see https://github.com/r-lib/rlang/issues/1057).
- Remove lifecycle following discussion at the team meeting. As @DavisVaughan mentioned, lifecycle is not needed when starting new packages.


hadley:
You don't _need_ it immediately, but it's lightweight and you'll need it eventually. (And this function is also applied to existing packages)

lionel-:
ok I have removed the lifecycle commit.

PR #1442, opened 2021-05-03
TYPO: Added testthat:: prefix for example

muschellij2:
Very simple change that isn't technically a typo, but the change is simply adding the testthat:: namespace for the default testthat example.

### DIFF
--- a/inst/templates/test-example-2.1.R
+++ b/inst/templates/test-example-2.1.R
@@ -1,3 +1,3 @@
-test_that("multiplication works", {
+testthat::test_that("multiplication works", {
   expect_equal(2 * 2, 4)
 })


malcolmbarrett:
testthat gets loaded when testing. Why would we need to add the namespace?

muschellij2:
True - you don't.  Only helps if you're running some tests interactively when not the full test suite - not the best practice, but still some people's workflow.

malcolmbarrett:
Hmm, I don't think we should codify that workflow here. We'd also have to namespace the expect function, and I think it's counter to both idiomatic testthat code and the test template that loads testthat. Thanks anyway!

PR #1443, opened 2021-05-04
Drop link to urlchecker

hadley:
Since it's now on CRAN (and the link makes it harder to copy and paste into R)


hadley:
I was just thinking about doing a PR for this 😆 Thanks past Hadley!

Issue #1450, opened 2021-05-19
New GitHub PAT Causing Error in validate_gh_pat(new_gh_pat(x))


atheobold:
I accidentally changed my PAT while demonstrating how `gitcreds::gitcreds_set()` works. I then decided the best plan was to regenerate a new PAT. However, upon trying to work through the process of having my PAT recognized, I've come across some major issues. 

When I run `gitcreds::gitcreds_set()`, I select option `2`, replacing the current credentials for https://github.com. I see the following output: 
```
-> Removing current credentials...

? Enter new password or token: NEW_PAT_ENTERED
-> Adding new credentials...
-> Removing credetials from cache...
-> Done.
```

However, when I then run `gh::gh_whoami()`, I get the following output:
``` r
gh::gh_whoami()
#> Error in validate_gh_pat(new_gh_pat(x)): A GitHub PAT must consist of 40 hexadecimal digits
```
<sup>Created on 2021-05-18 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

malcolmbarrett:
Have you tried updating gh? See this thread for some suggestions on this error: https://twitter.com/JennyBryan/status/1394824309554630656

atheobold:
@malcolmbarrett -- This completely fixed it! You are a life saver!

jennybc:
For anyone else who stumbles here, a more official document:

https://usethis.r-lib.org/articles/articles/git-credentials.html

robo-atheobold:
@jennybc -- I love the Git Credentials set-up you linked! In fact I reference it *every* time I teach version control. That being said, after spending hours pouring over this and the documentation for the **credentials** and **gitcreds** packages, I'm a bit disappointed that there was no indication that I needed to update **gh**. 

Could a sentence be added to the [Put your PAT into the Git credential store](https://usethis.r-lib.org/articles/articles/git-credentials.html#put-your-pat-into-the-git-credential-store) section, suggesting how to troubleshoot this issue? I'm happy to put in a PR for this change, especially since I can use the `pr_XXX()` functions again!

atheobold:
@jennybc -- I love the Git Credentials set-up you linked! In fact I reference it *every* time I teach version control. That being said, after spending hours pouring over this and the documentation for the **credentials** and **gitcreds** packages, I'm a bit disappointed that there was no indication that I needed to update **gh**.

[...]

PR #1454, opened 2021-06-21
update minimum version of gh

ijlyttle:
A minimal PR to help with #1450 - I don't know if this merits a NEWS item; I'll be happy to add one if you think it does.


jennybc:
Thanks!

PR #1460, opened 2021-07-15
Add 'easystats' to list of verboten meta-packages

bwiernik:
["easystats"](https:://github.com/easystats/easystats) is a meta-package similar to the tidyverse with many component packages ("performance", "parameters", "modelbased", etc.). It is designed to make probing, reporting, and plotting many kinds of statistical models in R easier, especially for new R users. Similar to `library(tidyverse)`, some statistics classes have begun to teach `library(easystats)`, which is appropriate for an analysis project, but not for package development.

### DIFF
--- a/R/package.R
+++ b/R/package.R
@@ -30,7 +30,7 @@
 #' }
 use_package <- function(package, type = "Imports", min_version = NULL) {
   if (type == "Imports") {
-    refuse_package(package, verboten = c("tidyverse", "tidymodels"))
+    refuse_package(package, verboten = c("tidyverse", "tidymodels", "easystats"))
   }
 
   changed <- use_dependency(package, type, min_version = min_version)


jennybc:
I'm not willing to merge this for (at least) two reasons: it doesn't feel right to hard-wire this for a package that's

1. outside the tidyverse umbrella, i.e. a package "we" don't "own".
  - Also: sets a precedent that's not sustainable. In general, we are learning that the overall success of usethis support is highest for packages within our direct sphere of influence and falls off pretty sharply from there. This is about functionality/usability and also maintainability.
2. not at high-risk of being `Import`ed on a large-scale

Hope this makes sense.

bwiernik:
That's understandable. I just wanted to propose and see what your thoughts would be. Thanks

PR #1463, opened 2021-07-22
dependences to dependencies line 23

adithirgis:
dependences to dependencies line 23


jennybc:
Thanks!

I feel like this is one of those files where there is a more central and definitive source for it that should also be checked for spelling. Maybe @batpigandme knows.

"Luckily" this is a sentence that will likely need updating before too long because we'll switch to suggesting people use pak.

PR #1465, opened 2021-07-24
Check testthat availability in testthat template

krlmlr:
Closes r-lib/testthat#1420.

### DIFF
--- a/inst/templates/testthat.R
+++ b/inst/templates/testthat.R
@@ -1,4 +1,6 @@
-library(testthat)
-library({{{ name }}})
-
-test_check("{{{ name }}}")
+if (require(testthat)) {
+  library({{{ name }}})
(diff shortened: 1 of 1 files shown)


jennybc:
@Hadley Do you have an opinion on this? I believe historically your position has been that the presence of testthat is *de facto* assumed, if you are actively *running the tests*. Does the linked issue change things enough that we want to change the testthat template in usethis?

hadley:
I'd want to understand the problem more before committing to this fix — it seems dangerous to just silently succeed when no tests have in fact been run.

jennybc:
This conversation was basically also in #1437 and I'm closing this based on

> I am happy to be flexible to figure some way out of this logjam, but I'd prefer to try a few other approaches before we commit to this one.

PR #1467, opened 2021-07-26
incorporate git `core.excludesfile`

### LINKED ISSUE #1461: `~/.gitignore` is ignored
I realize that I can't provide a reproducible example; I need to tell a story (sorry).

I recently upgraded my Mac to Big Sur. Of course, one of the first things I did was `install.packages("usethis")`, then `usethis::git_vaccinate()`.

I kept on seeing `.DS_Store` wanting to be committed, which was odd, so I kept adding it to local `.gitignore`. Finally, this morning, I said "enough" and started digging into it.

Of course, my `~/.gitignore` file looked OK, so a little investigation brought me to this [SO answer](https://stackoverflow.com/a/32487076), and sure enough, I did not have git's `core.excludesfile` set.

So I set `core.excludesfile`, and `.DS_Store` "went away".

I see that in [`edit_git_ignore()`](https://github.com/r-lib/usethis/blob/0d181f6905392b0c543a490d6047613334404bad/R/edit.R#L222-L227), if there is no global gitignore file: you set `core.excludesfile`, then call `git_vaccinate()`.

However, if you just call [`git_vaccinate()`](https://github.com/r-lib/usethis/blob/b3de0d7c18fdc1d0de0dc7616cbb85e823db9dac/R/git.R#L565-L568), I don't see that `core.excludesfile` gets set. I think this is how I got here.

As well [`git_vaccinated()`](https://github.com/r-lib/usethis/blob/b3de0d7c18fdc1d0de0dc7616cbb85e823db9dac/R/git.R#L570-L574) does not check that `core.excludesfile` is set.

I have a few ideas on some solutions, and will be happy to contribute, but I figure I'd let you react first.

Thanks!

ijlyttle:
Fix #1461

This is a first pass - just to get it off my computer and to give you something to react to (if you wish).

There are a couple of new internal functions: `git_excludesfile_global_get()` and `git_excludesfile_global_set()`. The new code is probably not as concise as it could be; I have tried to favour readability and preserving existing logic.

I'll want to look over the code with fresh eyes before removing the "draft" status.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -31,6 +31,8 @@
  
 * `use_package()` no longer guides the user on how to use a dependency when no 
   change was made (@malcolmbarrett, #1384)
+
+* `git_vaccinate()` and `git_vaccinated()` incorporate the value of git `core.excludesfile` (#1461, @ijlyttle).
   
 * The minimum version of gh has been bumped to help more people upgrade to the gh version that supports current GitHub PAT formats (#1454 @ijlyttle).
 
--- a/R/edit.R
+++ b/R/edit.R
@@ -221,9 +221,6 @@ edit_git_ignore <- function(scope = c("user", "project")) {
   file <- git_ignore_path(scope)
   if (scope == "user" && !file_exists(file)) {
     ui_done("Setting up new global gitignore: {ui_path(file)}")
-    # Describe relative to home directory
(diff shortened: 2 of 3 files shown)


ijlyttle:
superseded by #1477

Issue #1400, opened 2021-03-24
Mention remotes in GitHub creds vignette?


malcolmbarrett:
One issue that I've had in sending people to the GH creds vignette is that remotes doesn't yet support the git cred store approach used by gert and gh (https://github.com/r-lib/remotes/issues/587), a problem which we bump into quickly since we use private GH repos on our team.

Hopefully, this will be a non-issue soon, but is it worth mentioning (temporarily?) in the vignette that remotes doesn't yet support this approach?

jennybc:
Related to #1413 (where the coffee table the user banged their shins on was exactly `remotes::install_github()`).

jonthegeek:
I just banged my shins on this one, as well! I was setting up a new installation, and couldn't figure out why `remotes::install_github()` couldn't see my org's private repos, despite everything looking right in `usethis::git_sitrep()`. Adding the `GITHUB_PAT` in `.Renviron` took care of it.

jennybc:
I'd rather not go back to encouraging storage in `.Renviron`. I think a better temporary workaround (for example, until remotes gains this feature or something else saves us) is to tickle gitcreds in `.Rprofile`, i.e. to cause your PAT to be eagerly retrieved from the store and made available as `GITHUB_PAT` in the current R session. This is still suboptimal, because `GITHUB_PAT` is being set in all sessions, whether it's needed or not. But it's still safe than putting the PAT in `.Renviron`.

maxheld83:
I ran into this also, and I agree that it'd be nice to alert users to this in the vignette (until https://github.com/r-lib/remotes/issues/587 is resolved or we otherwise reach github auth nirvana).

I've also noticed that **`gitcreds::gitcreds_set()` and `credentials::set_github_pat()` don't (always) play nice together**: https://github.com/r-lib/credentials/issues/16.

I've thus decided to ** *only* rely on gitcreds** and amend my `.Rprofile` thus:

```r
# required b/c remotes can't access Git credential managers https://github.com/r-lib/remotes/issues/516
Sys.setenv(GITHUB_PAT = gitcreds::gitcreds_get(use_cache = FALSE)$password)
```

This fixes remotes.
Happily, gert also works with this, even though it depends on the rival (?) 

Still missing:

> suitably wrapped w/ some attention to interactivity, messaging, and possible failure

PR #1468, opened 2021-07-27
use remotes instead of devtools in the template

rmflight:
Users shouldn't have to install devtools to install the package from github

From my understanding devtools is using remotes under the hood anyway, and remotes has fewer dependencies.

### DIFF
--- a/inst/templates/package-README
+++ b/inst/templates/package-README
@@ -35,7 +35,7 @@ And the development version from [GitHub](https://github.com/) with:
 
 ``` r
 # install.packages("devtools")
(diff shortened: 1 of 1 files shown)


jennybc:
Please see #1415 #1400 and links therein. The next time this changes is likely to be when we can go all in on pak.

rmflight:
Ah, my bad @jennybc. I should have done some googling around to double check if this had been raised before.

jennybc:
No worries. Hopefully we can soon recommend pak for general use.

PR #1481, opened 2021-09-08
Functions to change or adapt to change in Git default branch

### LINKED ISSUE #1243: Detect if `master` --> `main`
Can we help people discover that the source repo has switched from master to main?

### LINKED ISSUE #1371: `use_git_default_branch()`
Our new team default is to use `main` as our default branch, and I'm adding a helper function to our usethis extension package to set this up. However, I'm wondering if usethis itself is a good fit for this function given the interest some folks have expressed in making this transition. Right now it looks something like this:

```r
use_git_default_branch <- function() {
  default_branch <- git_cfg_get("init.defaultbranch")
  if (isTRUE(default_branch == "main")) {
    return(invisible(NULL))
  }

  if (!is_empty(default_branch)) {
    override <- ui_yeah("Your current is default branch is \\
    {ui_value(default_branch)}. Set to {ui_value('main')}?")

    if (!override) {
      return(invisible(NULL))
    }
  }

  ui_done("Setting {ui_field('init.defaultBranch')} \\
          to {ui_value('main')} in global git config")
  gert::git_config_global_set("init.defaultBranch", "main")
}
```

That said, `gert::git_config_global_set("init.defaultBranch", "main")` itself is so simple, that it might just be good to stick in some documentation somewhere, instead!

jennybc:
Closes #1243, Closes #1371


R/git-default-branch.R
```diff
@@ -0,0 +1,408 @@
...
+#' enact or change.
+#'
+#' @return A branch name
+#' @export
+#'
+#' @examples
+#' \dontrun{
+#' git_default_branch()
+#' }
+git_default_branch <- function() {
```
jennybc, line 35: This function pre-existed. I just moved its definition from one file to another and renamed it (well, the old version still works but generates a deprecation warning).

I originally named it `git_branch_default()`, out of deference to all the `gert::git_branch_*()` functions. But now I think it makes more sense to use the more natural `git_default_branch()`, in the usethis context.

This needs to somehow be integrated or refactored in light of new default branch detection use cases. The tricky part is that sometimes we want to infer default branch in a very naive and lightweight way and sometimes we want to be skeptical and actively check for change in remote places.

R/git-default-branch.R
```diff
@@ -0,0 +1,408 @@
...
+#' Please call [git_default_branch()] instead. In hindsight, that is a better
+#' name for this function.
+#' @export
+#' @rdname usethis-defunct
+git_branch_default <- function() {
+  lifecycle::deprecate_soft("2.1.0", "git_branch_default()", "git_default_branch()")
+  git_default_branch()
+}
+
+#' Change (or discover a change in) the default Git branch
```
jennybc, line 115: This is the truly new stuff.

R/git-default-branch.R
```diff
@@ -0,0 +1,408 @@
...
+    },
+    error = function(e) NA_character_
+  )
+  if (is.data.frame(res)) {
+    res <- path_file(res$symref[res$ref == "HEAD"])
+  }
+  res
+}
+
+guess_local_default_branch <- function(ask = is_interactive()) {
```
jennybc, line 370: This is what (maybe) needs to become more integrated with the exported `git_default_branch()`.

R/git-default-branch.R
```diff
@@ -0,0 +1,408 @@
...
+  }
+
+  if (!ask || !is_interactive()) {
+    ui_done("
+      Local branch {ui_value(propose)} appears to play the role of \\
+      the default branch.")
+    return(propose)
+  }
+
+  if (ui_yeah("
```
jennybc, line 399: Hadley thinks I'm being too conservative and could proceed without this user confirmation. I see that point, but also kind of want the user to sign off before we plunge ahead.

R/git-default-branch.R
```diff
@@ -0,0 +1,408 @@
...
+  gb <- set_names(gb)
+  usual_suspects_branchname <- c("main", "master", "default")
+  branch_candidates <- purrr::discard(gb[usual_suspects_branchname], is.na)
+
+  if (length(branch_candidates) == 1) {
+    return(branch_candidates[[1]])
+  }
+  # either 0 or >=2 of the usual suspects are present
+
+  # Can we learn what HEAD points to on a relevant Git remote?
```
jimhester, line 53: Yes, use `gert::git_remote_ls()`

jimhester: Or I guess you are doing it a different way below

jennybc:
The linter failure is completely unrelated to the content of this PR and I'll handle that separately.

PR #1499, opened 2021-09-28
Replace ui_yeah()/ui_nope() default args with newly exposed vectors

charliejhadley:
This PR would be useful for me to reuse {usethis} yes/no options and might be useful internally elsewhere in {usethis}.


jennybc:
I'm closing this as I think it's best as a little wrapper out in a package-that-uses-usethis.

More relevant context: the plan is to get usethis out of the business of providing UI functions to other packages. usethis itself will begin to get this functionality from cli (although not right this moment). That is another reason that the `ui_*()` functions here aren't going to get more love. Also related: https://github.com/r-lib/cli/issues/228.

PR #1501, opened 2021-09-30
Document release_bullets()

hadley:


jennybc (approved):
LGTM

Sidebar: should there be some relationship between `release_bullets()` and `release_questions`? 🤔

https://github.com/r-lib/devtools/blob/d5901f8b1b986cfc623a3615e511015ad468ab0f/R/release.R#L142-L151

https://github.com/search?q=release_questions+org%3Atidyverse+org%3Ar-lib&type=code

https://github.com/search?q=release_bullets+org%3Atidyverse+org%3Ar-lib&type=code

hadley:
Hmmm I can’t tell if I forgot about release_questions or deliberately meant it to be different, but I don’t see why they should be.

PR #1502, opened 2021-09-30
Use bs4 with dev pkgdown

hadley:
Fixes https://github.com/r-lib/pkgdown/issues/1580

@maelle is this what you were thinking of?


NEWS.md
```diff
@@ -46,6 +46,9 @@ usethis has a more sophisticated understanding of the default branch and gains s
   
 * `use_lifecycle()` now imports `lifecycle::deprecated()` (#1419).
 
+* `use_pkgdown()` automatically uses bs4 if you have the devel version
```
maelle, line 49: maybe write a version number rather than devel version?

maelle:
Yes!

R/pkgdown.R
```diff
@@ -30,25 +29,31 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
...
   check_installed("pkgdown")
 
-  use_build_ignore(c(config_file, destdir))
-  use_build_ignore("pkgdown")
+  use_build_ignore(c(config_file, destdir, "pkgdown"))
   use_git_ignore(destdir)
 
-  ui_todo("
-    Record your site's {ui_field('url')} in the pkgdown config file \\
-    (optional, but recommended)")
```
jennybc, line 39: Do you really think we should omit all nudges about this? As it stands, we just silently put this into the config and open the file:

```
url: ~
```

That's not OK, right? Are we hoping the user also realizes they need to fill this in?

R/pkgdown.R
```diff
@@ -30,25 +29,31 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
...
+  config <- pkgdown_config(destdir)
+  config_path <- proj_path(config_file)
+  write_over(config_path, yaml::as.yaml(config))
+  edit_file(config_path)
 
-  if (has_logo()) {
-    pkgdown_build_favicons(proj_get(), overwrite = TRUE)
```
jennybc, line 42: Is favicon creation being done by pkgdown itself now?

Because if so, there's some more stuff we can remove when we delete this:

``` r
# usethis itself should not depend on pkgdown
# all usage of this wrapper is guarded by `check_installed("pkgdown")`
pkgdown_build_favicons <- function(...) {
  get("build_favicons", asNamespace("pkgdown"), mode = "function")(...)
}
```

(There is one remaining usage in `use_pkgdown_travis()` btw, which we should also remove.)

And `has_logo()` can probably go. And some mocks in tests ....

DESCRIPTION
```diff
@@ -56,6 +56,7 @@ Suggests:
     knitr,
     magick,
     mockr,
+    pkgdown,
```
jennybc, line 59: I suspect this will prove to be unnecessary.

R/pkgdown.R
jennybc: If you confirm that usethis is out of the favicon creation business, I can do all the other cleanup I refer to above.

R/pkgdown.R
hadley: `~ is fine for a default because it'll be read as `NULL`. 

I thought we also nudged elsewhere?

R/pkgdown.R
hadley: Yeah, usethis should definitely be out of the favicon creation business as pkgdown now does it (if needed) when you build the site.

DESCRIPTION
hadley: I added it because something was failing, but now I don't see what. Maybe I removed that code?

R/pkgdown.R
jennybc: In `use_pkgdown_github_pages()`, when dealing with our own stuff (repo belongs to tidyverse or r-lib orgs), we actually insert the (future?) URL into pkgdown config, because we know what it will be. I think that is what you are thinking of.

But for the general population, I think this is our only nudge, so suspect it should stay.

DESCRIPTION
jennybc: I think the failure may have had another explanation. I'll handle this.

R/pkgdown.R
jennybc: OK I'm about to push a commit deleting more code.

PR #1505, opened 2021-10-02
Enable "internal" repositories in `use_github()`

### LINKED ISSUE #1458: feature request: allowe `use_github()` to create "internal" repositories
First of all, many thanks for this awesome package. I bumped into this use case a number of times and, as far as I can tell, this feature request doesn't seem to have been discussed previously. Apologies if it was and a conclusion not to implement `visibility` had been reached.

Currently the only option when using `use_github()` is to create either a `public` or `private` repository. For enterprise accounts, there often is a 3rd use case, namely to create an `internal` repo. This is `private` as far as the rest of the world is concerned, but `public` from the point of view of the members of the organisation.  

I think it might be a case of adding an argument named `visibility` with the possible values `"public"`, `"private"` or `"internal"` (keeping in line with the [GitHub POST request parameters](https://docs.github.com/en/rest/reference/repos#create-an-organization-repository--parameters)). The default value for this argument could be `"public"` or `NULL`. Moreover, as far as I understand, the [GitHub docs](https://docs.github.com/en/rest/reference/repos) state `visibility` overrides `private` in case of conflict. What do you think?

The expected behaviour would be:

```r
# creates an "internal" repo
use_github(organisation =  "My_awesome_organisation", private = FALSE, visibility = "internal")

# creates an "internal" repo as visibility overrides "private"
use_github(organisation =  "My_awesome_organisation", private = TRUE, visibility = "internal")

# use_github should error when organisation = NULL and visibility = "internal"
use_github(visibility = "internal")
```

jennybc:
Closes #1458

In my hands, this does not seem to work and I don't know why.

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -25,6 +25,10 @@
 #'   such that you have permission to create repositories in this
 #'   `organisation`.
 #' @param private If `TRUE`, creates a private repository.
+#' @param visibility Only relevant for organisation-owned repos associated with
+#'   certain GitHub Enterprise products. The special "internal" `visibility`
+#'   grants read permission to all organisation members, i.e. it's intermediate
+#'   between "private" and "public" (within GHE).
 #' @inheritParams git_protocol
 #' @param host GitHub host to target, passed to the `.api_url` argument of
 #'   [gh::gh()]. If unspecified, gh defaults to "https://api.github.com",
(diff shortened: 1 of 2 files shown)


R/github.R
```diff
@@ -116,7 +138,10 @@ use_github <- function(organisation = NULL,
       org = organisation,
       name = repo_name,
       description = repo_desc,
-      private = private,
+      visibility = visibility,
+      # this is necessary to set `visibility` in GHE 2.22 (but not in 3.2)
+      # hopefully it's harmless when not needed
+      .send_headers = c(Accept = "application/vnd.github.nebula-preview+json"),
```
jennybc, line 144: I should be using `.accept` here. Either that argument is new (to me) or I just forgot it existed.

PR #1506, opened 2021-10-05
Include version number in commit message

### LINKED ISSUE #1452: Include version number in commit message of `use_version()` and variants?
This would create a more informative history. For instance, here is the vdiffr history right now:

```
a525fcf Increment version number
5f775bd Increment version number
```

Could it be something like this instead?

```
a525fcf Increment version number to 1.0.0.9000
5f775bd Increment version number to 1.0.0
```

I often need to check the commit contents to check if I bumped to the right version or which version is current.

malcolmbarrett:
Closes #1452

### DIFF
--- a/R/version.R
+++ b/R/version.R
@@ -57,7 +57,7 @@ use_version <- function(which = NULL) {
   use_c_version(new_ver)
 
   git_ask_commit(
-    "Increment version number",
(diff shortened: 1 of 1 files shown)


jennybc:
Thanks!

PR #1507, opened 2021-10-05
Don't include CRAN install instructions to start

### LINKED ISSUE #1451: Default README mentioning installing from CRAN is potentially confusing for new package authors
The default README issue template mentions the package can be installed from CRAN.
This can be confusing to first time package authors, and also users who encounter the readme without having the package actually submitted and approved.

It makes it seem like the package is (somehow) already on CRAN, just by adding the readme. Particularly if this step was done in combination with doing other things, like setting up CI, a package author could get a mistaken impression about the CRAN status of their package.

It will always be wrong for packages which are never submitted to CRAN (which may very well be the majority). It will also be incorrect at least temporarily when a package is first being developed and not yet on CRAN.

The alternative (just showing the devel installation instructions) is also problematic if the instructions are not changed when the package is submitted to CRAN.

But I think generally the authors who submit to CRAN are more capable of changing the installation instructions appropriately than first time R package authors many of whom many never submit a package to CRAN.

malcolmbarrett:
This closes #1451 by tweaking the default README instructions. We don't need to add a release bullet point reminding to tweak install instructions because, as it turns out, there's been one for a [few years already](https://github.com/r-lib/usethis/commit/87f743abdc76193ebc9a3c2009bfd680c0605034) (😆)

The downside of this simple change is that, if the repo is not on GitHub, nothing gets added to the installation section, and it will be just the header. We could only include the header if already on GH or we could leave it empty as a prompt. 

I think there's a lot of people (🤚) who create the README prior to putting on GH, as well.

### DIFF
--- a/inst/templates/package-README
+++ b/inst/templates/package-README
@@ -24,20 +24,15 @@ The goal of {{{ Package }}} is to ...
 
 ## Installation
 
-You can install the released version of {{{ Package }}} from [CRAN](https://CRAN.R-project.org) with:
-
-``` r
-install.packages("{{{ Package }}}")
-```
-
 {{#on_github}}
-And the development version from [GitHub](https://github.com/) with:
+You can install the the development version of {{{ Package }}} from [GitHub](https://github.com/) with:
 
 ``` r
 # install.packages("devtools")
 devtools::install_github("{{{ github_spec }}}")
 ```
 {{/on_github}}
+
 ## Example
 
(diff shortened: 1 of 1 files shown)


malcolmbarrett:
Oh wait, that's the checklist item. That makes sense

jennybc:
I brought `use_readme()` along. I think this is ready to merge @malcolmbarrett but would appreciate a second opinion.

PR #1509, opened 2021-10-06
Add `use_github_file()` and use it in `use_github_action*()`

### LINKED ISSUE #1407: Generalizing `use_github_action()` with `use_github_file()`
I'm very interested in the pattern used by `use_github_action()` as an alternative to `use_template()` when it makes sense to host files somewhere else (e.g. we want to make it easier to grab a file from R but don't necessarily want to host it in the package itself because it's useful for other purposes). 

I think it would be nice to refactor that internal code to a new exported function, e.g. `use_github_file()`, which could then be called by `use_github_action()`. The pattern for reading a file from GitHub is, thankfully, not that complex, but it's very useful. 

Happy to prepare a PR if it seems like it's worth exporting that code

### LINKED ISSUE #1483: Follow symlinks when grabbing GHA workflows
https://docs.github.com/en/rest/reference/repos#if-the-content-is-a-symlink

If we did this instead of naive URL/file grab, @jimhester could go back to using symlinks in https://github.com/r-lib/actions

jennybc:
Closes #1407, closes #1483

### DIFF
--- a/R/use_github_file.R
+++ b/R/use_github_file.R
@@ -31,12 +31,13 @@
 #' @examples
 #' \dontrun{
 #' use_github_file(
-#'   "https://github.com/r-lib/actions/blob/HEAD/examples/check-standard.yaml"
+#'   "https://github.com/r-lib/actions/blob/v1/examples/check-standard.yaml"
 #' )
 #'
 #' use_github_file(
 #'   "r-lib/actions",
 #'   path = "examples/check-standard.yaml",
+#'   ref = "v1",
 #'   save_as = ".github/workflows/R-CMD-check.yaml"
 #' )
 #' }
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -110,7 +110,7 @@ use_tidy_github_actions <- function() {
 #' Configure an individual, specific [GitHub
 #' Actions](https://github.com/features/actions) workflow, either one of the
 #' examples from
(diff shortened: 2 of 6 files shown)


jennybc:
@jimhester I haven't wired this up to the GHA functions yet, but that's next. If you have any observations, please share. Or wait 'til this is more done.

jennybc:
@malcolmbarrett Is this what you had in mind?

R/use_github_file.R
```diff
@@ -0,0 +1,130 @@
...
+
+  if (!dat$host %in% c("raw.githubusercontent.com", "github.com") ||
+      !nzchar(dat$fragment) ||
+      (dat$host == "github.com" && !grepl("^/blob/", dat$fragment))) {
+    ui_stop("Can't parse the URL provided via {ui_code('repo_spec')}.")
+  }
+  out$parsed <- TRUE
+
+  dat$fragment <- sub("^/(blob/)?", "", dat$fragment)
+  dat_fragment <- re_match(dat$fragment, "^(?<ref>.+?)/(?<path>.+)$")
```
jimhester, line 122: I think this would be better written as something like

```r
"^(?<ref>[^/]+)/(?<path>.+)$"
```

In general I try to avoid lazy quantifiers when writing regular expressions, as they can cause catastrophic backtracking.

malcolmbarrett:
Looks and works great so far! Presumably, you'll refactor `use_github_action()` to use `use_github_file()` internally?

PR #1516, opened 2021-10-13
Add `use_ragg()`

### LINKED ISSUE #1515: `use_ragg()`
Now that you can set ragg as your default graphics device in RStudio, I find myself recommending it to students because it's such a clear improvement to plot output, even on non-Windows machines. It would be nice to have something like `use_ragg()` that I can hand off to students much in the way I hand off `use_blank_slate()`.

Alternatively, there are also some options you can set in your R session: https://github.com/rstudio/rstudio/issues/9275

Supporting specific RStudio preferences is a slippery slope, of course. I do feel that ragg is an unusually strong improvement to the default (similar to `use_blank_slate()`) that merits considering a specific function in usethis, but @ddsjoberg's [rstudio.prefs](https://github.com/ddsjoberg/rstudio.prefs) package also works well as a generalized solution to managing preferences programmatically.

malcolmbarrett:
Closes #1515 

This PR adds `use_ragg()` to set the user's default graphics backend in RStudio to use ragg.

Currently, there's no project-level support for ragg--just a global option. I went the simple route and just supported the global level; however, we could also offer guidance at the project level to set `options(RStudioGD.backend = "ragg")` somewhere in the project. That would be more consistent with other config-setting functions in usethis, but I'm not sure how useful it is. https://github.com/rstudio/rstudio/issues/9275 suggests there might be some use cases for such a strategy, though

### DIFF
--- a/tests/testthat/test-rstudio.R
+++ b/tests/testthat/test-rstudio.R
@@ -97,7 +97,10 @@ test_that("use_ragg() modifies user-level RStudio prefs", {
   path <- withr::local_tempdir()
   withr::local_envvar(c("XDG_CONFIG_HOME" = path))
 
-  use_ragg()
+  with_mock(
+    check_installed = function(pkg) TRUE,
+    use_ragg()
+  )
 
   prefs <- rstudio_prefs_read()
   expect_equal(prefs[["graphics_backend"]], "ragg")


jennybc:
The heart of this PR is:

``` r
use_rstudio_config(list(graphics_backend = "ragg"))
```

So maybe the real solution is to export something for configuring RStudio (e.g. `use_rstudio_config()`). Then people can publicize short code snippets to achieve configurations they highly recommend without us opening ourselves up to a large proliferation of such functions.

jennybc:
> But if I start recommending that, there's a little bit of why not just this instead of `use_blank_slate()` ...

This is a valid point. It's because `use_blank_slate()` appeared *before* RStudio prefs were editable the way they are now. I'm not sure we'd create that function today, although I do like that it bundles all 3 settings together.

But maybe usethis should expose some general RStudio preference tooling for the more typical, simple "set-one-thing" case.

PR #1527, opened 2021-10-21
Only remove trailing slashes in `tidyverse_url()`

### LINKED ISSUE #1526: `use_pkgdown_github_pages()` causes warning in urlchecker
And on CRAN checks.

I believe it's because the trailing `/` is removed, as the corrected URL has the trailing `/`:

https://github.com/r-lib/usethis/blob/196b85a59e152b8c47e20cbd6e065de68f92839f/R/pkgdown.R#L77

However, this is a peculiar thing to find in the code, which makes me think it fixes some other problem.

Here's the result from [tidysmd](https://github.com/malcolmbarrett/tidysmd)

``` r
urlchecker::url_check()
#> fetching [ 0 / 8 ]fetching [ 1 / 8 ]fetching [ 2 / 8 ]fetching [ 3 / 8 ]fetching [ 4 / 8 ]fetching [ 5 / 8 ]fetching [ 6 / 8 ]fetching [ 7 / 8 ]                       
#> x Error: README.md:12:56 404: Not Found
#> status](https://www.r-pkg.org/badges/version/tidysmd)](https://CRAN.R-project.org/package=tidysmd)
#>                                                        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#> ! Warning: DESCRIPTION:12:5 Moved
#>     https://malcolmbarrett.github.io/tidysmd
#>     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#>     https://malcolmbarrett.github.io/tidysmd/
```

<sup>Created on 2021-10-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

malcolmbarrett:
Closes #1526

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+pkgdown-related functions no longer automatically strip a trailing slash from the pkgdown site URL, in order to play more nicely with CRAN's URL checks (#1526).
+
 `edit_pkgdown_config()` is a new function that opens the pkgdown YAML configuration file for the current Project, if such a file exists.
 
 The error thrown when reporting an unsupported GitHub configuration has been fixed for forward compatibility with a future version of rlang, i.e. what is anticipated to be rlang v1.0.0.
--- a/R/code-of-conduct.R
+++ b/R/code-of-conduct.R
@@ -42,6 +42,7 @@ use_code_of_conduct <- function(contact, path = NULL) {
 
   href <- pkgdown_url(pedantic = TRUE) %||%
     "https://contributor-covenant.org/version/2/0"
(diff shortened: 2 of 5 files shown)


jennybc:
I think it's clear our default stance should be to leave pkgdown URL's alone, i.e. to NOT be eager to strip a trailing slash. So I've also changed `pkgdown_url()`. 

It looks like the only time we want to form a URL without the trailing slash is in cases like `https://PKG.tidyverse.org`. We handle that when first recording such a pkgdown URL.

In general, now we take pkgdown URLs as we find them.

malcolmbarrett:
Sounds good to me

Issue #1541, opened 2021-12-06
Versioning GH actions


gaborcsardi:
It seems like usethis always uses the `v1` tag of the actions repo, which was fine until now, but I am about to update the `setup-r-dependencies` action and the related example workflows that use it in a way that might not be backwards compatible. 

So I would create a new branch for this (or just use `main`?). For a short time this branch would live as the "devel" workflow, and then we will make it the default, and then `r-lib/actions` will receive a `v2` tag. 

So it seems like it would make sense to support multiple tags/branches in usethis as well. usethis could  pick up the latest release / tag from the `r-lib/actions` repo, this could be the default.

jennybc:
How do you envision usethis discovering the default ref from `r-lib/actions`?

This will be a fairly easy adjustment, since all the workflow-downloading functions now route through `use_github_file()`, which already has a `ref` argument.

gaborcsardi:
usethis could pick up the latest release / tag from the `r-lib/actions` repo.

jennybc:
Yeah I meant ... latest in terms of the date modified? Or latest in an alpha/numerical sense?

gaborcsardi:
Sorry, yes, we would order the numbers. I don't think we would have version numbers, so just removing the `v` and then ordering the rest numerically seems good?

jennybc:
OK, sounds good.

jennybc:
I'm beginning to question defaulting to the latest tag in `r-lib/actions`, in a version number sense.

Because, for example, right now that might deliver the not-yet-quite-official `v2` to people who probably want `v1`. And you can imagine the same cycle repeating itself in the future whenever there's a need to start experimenting with `v3`.

@gaborcsardi Do you think there should be some other mechanism (tag, probably) for designating the version that's currently considered the best for more users?

gaborcsardi:
Aren't pre-releases marked in the API?

gaborcsardi:
```
gh::gh("/repos/r-lib/actions/releases")
```

```
...
    "node_id": "RE_kwDODMSrZM4DSp0x",
    "tag_name": "v1",
    "target_commitish": "master",
    "name": "v1",
    "draft": false,
    "prerelease": false,
...
    "node_id": "RE_kwDODMSrZM4DSVjI",
    "tag_name": "v2",
    "target_commitish": "master",
    "name": "v2 ",
    "draft": false,
    "prerelease": true,
...
```

So we can choose the latest release that is not a pre-release.

[...]

PR #1557, opened 2022-01-03
Quick hack to use latest actions

hadley:
We need to put a little thought into a better solution. At a minimum, version should probably become a parameter.


gaborcsardi:
Yeah, we should probably do this now and eventually implement #1541

jennybc:
OK I will put this in place as a stopgap until I do #1541.

PR #1558, opened 2022-01-05
fix github pages branch name

kevinushey:
As the branch is normally called `gh-pages`, not `github-pages`.


jennybc:
Oops. Thanks!

PR #1567, opened 2022-01-14
Consult ref for GHA workflows

### LINKED ISSUE #1541: Versioning GH actions

jennybc:
Closes #1541

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -261,3 +261,21 @@ check_uses_github_actions <- function() {
     uses GitHub Actions.
     Do you need to run {ui_code('use_github_actions()')}?")
 }
+
+latest_release <- function(repo_spec = "https://github.com/r-lib/actions") {
+  parsed <- parse_repo_url(repo_spec)
+  tr <- list(
+    repo_owner = spec_owner(parsed$repo_spec),
+    repo_name = spec_repo(parsed$repo_spec),
+    api_url = parsed$api_url %||% "https://api.github.com"
+  )
+  gh <- gh_tr(tr)
+  # https://docs.github.com/en/rest/reference/releases#list-releases
+  raw_releases <- gh("/repos/{owner}/{repo}/releases", .limit = Inf)
+  releases <- purrr::discard(
+    map_chr(raw_releases, "tag_name"),
+    map_lgl(raw_releases, "prerelease")
+  )
+  # expected tag_names look like v1, v2, etc.
+  releases[which.max(as.integer(gsub("[^0-9]", "", releases)))]
+}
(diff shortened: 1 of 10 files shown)


jennybc:
I haven't done anything about incorporating the SHA into the copied workflow file. I think I'll just wait and see if we ever think that would be terribly useful.

R/github-actions.R
```diff
@@ -259,3 +264,39 @@ use_tidy_github_actions <- function() {
...
+    api_url = parsed$api_url %||% "https://api.github.com"
+  )
+  gh <- gh_tr(tr)
+  # https://docs.github.com/en/rest/reference/releases#list-releases
+  raw_releases <- gh("/repos/{owner}/{repo}/releases", .limit = Inf)
+  releases <- purrr::discard(
+    map_chr(raw_releases, "tag_name"),
+    map_lgl(raw_releases, "prerelease")
+  )
+  # expected tag_names look like v1, v2, etc.
```
gaborcsardi, line 300: I wonder if we should make this a bit more robust and handle the `v2.0` and `v2.0.0` etc. forms as well, just in case, we forget about this contract and use them later. Something like 
```r
vers <- package_version(gsub("[^0-9]", "", releases))
```
can order the versions and select the max. It does not have a `which.max` method, though.

gaborcsardi (commented):
Looks great! Left a comment about handling more general version numbers.

R/github-actions.R
jennybc: Good idea. I'll do this before I merge.

jennybc: Can I assume you intend to handle versions like this?

https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management

If so, if I both `v2.0.1` and `v2` are published releases, I think my `latest_release()` helper should return `v2`, yeah?

gaborcsardi: Good question. Currently `v2` is always the latest, and there are no _real_ `vx.y.z` releases, so this is quite hypothetical. So I don't mind it you choose `v2.0.1` or `v2` in this case. Whichever is easier to implement, maybe? I think as long as we don't error it is fine.

jennybc: OK I'm going to proceed according to this take on how semver and "using tags for GHA release management" intersect:

Once there's any release tag with a major number X, vX should point at the latest released version in the X.?.? series.

jennybc: In the end I just went with `max()` release, because it's not worth the trouble to do something more complicated for a situation that does not exist yet and it is not expected to exist soon.

PR #1570, opened 2022-01-19
In release checklist, be explicit about when to push/pull

hadley:
It's easy to forget these, so we might as will be explicit.

cc @DavisVaughan


tests/testthat/_snaps/release.md
```diff
@@ -30,6 +32,7 @@
       Wait for CRAN...
       
       * [ ] Accepted :tada:
+      * [ ] `git push`
       * [ ] `usethis::use_github_release()`
       * [ ] `usethis::use_dev_version()`
       * [ ] `usethis::use_news_md()`
```
DavisVaughan, line 38: Is this the right placement? Or after `use_news_md()`? Just struck me as odd

jennybc: Yeah I think it makes sense.

This `usethis::use_news_md()` item only applies to a very specific situation: right after your first CRAN release, if you didn't have any NEWS before.

DavisVaughan: `use_dev_version()` requires a push though, right?

jennybc: I think the actual point is: do we want *another* `git push` at the end? I.e. after all the bullet points that are basically starting a new dev version.

hadley: I just came here to say that too 😄

PR #1577, opened 2022-01-28
Error with abs paths in `proj_path()`

malcolmbarrett:
`proj_path()` always returns paths with the project root prepended to it, even if the path is already absolute. This is less of a big deal with using `proj_path()` directly (since it's obviously intended to do that) and more when we use it internally to process paths, as in `use_github_file()`.  (I recently tried to save a file using an absolute path interactively, which did not work)

This PR modifies `proj_path()` to ~~respect~~ reject absolute paths

### DIFF
--- a/tests/testthat/test-proj.R
+++ b/tests/testthat/test-proj.R
@@ -48,11 +48,11 @@ test_that("proj_path() appends to the project path", {
   expect_identical(proj_path("a", "b", "c"), proj_path("a/b/c"))
 })
 
-test_that("proj_path() respects absolute paths", {
+test_that("proj_path() errors with absolute paths", {
   create_local_project()
-  expect_equal(
+  expect_error(
     proj_path(c("/a", "b", "/c")),
-    path(c("/a", path(proj_get(), "b"), "/c"))
+    "Paths must be relative to the active project"
   )
 })
 
--- a/tests/testthat/test-proj.R
+++ b/tests/testthat/test-proj.R
@@ -50,10 +50,7 @@ test_that("proj_path() appends to the project path", {
 
 test_that("proj_path() errors with absolute paths", {
   create_local_project()
-  expect_error(
-    proj_path(c("/a", "b", "/c")),
-    "Paths must be relative to the active project"
-  )
+  expect_snapshot(proj_path(c("/a", "b", "/c")), error = TRUE)
 })
 
 test_that("proj_rel_path() returns path part below the project", {
--- a/tests/testthat/test-proj.R
+++ b/tests/testthat/test-proj.R
@@ -48,6 +48,14 @@ test_that("proj_path() appends to the project path", {
   expect_identical(proj_path("a", "b", "c"), proj_path("a/b/c"))
 })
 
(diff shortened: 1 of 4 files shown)


jennybc:
I can get on board with doing *something* here. It does seem wrong to mechanically shove the project path onto the front end of anything.

But `proj_path()` is designed to do a very specific job. I can convince myself that the right thing to do is **error** for an absolute path.

In terms of what you were doing with `use_github_file()`, I suspect there are other "off-label" uses of various usethis path and/or project functions that would achieve whatever your goal was.

malcolmbarrett:
This is pretty consistent with what `here()` does, which strikes me as an analogous sort of function

jennybc:
I think `proj_path()` is explicitly _not_ `here::here()` though. From the current docs:

> In general, end user scripts should not contain direct calls to `usethis::proj_*()` utility functions. They are internal functions that are exported for occasional interactive use or use in packages that extend usethis. End user code should call functions in [rprojroot](https://rprojroot.r-lib.org/) or its simpler companion, [here](https://here.r-lib.org/), to programmatically detect a project and build paths within it.

The only reason `proj_path()` is even exported is to help someone developing a custom usethis for their organization.

jennybc:
> but if you think throwing an error is better in this case, I can refactor it to do that. I'll trust you!

Yeah, I really do think it's best that, if we're going to introduce handling for absolute paths, that we reject them in `proj_path()`. It looks like fs has a function we can use for detection: [`fs::is_absolute_path()`](https://fs.r-lib.org/reference/is_absolute_path.html).

> One thing, though: if we firmly believe these should be relative paths, I think we should mention that a bit more in the arguments that then get processed by `proj_path()`. (I'm saying this without looking at them so maybe they're clearer than I remember)

You're welcome to propose a change to the docs. I do feel like the intent of the function is already pretty well documented:

> `proj_path()`: Builds a path within the active project returned by `proj_get()`. Thin wrapper around [fs::path()](https://fs.r-lib.org/reference/path.html).

I guess we could make it even clearer by showing the approximate code, i.e. `fs::path(proj_get(), ...)`.

malcolmbarrett:
Sorry, I was mistaken about the docs! I was thinking about the arguments that were getting processed internally by `proj_path()`, but I could only find one (not even the one that set me out on this PR) that didn't explicitly mention it needed to be relative. Everything else did. I'll blame it on new dad brain 😵

tests/testthat/test-proj.R
```diff
@@ -48,6 +48,14 @@ test_that("proj_path() appends to the project path", {
   expect_identical(proj_path("a", "b", "c"), proj_path("a/b/c"))
 })
 
+test_that("proj_path() errors with absolute paths", {
+  create_local_project()
+  expect_error(
```
jennybc, line 53: I think we favor snapshot tests over `expect_error(regexp =)` these days.

I'll leave it up to you whether to change.

jennybc (approved):

malcolmbarrett:
Thanks!

PR #1583, opened 2022-02-11
missing `glue()` to replace correct `ref`

cderv:
Small fix so that `{ref}` gets correctly replaced. 

Otherwise we get 
````
* Learn more at <https://github.com/r-lib/actions/blob/{ref}/examples/README.md>.
````


malcolmbarrett:
Duplicate of #1576

cderv:
Oups sorry. I checked if it was fixed in main but did not thought it would still be an opened PR and not yet fixed.

PR #1593, opened 2022-02-28
Add upkeep bullets for 2022

tracykteal:
Updating tidy_upkeep checklist for 2022

### DIFF
--- a/R/tidy-upkeep.R
+++ b/R/tidy-upkeep.R
@@ -91,10 +91,10 @@ if (year <= 2022) {
      "",
      todo("`usethis::use_tidy_coc()`"),
      todo("Update errors to rlang 1.0.0"),
-     todo("Update pkgdown site using instructions at [http://tidytemplate.tidyverse.org/](http://tidytemplate.tidyverse.org/)"),
-     todo("Ensure `dev:auto` in pkgdown"),
-     todo("Reploy released site using [r-lib/pkgdown#2051](https://github.com/r-lib/pkgdown/pull/2051)"),
-     todo("Handle and close any still-open master to main issues"),
+     todo("Update pkgdown site using instructions at <https://tidytemplate.tidyverse.org>"),
+     todo("Re-publish released site using [r-lib/pkgdown#2051](https://github.com/r-lib/pkgdown/pull/2051)"),
+     todo("Ensure pkgdown `development` is `mode: auto` in pkgdown config"),
+     todo("Handle and close any still-open `master` --> `main` issues"),
      todo("Update README badges"),
      ""
     )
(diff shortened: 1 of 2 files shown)


R/tidy-upkeep.R
```diff
@@ -85,6 +85,20 @@ upkeep_checklist <- function(year = NULL) {
       ""
     )
   }
+if (year <= 2022) {
+    bullets <- c(bullets,
+     "2022",
+     "",
+     todo("`usethis::use_tidy_coc()`"),
+     todo("Update errors to rlang 1.0.0"),
```
jennybc, line 93: Is there a guide for this?

R/tidy-upkeep.R
```diff
@@ -85,6 +85,20 @@ upkeep_checklist <- function(year = NULL) {
...
+     todo("Update pkgdown site using instructions at <https://tidytemplate.tidyverse.org>"),
+     todo("Re-publish released site using [r-lib/pkgdown#2051](https://github.com/r-lib/pkgdown/pull/2051)"),
```
jennybc, line 95: Obviously we'll change this once the PR is merged

R/tidy-upkeep.R
```diff
@@ -85,6 +85,20 @@ upkeep_checklist <- function(year = NULL) {
...
+     todo("Ensure pkgdown `development` is `mode: auto` in pkgdown config"),
```
jennybc, line 96: FWIW this is included in the proposed pkgdown PR re: updating a released site.

R/tidy-upkeep.R
```diff
@@ -85,6 +85,20 @@ upkeep_checklist <- function(year = NULL) {
...
+     todo("Handle and close any still-open `master` --> `main` issues"),
+     todo("Update README badges"),
```
jennybc, line 98: We don't really say what this means or how to do it. But it's also very fiddly for little pay-off, so I'm not sure what to do about this.

jennybc:
I've noted some loose ends but will merge anyway, to get the ball rolling.

R/tidy-upkeep.R
tracykteal: I think @lionel- has this in progress, so we'll update once it's available

R/tidy-upkeep.R
tracykteal: should we take it out of the list then?

R/tidy-upkeep.R
lionel-: There's https://rlang.r-lib.org/reference/topic-error-call.html and a guide about error chaining coming up.

lionel-: Here is the topic about error chaining: https://rlang.r-lib.org/reference/topic-error-chaining.html
And the one about cli formatting: https://rlang.r-lib.org/reference/topic-condition-formatting.html

jennybc: OK added.

PR #1595, opened 2022-03-01
updating language around README in code-of-conduct.R

tracykteal:
Changing the language in code-of-conduct.R from requiring ("don't forget) to suggesting ("you may also want to"). Since the Code of Conduct is linked in other locations, it's not required that it go in the README, but some projects may still want to include it, especially if they don't have other places, like pkgdown sites, where it's as documented.


jennybc:
Thanks!

PR #1601, opened 2022-03-03
Account for more ways that RStudio can appear in pkg author's (or not appear)

### LINKED ISSUE #1600: `usethis:::is_rstudio_funded()` could returns NA which error in `upkeep_checklist()`
Discovered while trying to use ` usethis::use_tidy_upkeep_issue()`

````r
> usethis::use_tidy_upkeep_issue()
√ Setting active project to 'C:/Users/chris/Documents/DEV_R/rticles'
Which repo should we target? 

1: cderv/rticles = 'origin'
2: rstudio/rticles = 'upstream'

Sélection : 2
Error in if (cond) { : missing value where TRUE/FALSE needed
````

This is due to 
https://github.com/r-lib/usethis/blob/b8b0209425a26d8a439617e0dba5c54e31f40264/R/tidy-upkeep.R#L85

because 
``` r
usethis:::is_rstudio_funded()
#> [1] NA
```

I believe it should TRUE or FALSE only to be used in a `if()` condition

This is because in
https://github.com/r-lib/usethis/blob/a545714f8f94e916bbd52e007092b149dda1db79/R/release.R#L380-L381

`.x$given` is NULL. and can't be compared using `==`. Using `identical()` would be working

cderv:
Fixes #1600 

in `person()`,  `given` field can be NULL which does not play well with `==` comparaison. In that case `is_rstudio_funded()` won't return TRUE or FALSE, but NA. 

Using `identical()` fix the issue.

I can add a NEWS bullet for this if needed.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -369,16 +369,58 @@ news_latest <- function(lines) {
 }
 
 is_rstudio_pkg <- function() {
-  is_rstudio_funded() || is_in_rstudio_org()
+  is_rstudio_cph_or_fnd() || is_in_rstudio_org()
 }
 
-is_rstudio_funded <- function() {
+is_rstudio_cph_or_fnd <- function() {
   if (!is_package()) {
     return(FALSE)
   }
+  roles <- get_rstudio_roles()
+  "cph" %in% roles || "fnd" %in% roles
+}
+
+is_rstudio_person_canonical <- function() {
+  if (!is_package()) {
+    return(FALSE)
+  }
+  roles <- get_rstudio_roles()
+  length(roles) > 0 &&
+    "fnd" %in% roles &&
+    "cph" %in% roles &&
+    attr(roles, "appears_in", exact = TRUE) == "given" &&
+    attr(roles, "appears_as", exact = TRUE) == "RStudio"
+}
+
+get_rstudio_roles <- function() {
+  if (!is_package()) {
+    return()
+  }
+
   desc <- desc::desc(file = proj_get())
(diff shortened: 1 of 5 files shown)


jennybc:
Should we do even more, in terms of catching sub-optimal specifications that RStudio is the funder? Like the thing you mentioned about some packages using `family`?

cderv:
Yes we could do that if we think that using `given` is better. rmarkdown and probably other packages are using `family` when adding organisation in DESCRIPTION. Until now, I thought that was the way to go for adding RStudio in DESCRIPTION until I just discover that tidyverse practice is to use `given`. 

Other difference in rmarkdown package, `"RStudio, PBC"` is used and not just RStudio.

If we want to be the same across all RStudio funded package (which would make sense), maybe a small analysis of all the DESCRIPTION could be useful to see the practice. Do we have a complete list of all RStudio funded packages ?

jennybc:
I definitely have scripts from my `master` --> `main` adventure that can generate lots of info for us. I'll report back here when I've done some analysis.

jennybc:
OK, I think we can cope with almost everything now.

Once I did the analysis, it wasn't that much harder to account for all the ways people credit RStudio in DESCRIPTION (or fail to).

Now the upkeep issue will nudge the maintainer to add or update RStudio's `person()` if it's missing, incomplete, or "creative". I can imagine more spring cleaning exercises in the future, across open source teams, so this will be useful.

cderv:
Thanks a lot! 

I am definitely for the spring cleaning exercise, and that is why I tried to apply already.

PR #1612, opened 2022-03-18
Unexported tooling to setup VS Code debugging config

jennybc:


lionel- (commented):
Looks useful!

Could this be configured to take an optional `pid`, in which case an `"attach"` request would be used instead of `"launch"`?

Also if ESS is detected, could you use `"codelldb"` instead of `"lldb"` please? This is for historical reasons (the Emacs module patches it up to `"lldb"` eventually).

lionel-:
> Also if ESS is detected, could you use "codelldb" instead of "lldb" please? This is for historical reasons (the Emacs module patches it up to "lldb" eventually).

oops please disregard this request. I had forgotten that I decided the vscode UI is much more practical than the Emacs one. Also the Emacs DAP client expects to find the `launch.json` file in the project root so supporting that would be a bit more work.

I would still find it useful to be able to set up an attach request though. This way I can interact with R from ESS while debugging with vscode.

jennybc:
Yeah I can add an attach configuration. There is one in @DavisVaughan's blog post.

inst/templates/vscode-c_cpp_properties.json
```diff
@@ -0,0 +1,21 @@
...
+        "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include",
+        "/usr/local/include"
+      ],
+      "macFrameworkPath": [
+        "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks"
+      ],
+      "compilerPath": "/usr/bin/clang",
+      "cStandard": "c99",
+      "cppStandard": "c++11",
+      "intelliSenseMode": "macos-clang-x64"
```
DavisVaughan, line 17: I _think_ this comes automatically with the compiler path so it probably doesn't need to be set.

I noticed that sometimes when I change the compiler in the actual UI for this properties file, if I change from clang->gcc it gives me a note that it automatically updated the intelliSenseMode

inst/templates/vscode-launch.json
```diff
@@ -0,0 +1,21 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "(lldb) Launch R  ",
```
DavisVaughan, line 5:
```suggestion
      "name": "(lldb) Launch R",
```

inst/templates/vscode-launch.json
```diff
@@ -0,0 +1,21 @@
...
+        "--vanilla",
+        "-e",
+        "source('debug/debug.R')"
+      ],
+      "env": {
+        "R_HOME" : "/Library/Frameworks/R.framework/Resources"
+      },
+      "terminal": "console",
+      "stopOnEntry": false
+    }
```
DavisVaughan, line 19: Agree with @lionel- about including an attach configuration, so like

```
    {
      "name": "(lldb) Attach to R",
      "type": "lldb",
      "request": "attach",
      "pid": "${command:pickMyProcess}",
      "stopOnEntry": false
    }
```

`pickMyProcess` pops up this search bar where in theory you can start typing and it will filter down to the process of interest, but we just type in....R, so that doesn't filter out very much (because it partially matches anything with an R in it)

There is also an _input variable_ option, so in theory we could use something like this
https://github.com/vadimcn/vscode-lldb/issues/169#issuecomment-473622959

That is basically all we'd need because I'm just pasting in the process id that `Sys.getpid()` gives me, but the copy/paste also works with `pickMyProcess`, so I guess we can just keep it as that until we see a need to change it.

inst/templates/vscode-c_cpp_properties.json
jennybc: I haven't set this on purpose, but it came from one of my actual config files. So yeah I guess something I did in the UI set it. I can remove and see if this tends to eventually get added by VS Code.

jennybc:
OK the launch config is added.
I think this is ready to merge.

DavisVaughan:
LGTM

inst/templates/vscode-launch.json
lionel-: Note that I'd be calling the usethis function from within Emacs where the R process is running, so `pickMyProcess` wouldn't work. It seems like a good default for vscode users who run the process within vscode though. This is why I was thinking about an optional `pid` argument for the usethis function so I could pass `Sys.getpid()`.

lionel-: Having multiple configs will cause vscode to prompt about which config to use right? If that's the case, it would also be nice if there was an option to disable configs. This way people who use only one of the configs would be able to skip the prompt.

jennybc: > This is why I was thinking about an optional pid argument for the usethis function so I could pass `Sys.getpid()`.

I am thinking of this function as just about initial setup. I assume the active launching / attaching is still done through some IDE-specific gesture. At least that's what happens in VS Code.

I haven't used the attach method yet, nor do I use Emacs today. So @lionel- you could always propose another function that's useful when you want to initiate debugging.

> Having multiple configs will cause vscode to prompt about which config to use right?

It seems like the top one is selected as default and one only has to click the ▶︎.

I'm wondering if you might just want a slight different helper, `use_dap_debug()` or similar, that caters to your Emacs workflow?

lionel-: I wondered about that as well but if "method" selection makes sense as part of this initial setup, then a `pid` argument is only a small change to add. Though for initial setup it's also easy to edit the file manually. I'd also be happy to use a different function for attaching to a live process.

By the way, setting a `pid` is not only for Emacs users, it's also for RStudio users or users of any other editors. The goal of this workflow is to keep interacting with R as usual and only use vscode for interacting with the debugger. I think it should also be possible to use it to attach to a crashing R process (when R prompts about exit method, e.g. with core dump), which seems generally useful.

DavisVaughan: > Having multiple configs will cause vscode to prompt about which config to use right?

You just choose between them in the debug window
Screen Shot 2022-03-21 at 9 22 19 AM

lionel-: That's what I thought, this makes `<f5>` a little bit less handy.

jennybc:
@lionel- I'm going to merge this for the simple reason that "it works for me" and I don't think I completely grasp your use case. I'm happy to help hammer out functions that make your workflow more pleasant. Either you can make a PR or just give me a sketch and I can translate it into usethis-y code.

PR #1620, opened 2022-04-06
add note to use_readme_rmd() docs

tjmahr:
I have run into a couple of times where I need to undo the git hook* and I don't know how to, so this is a brief note about the pre-commit hook in use_readme_rmd() behavior.

\* e.g., the README.Rmd reads in a file for demonstration. The file changes and README.Rmd is knitted to give README.md. README.Rmd is unchanged, README.md is changed, but README.md cannot be committed because the unchanged README.Rmd is not also staged.

### DIFF
--- a/R/readme.R
+++ b/R/readme.R
@@ -1,8 +1,8 @@
 #' Create README files
 #'
 #' @description
-#' Creates skeleton README files with sections for
-#' * a high-level description of the package and its goals
+#' Creates skeleton README files with possible stubs for
+#' * a high-level description of the project/package and its goals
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks!

> e.g., the README.Rmd reads in a file for demonstration. The file changes and README.Rmd is knitted to give README.md. README.Rmd is unchanged, README.md is changed, but README.md cannot be committed because the unchanged README.Rmd is not also staged.

FYI, you can still still commit in this situation by following the displayed instructions re: `git commit --no-verify -m <message>`. But I'm assuming you know this.

Some people really would rather remove the hook altogether, which is also fine.

Issue #1619, opened 2022-04-04
Function for specifying minimum R version


JosiahParry:
I am presently writing a package utilizing the native pipe. As such, I would like to specify the minimum R version as 4.1. I was hoping to be able to do this via usethis as I do not know how to format this in a DESCRIPTION file. 

I see that there is a tangential closed issue at https://github.com/r-lib/usethis/issues/631. My suggestion, however, is not to infer it from `use_data()` but rather be something stated explicitly by the developer themselves. It could, however, be inferred that R v4.1+ is necessary when the native pipe is used.

malcolmbarrett:
Sorry, I made it sound like a proposal rather than a solution. That code works out of box & is what I do, for better or worse

jennybc:
Yeah the `use_package()` approach shown by @malcolmbarrett is the intended way, if you want to use usethis to do this. It's not something that comes up nearly as often as the main `use_package()` usage.

JosiahParry:
Good to know! Thanks :)

JosiahParry:
In light of the new check for R 4.1 incompatibilities in R-devel, it would be quite nice to have to a way to set the minimum version of R. I can never quite remember how to do it and come back to this issue every few months or so.

malcolmbarrett:
Doesn't this solution still work?

JosiahParry:
It definitely still works! Its just exceptionally difficult to remember how to do it and then quite tough to find this issue 

<details>
<summary> 
My search history trying to get here
</summary>

<img width="910" alt="Image" src="https://github.com/user-attachments/assets/1c8d05e7-de42-4284-8e29-1c3a5dded1e8" />

</details>

PR #1621, opened 2022-04-06
Improve `use_package("R", ...)` experience and add example

malcolmbarrett:
Related to #1619, this PR smooths out the process of using `use_package()` to bump the R version dependency:
* Add an example to docs
* Don't advise against using `Depends` for `R`
* Allow `min_version = TRUE` to detect current R version 
* Require that `min_version` is set and that `type` is `Depends`

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -6,10 +6,12 @@ use_dependency <- function(package, type, min_version = NULL) {
     check_installed(package)
   }
 
-  if (isTRUE(min_version)) {
+  if (isTRUE(min_version) && package == "R") {
+    min_version <- glue("{R.Version()$major}.{R.Version()$minor}")
+  } else if (isTRUE(min_version)) {
     min_version <- utils::packageVersion(package)
   }
-  version <- if (is.null(min_version)) "*" else paste0(">= ", min_version)
+  version <- if (is.null(min_version)) "*" else glue(">= {min_version}")
 
   types <- c("Depends", "Imports", "Suggests", "Enhances", "LinkingTo")
   names(types) <- tolower(types)
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -7,11 +7,13 @@ use_dependency <- function(package, type, min_version = NULL) {
   }
 
   if (package == "R" && type != "Depends") {
-    ui_stop("Specify R version requirements in Depends instead of {type}")
+    ui_stop("Set {ui_code('type = \"Depends\"')} when specifying an R version")
+  } else if (package == "R" && is.null(min_version)) {
+    ui_stop("Specify {ui_code('min_version')} when {ui_code('package = \"R\"')}")
   }
 
   if (isTRUE(min_version) && package == "R") {
-    min_version <- glue("{R.Version()$major}.{R.Version()$minor}")
+    min_version <- r_version()
   } else if (isTRUE(min_version)) {
     min_version <- utils::packageVersion(package)
   }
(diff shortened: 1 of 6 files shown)


R/package.R
```diff
@@ -27,6 +27,9 @@
 #' use_package("ggplot2")
 #' use_package("dplyr", "suggests")
 #' use_dev_package("glue")
+#'
+#' # Depend on R version 4.1.0
+#' use_package("R", type = "Depends", min_version = "4.1.0")
```
jennybc, line 32:
```suggestion
#' # Depend on R version 4.1
#' use_package("R", type = "Depends", min_version = "4.1")
```

jennybc (commented):
Are you game to add a snapshot test, just so we've captured the user experience and how it differs from a regular package?

If we detect that the package is `"R"`, should we only accept `type = "Depends"`?

Just to record this from WRE for us to reference:

https://rstudio.github.io/r-manuals/r-exts/Creating-R-packages.html#package-dependencies


> The Depends field can also specify a dependence on a certain version of R — e.g., if the package works only with R version 4.0.0 or later, include R (>= 4.0) in the Depends field. (As here, trailing zeroes can be dropped and it is recommended that they are.) You can also require a certain SVN revision for R-devel or R-patched, e.g. R (>= 2.14.0), R (>= r56550) requires a version later than R-devel of late July 2011 (including released versions of 2.14.0).
>
> It makes no sense to declare a dependence on R without a version specification ....
>
> It is inadvisable to use a dependence on R with patchlevel (the third digit) other than zero. Doing so with packages which others depend on will cause the other packages to become unusable under earlier versions in the series, and e.g. versions 4.x.1 are widely used throughout the Northern Hemisphere academic year.

malcolmbarrett:
Thanks!

> Are you game to add a snapshot test, just so we've captured the user experience and how it differs from a regular package?

Will do

> If we detect that the package is "R", should we only accept type = "Depends"?

Yeah, thank makes sense.

R/helpers.R
```diff
@@ -68,6 +74,8 @@ use_dependency <- function(package, type, min_version = NULL) {
   invisible(TRUE)
 }
 
+r_version <- function() glue("{R.Version()$major}.{R.Version()$minor}")
```
jennybc, line 77: This will produce an R version with 3 components, whereas I think we want to default to 2 components. I.e. we want `4.1` instead of `4.1.2`. Despite the name `minor` appears to return `minor.patch`.

tests/testthat/test-package.R
```diff
@@ -14,6 +14,24 @@ test_that("use_package() guides new packages but not pre-existing ones", {
...
 
+test_that("use_package() handles R versions with aplomb", {
+  create_local_package()
+  withr::local_options(usethis.quiet = FALSE)
+  expect_snapshot(use_package("R"), error = TRUE)
+  expect_snapshot(use_package("R", type = "Depends"), error = TRUE)
+  expect_snapshot(use_package("R", type = "Depends", min_version = "3.6"))
+  expect_equal(subset(desc::desc_get_deps(), package == "R")$version, ">= 3.6")
+  with_mock(
+    r_version = function() glue("4.1"),
```
jennybc, line 25:
```suggestion
    r_version = function() "4.1",
```

R/helpers.R
malcolmbarrett: Oh, good catch. Thanks! Relatedly, do you think we embed CRAN's advice of only specifying to the minor level by inserting a check when the user supplies `min_version`, something like `check_r_version <- function(min_version) if length(strsplit(min_version, "\\.")[[1]]) != 2 {...}`?

jennybc: I think if someone provides the full version number, we assume they know what they're doing.

malcolmbarrett:
Thanks, this should be ready for a final sweep

PR #1627, opened 2022-04-13
Fix `use_r("")` and similar

malcolmbarrett:
Currently, `use_r("")` creates `R/.R`, which I don't think is ever what someone wants. This PR adds a check for `use_r()` and friends to make sure the file actually has a name.

### DIFF
--- a/R/r.R
+++ b/R/r.R
@@ -183,7 +183,6 @@ check_file_name <- function(name) {
   if (!is_string(name)) {
     ui_stop("Name must be a single string")
   }
-
   if (!valid_file_name(path_ext_remove(name))) {
     ui_stop(c(
       "{ui_value(name)} is not a valid file name. It should:",
--- a/R/r.R
+++ b/R/r.R
@@ -49,10 +49,10 @@
 #'   [R Packages](https://r-pkgs.org).
 #' @export
 use_r <- function(name = NULL, open = rlang::is_interactive()) {
+  check_not_empty_file_name(name)
   name <- name %||% get_active_r_file(path = "tests/testthat")
   name <- gsub("^test-", "", name)
   name <- slug(name, "R")
-  check_not_empty_file_name(name)
   check_file_name(name)
 
   use_directory("R")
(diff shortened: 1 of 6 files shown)


R/r.R
```diff
@@ -150,6 +150,9 @@ check_file_name <- function(name) {
   if (!is_string(name)) {
     ui_stop("Name must be a single string")
   }
+  if (path_file(name) %in% c(".R", "test-.R", ".cpp", ".c")) {
```
malcolmbarrett, line 153: This isn't scaleable but I figured our scope is limited and this is easier than regex, particularly since we're most likely adding an extension to `""`

jennybc: Have you been bitten by this in practice? It feels like the check is awkwardly distant / indirect / at a different level from actual user input via argument.

I'm tempted to just leave this edge case unhandled. If it seems really important, can we rephrase it as an immediate and simple check on user input?

malcolmbarrett: Believe it or not, I have several times due to a slip of the finger, and it's actually pretty annoying to then clean up. I'll refactor the checks to the front

jennybc:
Do you consider this done?

Do you want to rebase on `main` to get a green check (or I can)?

malcolmbarrett:
I'll update the error tests to snapshots but close to done

jennybc:
Something seems to have gone a bit sideways Git-wise, maybe with the rebase? I.e. all sorts of unrelated changes are showing up now.

jennybc (approved):
Huh the One Weird Trick is new to me!

In any case, it looks good now! I'm approving.

IMO you don't need so many snapshot tests. I think once we've captured it once, the others are mostly superfluous, since this isn't a particularly tricky bit of logic.

I'll leave it in your hands to merge.

malcolmbarrett:
ha, I actually caught a bug because of the tests (code coverage ftw!) where things were misplaced in one but not the others. Helped me fix it quickly. So I think it's earned its keep unless you feel particularly strongly.

PR #1640, opened 2022-05-24
Restore ability to get GHA config from local file

### LINKED ISSUE #1548: `use_github_action()` can't be used anymore with local template
Previously, when passed a local file path as `url` parameter , `use_github_action()` would work. I was using that as a way to share GHA template YAML in local package.

Following change in https://github.com/r-lib/usethis/pull/1509, `use_github_file()` is used and it now assumes that the file is only on Github. Previously `read_utf8(url)` was handling both.

I may have been using `use_github_action()` in an unexpected way, but it was useful because `use_github_action()` does more than just copying. Some steps I would need to reproduce if I were using  `use_template()`. 

Could support for using a local action YAML file in addition to a URL be brought back ? 
Or is this change intentional ? 

I just wanted to share as obviously this is a breaking change for an unwarranted usage I guess and I have workaround. But maybe it would be useful to others too.

jennybc:
Closes #1548

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -139,8 +139,14 @@ use_github_action <- function(name,
   save_as <- path(".github", "workflows", save_as)
   create_directory(path_dir(proj_path(save_as)))
 
-  # `ignore = FALSE` because we took care of this at directory level, above
-  new <- use_github_file(url, save_as = save_as, ignore = FALSE, open = open)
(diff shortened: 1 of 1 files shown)


jennybc:
@cderv does this meet your needs?

cderv:
Yes it is working as I expect. This simplify the usage for me. Thank you.

PR #1643, opened 2022-06-01
Fix issues when creating nested package

lionel-:
I would like to optimise the creation of packages in the `testthat` folder when I work on packages like pkgload. This PR fixes two bugs:

- Fix the escape hatch in `project_name()` so that it works in nested packages. Currently if I create a package in `pkgload/testthat/tests` the nested package is named `pkgload`. A subsequent `load_all()` inside that package then overwrites pkgload itself with the empty package, which causes puzzling errors.

- Remove the early exit in `proj_set()` to allow `local_project()` to set a nested project. This didn't break any tests and seems like this would prevent other bugs when working with nested packages. We could find a more localised workaround for `create_package()` though.

The PR also adds an undocumented global option `usethis.allow_nested_package` to disable the user challenge.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # usethis (development version)
 
-* `create_package()` now returns the correct package path when creating a package nested in another project.
+* `create_package()` now returns the correct package path and infers the correct package name when creating a package nested in another project.
 
 # usethis 2.1.6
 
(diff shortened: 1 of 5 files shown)


jennybc:
Just FYI, I will want to review this carefully, because it definitely touches fiddly bits of usethis. However, I also think the test coverage of the known-to-be-trick parts it's pretty good, so that should help.

PR #1646, opened 2022-06-09
plain licenses

pachadotdev:
Now `use_*_license()` writes LICENSE.md and LICENSE files, in order to provide the correct license information on GitHub repositories and in R packages. 

The plain file is not added to .Rbuildignore. By following https://github.com/tidyverse/dplyr, I used {{ }} to write the year and author in plain MIT license, while Apache and others are kept unaltered. Also in dplyr, LICENSE is not ignored in the build, the same pattern is used in here.

Here's an example, https://github.com/pachadotdev/gravity now shows license Apache, but when I added LICENSE.md the license field showed "Unknown".

Screenshot 2022-06-09 at 17-40-32 pachadotdev_gravity R package that provides estimation methods for Gravity Models

The license plain files were obtained from GitHub, except for https://www.gnu.org/licenses/lgpl-3.0.txt and https://creativecommons.org/licenses/by/4.0/legalcode.txt

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -316,6 +316,8 @@ New `use_proprietary_license()` allows your package to pass R CMD check while ma
   
 The `name` argument to `use_mit_license()` has been changed to `copyright_holder` to make the purpose more clear. The `name` argument has been removed from all other license functions because it is not needed; no other license makes an assertion about who the copyright holder is.
 
+Now `use_*_license()` writes LICENSE.md and LICENSE files, in order to provide the correct license information on GitHub repositories and in R packages. The plain file is not added to .Rbuildignore. (@pachadotdev, #1646)
+
 ## RStudio preferences
 
 usethis is now fully cognizant of the [changes to RStudio preferences](https://www.rstudio.com/blog/rstudio-1-3-preview-configuration/) in RStudio 1.3:
@@ -434,7 +436,7 @@ Patch release to align some path handling internals with an update coming in the
 
 * New `ui_silence()` makes it easier to selectively silence some UI output.
 
-* New `use_agpl3_license()` (@pachamaltese, #870).
+* New `use_agpl3_license()` (@pachadotdev, #870).
 
 * New `use_data_table()` to set up a package for Import-ing `data.table` 
   (@michaelchirico, #897).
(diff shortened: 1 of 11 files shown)


malcolmbarrett:
How usethis uses `LICENSE` and `LICENSE.md` is due to [CRAN policies](https://r-pkgs.org/license.html#key-files): 

> LICENSE.md includes a copy of the full text of the license. All open source licenses require a copy of the license to be included, but CRAN does not permit you to include a copy of standard licenses in your package, so we also use .Rbuildignore to make sure this file is not sent to CRAN.

I'm also aware of the weird stuff GitHub does in the presence of these files, but it seems like this might be a better feature request on their end due to CRAN restrictions on ours

pachadotdev:
> 

thanks a lot, would it make sense to include the plain LICENSE but ignore it in .Rbuildignore?

pachadotdev:
> How usethis uses `LICENSE` and `LICENSE.md` is due to [CRAN policies](https://r-pkgs.org/license.html#key-files):
> 
> > The LICENSE file can also contain the full text of non-standard and non-open source licenses. You are not permitted to include the full text of standard licenses
> 
> > LICENSE.md includes a copy of the full text of the license. All open source licenses require a copy of the license to be included, but CRAN does not permit you to include a copy of standard licenses in your package, so we also use .Rbuildignore to make sure this file is not sent to CRAN.
> 
> I'm also aware of the weird stuff GitHub does in the presence of these files, but it seems like this might be a better feature request on their end due to CRAN restrictions on ours

thanks a lot, would it make sense to include the plain LICENSE but ignore it in .Rbuildignore?

malcolmbarrett:
I don't think so. From [Writing R Extensions](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Licensing):

> It is very important that you include license information! Otherwise, it may not even be legally correct for others to distribute copies of the package, let alone use it. 

I also suspect CRAN wouldn't be too happy to see `file LICENSE` with no file included in the build

jennybc:
This aspect of usethis has been very intentionally crafted (most recently in #1208) to come as close as possible to open source and common sense norms, while staying within the CRAN rules. I am not aware of any changes at CRAN that would suggest we need to rework things.

PR #1647, opened 2022-06-10
Rationalize nested projects

jennybc:
Closes #1643

We discourage nested projects, but it is possible, mostly for developers. So we should at least behave correctly in this code path.

Previously, a nested project would put the name of the parent project in DESCRIPTION, which causes a variety of downstream problems. The `.Rproj` file gets the wrong name. Every call to `project_name()` returns the wrong name.

Also, when `proj_set(force = TRUE)`, we should bypass the initial early return if we discover that we are already in a project. The whole point of `force = TRUE` is to do exactly as we are told. This was causing the active project to be still be set to the parent, when trying to work with a nested project.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* Although nested projects are discouraged, they can be useful in development contexts. `create_package()` now sets the correct package name and returns the correct package path for a package nested inside a project (#1647). 
+
 # usethis 2.1.6
 
 ### GitHub-related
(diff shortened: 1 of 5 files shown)


jennybc:
@lionel- I think this should solve your problem and this is my preferred solution.

lionel- (commented):
LGTM. Does this still prompt users about nested projects? If yes, could you also take a look at the undocumented option added in the other PR please?

jennybc:
> Does this still prompt users about nested projects? If yes, could you also take a look at the undocumented option added in the other PR please?

OK now this PR will honour the undocumented option used in your PR (and I'm using that in the usethis tests instead of a mock).

lionel-:
Thanks!

PR #1650, opened 2022-06-24
Better defaults for use_latest_dependencies()

hadley:
I doubt anyone uses this apart from me, and I have to remember to change these arguments every time.

### DIFF
--- a/R/latest-dependencies.R
+++ b/R/latest-dependencies.R
@@ -5,10 +5,11 @@
 #'
 #' @keywords internal
 #' @export
-#' @param overwrite By default (`FALSE`), only dependencies without version
-#'   specifications will be modified. Set to `TRUE` to modify all dependencies.
+#' @param overwrite By default (`TRUE`), all dependencies will be modified.
+#'   Set to `FALSE` to only modify dependencies without version
+#'   specifications.
 #' @param source Use "local" or "CRAN" package versions.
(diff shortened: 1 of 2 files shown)


jennybc (approved):
I made some changes to get tests passing.

If you're happy, I'm happy. It's true that I do not use this function.

R/latest-dependencies.R
```diff
@@ -5,18 +5,20 @@
...
-#' @param overwrite By default (`FALSE`), only dependencies without version
-#'   specifications will be modified. Set to `TRUE` to modify all dependencies.
+#' @param overwrite By default (`TRUE`), all dependencies will be modified.
+#'   Set to `FALSE` to only modify dependencies without version
+#'   specifications.
 #' @param source Use "local" or "CRAN" package versions.
-use_latest_dependencies <- function(overwrite = FALSE, source = c("local", "CRAN")) {
+use_latest_dependencies <- function(overwrite = TRUE, source = c("CRAN"," local")) {
   deps <- desc::desc_get_deps(proj_get())
+  source <- arg_match(source)
```
jennybc, line 14: This isn't actually necessary, once I give the `update_versions()` helper the same defaults. But it still seems like a good idea, e.g. in terms of the error message.

hadley: I think it might be because of the complicated way that `arg_match()` relies on missing-ness.

PR #1661, opened 2022-08-11
Fix dependencies link in R/package.R

andreranza:


jennybc:
Thanks! This (where stuff is in R Packages) is very much in flux. But this is certainly a better URL at this moment. Hopefully we don't change it again :grimacing:

Issue #1554, opened 2021-12-18
More accessible badges


jennybc:
Since usethis is the package that adds a lot of badges to READMEs, maintainers have asked us to help make those badges more accessible to screen readers.

I suspect the best way to do this is via the badge SVGs themselves.

@gaborcsardi already added this to METACRAN's badges:

``` r
library(tidyverse)

show_aria_label <- function(badge) {
  badge %>%
    xml2::read_xml() %>%
    xml2::as_list() %>%
    pluck("svg") %>%
    attributes() %>%
    pluck("aria-label")
}
```

<img src="https://www.r-pkg.org/badges/version/usethis">

``` r
show_aria_label("https://www.r-pkg.org/badges/version/usethis")
#> [1] "CRAN 2.1.5"
```

<img src="https://www.r-pkg.org/badges/last-release/usethis">

``` r
show_aria_label("https://www.r-pkg.org/badges/last-release/usethis")
#> [1] "CRAN 2021-12-09"
```

<img src="https://cranlogs.r-pkg.org/badges/usethis">

``` r
show_aria_label("https://cranlogs.r-pkg.org/badges/usethis")
#> [1] "CRAN downloads 270K/month"
```

Any badge produced via <https://shields.io/> is already accessible:
<img src="https://img.shields.io/github/workflow/status/r-lib/usethis/R-CMD-check?label=R-CMD-check">

``` r
show_aria_label("https://img.shields.io/github/workflow/status/r-lib/usethis/R-CMD-check?label=R-CMD-check")
#> [1] "R-CMD-check: passing"
```

<img src="https://img.shields.io/badge/my__label-my__message-orange">

``` r
show_aria_label("https://img.shields.io/badge/my__label-my__message-orange")
#> [1] "my_label: my_message"
```

The features of <https://shields.io/> badges mean we can often substitute an accessible badge, even if the "native" badge is not (e.g. for GitHub or Codecov). I am about to explore that idea further in a pull request.

I've already placed requests with the lifecycle package (https://github.com/r-lib/lifecycle/issues/117) and with GitHub Actions ([community post](https://github.community/t/screen-reader-accessibility-of-github-actions-badges/218044), [post in Actions and Packages Feedback](https://github.com/github/feedback/discussions/8974)).

jennybc:
Notes on usethis's badge functions and the associated services.

- `use_binder_badge()` 
- `use_bioc_badge()` *encourage bioconductor to address?*
[...]

PR #1663, opened 2022-08-16
Add aria-labels and titles to lifecycle badges.

batpigandme:
Add `aria-label` and semantic title to lifecycle-badge SVGs for accessibility.

Part of https://github.com/r-lib/usethis/issues/1554 and related to https://github.com/r-lib/lifecycle/issues/117.


jennybc:
So IIUC this PR to usethis is indicative of the badge update that any package that's embedded lifecycle badges would need to do, in order to get more accessible badges. usethis is sort of a guinea pig. Full story in https://github.com/r-lib/lifecycle/issues/117.

It demonstrates that the `use_lifecycle_badge()` bullet of #1554 can be updated to reflect that this is fixed upstream (or soon will be).

PR #1664, opened 2022-08-19
Move url check up

thomasp85:
This is a slight tweak to the release issue. I find that bad urls are often present in the readme (e.g. from codecov) so making the check before building the readme seems safer in terms of ensuring the fix gets propagated to the .md version


malcolmbarrett:
I've bumped into that, too. Thanks!

PR #1676, opened 2022-09-14
remove usethis.protocol  = "ssh" from example .Rprofile options

shannonpileggi:
Per discussion on R packages 2e https://github.com/hadley/r-pkgs/issues/854


jennybc:
Thanks!

PR #1682, opened 2022-09-20
Add reformat argument to `use_rstudio()`

### LINKED ISSUE #1679: `use_rstudio()` inside of `create_from_github()` needs to be conservative
i.e. match the line endings and don't trim whitespace. Comes up when creating PRs for revdeps

hadley:
And set to `FALSE` in `create_from_github()`. Fixes #1679

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # usethis (development version)
 
-`use_rstudio()` gains a `reformat` argument which allows you to opt-out of the default reformatting. `create_from_github()` uses this option in order to preserve the defaults of whoever created the project (#1679).
+`use_rstudio()` gains a `reformat` argument which omits `.Rproj` settings that enforce file formatting conventions, e.g. around whitespace. `create_from_github()` uses this option when it introduces an `.Rproj` to a project that lacks one, making it easier to follow the project's existing conventions (#1679).
 
 The SVG badges placed by `use_lifecycle()` have improved accessibility features, i.e. they advertise the lifecycle stage via the `aria-label` attribute (#1554, https://github.com/r-lib/lifecycle/issues/117).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+`use_rstudio()` gains a `reformat` argument which allows you to opt-out of the default reformatting. `create_from_github()` uses this option in order to preserve the defaults of whoever created the project (#1679).
+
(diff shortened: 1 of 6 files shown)


R/rstudio.R
```diff
@@ -11,16 +11,27 @@
 #'   * Adds RStudio files to `.Rbuildignore`, if project is a package
 #'
 #' @param line_ending Line ending
+#' @param reformat If `TRUE`, the `.Rproj` is setup with common options that
+#'   reformat files on save: adding a trailing newline, trimming trailing
+#'   whitespace, and setting the line-ending. This is best practice for
+#'   new projects.
+#'
+#'   If `FALSE`, leaves all these settings alone, as is most appropriate
+#'   when you're contributing to someone else's project.
```
jennybc, line 20:
```suggestion
#'   If `FALSE`, these options are left unset, which is more appropriate when
#'   you're contributing to someone else's project that does not have its own
#'   `.Rproj` file.
```

At first the wording made me think different settings were being harmonized or merged. But this is really about preventing an external contributor's preference for using RStudio from "leaking" whitespace changes into a PR.

R/create.R
```diff
@@ -289,7 +289,7 @@ create_from_github <- function(repo_spec,
   rstudio <- rstudio %||% rstudio_available()
   rstudio <- rstudio && !is_rstudio_project(proj_get())
   if (rstudio) {
-    use_rstudio()
+    use_rstudio(reformat = FALSE)
```
jennybc, line 292:
```suggestion
    use_rstudio(reformat = !fork)
```

There is part of me that thinks we should only set reformat = FALSE in a fork, but I can also ignore that part of me.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+`use_rstudio()` gains a `reformat` argument which allows you to opt-out of the default reformatting. `create_from_github()` uses this option in order to preserve the defaults of whoever created the project (#1679).
```
jennybc, line 3:
```suggestion
`use_rstudio()` gains a `reformat` argument which omits `.Rproj` settings that enforce file formatting conventions, e.g. around whitespace. `create_from_github()` uses this option when it introduces an `.Rproj` to a project that lacks one, making it easier to follow the project's existing conventions (#1679).
```

jennybc (approved):
Coming in cold, it took me a moment to really absorb what the issue here is. I wondered if this was about reformatting the `.Rproj` file itself. Or if we were trying to merge local and remote `.Rproj` files. (Neither of which is true.)

So I made a few suggestions that helped me understand what's actually going on.

R/create.R
hadley: Regardless of whether you're forking or not, I feel like you should leave the code alone.

Issue #1677, opened 2022-09-14
Change constant `git_ignore_lines` element `.Rdata` to`.RData`


dchiu911:
I believe the standard R workspace is called `.RData`. However, the constant `git_ignore_lines` has the value ".Rdata".  So if the user already has `.RData` in their `.gitignore` and runs `usethis::use_git()` to initialize a repo,  they'd have `.Rdata` added when `.RData` is the same

jennybc:
Yeah from the history it looks like `use_git()` used to ignore `.Rhistory`, `.RData`, and `.Rproj.user` and `git_vaccinate()` ignored `.Rproj.user`, `.Rhistory`, `.Rdata`, `.httr-oauth`, and `.DS_Store` and then they were harmonized (https://github.com/r-lib/usethis/commit/26d2c4460ab9cb605a3d0d6682cfcf1f20c8285b) with `.Rdata` surviving, when it seems like `.RData` would make more sense.

Would you like to make a PR? This should have a NEWS bullet.

PR #1686, opened 2022-09-26
Fix git ignore and other small updates

dchiu911:
Fix git ignore constant patterns in #1677 and a couple other small updates

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,11 @@
 # usethis (development version)
 
+* Add trailing slash to DESCRIPTION URLs created from `use_github_links()` to adhere to new CRAN URL checks
+
+* Update TODO note in `use_github_release()` from "Edit draft" to "Edit release" (#1678)
+
+* Change `.gitignore` default pattern from `.Rdata` to `.RData` (#1677)
+
 * `use_rstudio()` gains a `reformat` argument which omits `.Rproj` settings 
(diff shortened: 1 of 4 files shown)


NEWS.md
```diff
@@ -1,5 +1,11 @@
 # usethis (development version)
 
+* Add trailing slash to DESCRIPTION URLs created from `use_github_links()` to adhere to new CRAN URL checks
```
jennybc, line 3: Let's keep the PR more focused on one thing. I'm pretty sure we fixed our slash problems already? E.g. #1526 and #1527. Even if not, it's easier to think about these things separately.

jennybc (commented):
> Fix git ignore constant patterns in #1677 and a couple other small updates

The above won't actually close the relevant issue because it's too far from the magic utterance. Also I think this closes a 2nd issue too? (I'm suggesting you split off the 3rd thing)

dchiu911:
I'm using the magic in the commit messages, I think those ones should close the issues? The PR comment is just a general reiteration of what's being done.

NEWS.md
dchiu911: Sounds good, I'll revert and open another PR. I think the slash problems were only addressed in `code-of-conduct.R`, `logo.R`, `pkgdown.R`

PR #1700, opened 2022-11-08
Refactor `use_vignette()`

jennybc:
Motivated by experiments made while updating R Packages

Basically I found it confusing that knitr was added to `DESCRIPTION` early on and then ... later ... rmarkdown was added.

I can't think of any reason why we had it this way and looking through git blame didn't shed any light either.

This seems more obvious and obviously correct to me.

---

BEFORE

``` r
> use_vignette("my-vignette")
✔ Adding 'knitr' to Suggests field in DESCRIPTION
✔ Setting VignetteBuilder field in DESCRIPTION to 'knitr'
✔ Adding 'inst/doc' to '.gitignore'
✔ Creating 'vignettes/'
✔ Adding '*.html', '*.R' to 'vignettes/.gitignore'
✔ Adding 'rmarkdown' to Suggests field in DESCRIPTION
✔ Writing 'vignettes/my-vignette.Rmd'
• Modify 'vignettes/my-vignette.Rmd'
```

AFTER

``` r
> use_vignette("my-vignette")
✔ Adding 'knitr' to Suggests field in DESCRIPTION
✔ Adding 'rmarkdown' to Suggests field in DESCRIPTION
✔ Setting VignetteBuilder field in DESCRIPTION to 'knitr'
✔ Adding 'inst/doc' to '.gitignore'
✔ Creating 'vignettes/'
✔ Adding '*.html', '*.R' to 'vignettes/.gitignore'
✔ Writing 'vignettes/my-vignette.Rmd'
• Modify 'vignettes/my-vignette.Rmd'
```

### DIFF
--- a/R/vignette.R
+++ b/R/vignette.R
@@ -56,12 +56,11 @@ use_vignette_template <- function(template, name, title, subdir = NULL) {
   use_git_ignore(c("*.html", "*.R"), directory = "vignettes")
   use_dependency("rmarkdown", "Suggests")
 
-  if (!is.null(subdir)) {
-    path <- path("vignettes", subdir, asciify(name), ext = "Rmd")
-  } else {
+  if (is.null(subdir)) {
     path <- path("vignettes", asciify(name), ext = "Rmd")
+  } else {
+    path <- path("vignettes", subdir, asciify(name), ext = "Rmd")
   }
-
 
   data <- list(
     Package = project_name(),
--- a/R/vignette.R
+++ b/R/vignette.R
@@ -26,6 +26,8 @@ use_vignette <- function(name, title = name) {
   check_vignette_name(name)
 
   use_dependency("knitr", "Suggests")
+  use_dependency("rmarkdown", "Suggests")
+
   use_description_field("VignetteBuilder", "knitr", overwrite = TRUE)
   use_git_ignore("inst/doc")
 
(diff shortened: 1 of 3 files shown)


R/vignette.R
```diff
@@ -26,6 +26,8 @@ use_vignette <- function(name, title = name) {
   check_vignette_name(name)
 
   use_dependency("knitr", "Suggests")
+  use_dependency("rmarkdown", "Suggests")
```
jennybc, line 29: I hoisted this up out of a helper.

R/vignette.R
```diff
@@ -39,6 +41,13 @@ use_vignette <- function(name, title = name) {
 use_article <- function(name, title = name) {
   check_is_package("use_article()")
 
+  deps <- desc::desc_get_deps(proj_get())
+  if (!"rmarkdown" %in% deps$package) {
+    ui_done("
+      Adding {ui_value('rmarkdown')} to {ui_field('Config/Needs/website')}")
+    use_description_list("Config/Needs/website", "rmarkdown")
```
jennybc, line 48: Since we're doing this here in `use_article()` specifically, we know enough to avoid adding markdown as an actual dependency in the case of an "articles only" package. I.e. then you only need it in `Config/Needs/website`.

R/vignette.R
```diff
@@ -54,15 +63,13 @@ use_vignette_template <- function(template, name, title, subdir = NULL) {
     use_directory(path("vignettes", subdir))
   }
   use_git_ignore(c("*.html", "*.R"), directory = "vignettes")
-  use_dependency("rmarkdown", "Suggests")
 
-  if (!is.null(subdir)) {
```
jennybc, line 59: Drive-by change. I just prefer `if (TRUE)` instead of `if (!TRUE)`.

R/vignette.R
```diff
@@ -54,15 +63,13 @@ use_vignette_template <- function(template, name, title, subdir = NULL) {
     use_directory(path("vignettes", subdir))
   }
   use_git_ignore(c("*.html", "*.R"), directory = "vignettes")
-  use_dependency("rmarkdown", "Suggests")
```
jennybc, line 57: I can't think of any reason why we did this *here* specifically :thinking:

R/vignette.R
hadley: This seems worthy of a news bullet.

hadley (approved):

PR #1710, opened 2022-12-08
Documentation readability edits.

jonthegeek:
We're reading all of the usethis docs "cover to cover" in a new "book club" at R4DS. These are some tweaks I noticed while reading the Configuration and Git & GitHub sections. I didn't re-document, since I *think* that happens automatically.

I removed the git_vaccinate() reference in git_sitrep help, because I *think* it's left over from them previously living in the same doc.


jennybc:
Thanks!

PR #1713, opened 2022-12-12
Don't mention GHA for README-rendering in README templates

jennybc:
Inspired by R Package 2e revision. We don't actually wire this up in our own repos, so I think we should get it out of the templates. Streamlining content for the majority is always good. The GHA/CI keeners are always able to use the GHA example for rendering an `.Rmd`, if they are so motivated.

### DIFF
--- a/inst/templates/project-README
+++ b/inst/templates/project-README
@@ -27,7 +27,7 @@ What is special about using `README.Rmd` instead of just `README.md`? You can in
 summary(cars)
 ```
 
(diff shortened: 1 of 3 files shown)


inst/templates/project-README
```diff
@@ -27,7 +27,7 @@ What is special about using `README.Rmd` instead of just `README.md`? You can in
 summary(cars)
 ```
 
-You'll still need to render `README.Rmd` regularly, to keep `README.md` up-to-date. `devtools::build_readme()` is handy for this. You could also use GitHub Actions to re-render `README.Rmd` every time you push. An example workflow can be found here: <https://github.com/r-lib/actions/tree/v1/examples>.
+You'll still need to render `README.Rmd` regularly, to keep `README.md` up-to-date.
```
jennybc, line 30: I even deleted the reference to `devtools::build_readme()` here, because it's not relevant to a non-package project and, in fact, I don't even think it will work. In a non-package project, folks can use whatever their normal workflow is to render `README.Rmd`.

PR #1719, opened 2023-01-05
refactor `git_cfg_get()`

### LINKED ISSUE #1718: refactor internal function `git_cfg_get()`
This emerged from #1714:

I have two emails set up in my git config: my work email is `global` email, but I (try to) use my personal email as `local` for open-source projects.

When I run `git_sitrep()`, with both `global` and `local` emails set:

```
Warning message:
In grepl(de_facto_email, addresses) :
  argument 'pattern' has length > 1 and only the first element will be used
```

I traced this to a call to `git_cfg_get("user.email", "de_facto")`, which returns both emails, e.g.:

```r
git_cfg_get("user.email", "global")
#> "ian.lyttle@<work>.com"

git_cfg_get("user.email", "local")
#> "ijlyttle@me.com"  

git_cfg_get("user.email", "de_facto")
#> "ian.lyttle@<work>.com" "ijlyttle@me.com"  
```

@jennybc confirms `git_cfg_get("foo", "de_facto")` should return a single value, where the `local` value trumps the `global` value.

ijlyttle:
Fixes #1718

There may be a cleaner way to get this done; I think this makes a small change, which I hope will also be clear in intent.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,8 +24,12 @@
   shortcut is expanded via 
   [`fs::path_expand()`](https://fs.r-lib.org/reference/path_expand.html) 
   (@dpprdan, #1560).
+
+* Refactors internal function to return, at most, one value for 
+  git-configuration parameters (#1718, @ijlyttle).
   
 * Updates snapshot tests (#1715).
+
 
 # usethis 2.1.6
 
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks! I removed the NEWS bullet because this shouldn't be perceivable by a regular user.

PR #1730, opened 2023-01-17
Refactor get_active_file()

### LINKED ISSUE #1566: Teach `use_test()`/`test_active_file()` about `src`
So that `src/{foo}.c` or (.cpp, .h, etc) would correspond to `tests/testthat/test-{foo}.R`.

### LINKED ISSUE #1690: use_test() creates wrong file if period in file name
I'm on a "data.frame.R" tab, it will create a "test-data.R", I expect a "test-data.frame.R" tab.

It's due to `slug()`, due in turn to `fs::path_ext_set()` but I think for the latter it's the intended behaviour.

```
usethis:::slug("data.frame.R", "R")
#> [1] "data.R"

fs::path_ext_set("data.frame", "R")
#> data.R
```

hadley:
* Make it much broader so it now works with any file in `R/`, `src/` or `tests/testthat`. Fixes #1566
* This new foundation made it possible to make `use_c()` and `use_rcpp()` (and friends) to work the same way as `use_r()`/use_test()`.
* Since I was in here I fixed #1690.
* I also refactored the code to make it easier to test.
* I switched the errors I touch to use `cli::cli_abort()` so the messaging can point to the correct user facing function.
* I generally refactored the error handling so it happens in one central location.

### DIFF
--- a/R/r.R
+++ b/R/r.R
@@ -49,7 +49,7 @@ use_r <- function(name = NULL, open = rlang::is_interactive()) {
   use_directory("R")
 
   path <- path("R", compute_name(name))
-  edit_file(path, open = open)
+  edit_file(proj_path(path), open = open)
 
   invisible(TRUE)
 }
@@ -65,7 +65,7 @@ use_test <- function(name = NULL, open = rlang::is_interactive()) {
   if (!file_exists(path)) {
     use_template("test-example-2.1.R", save_as = path)
   }
-  edit_file(path, open = open)
+  edit_file(proj_path(path), open = open)
 
   invisible(TRUE)
 }
--- a/R/r.R
+++ b/R/r.R
@@ -205,13 +205,13 @@ compute_active_name <- function(path, ext, error_call = caller_env()) {
   ## R's notion of user's home directory
   path <- proj_path_prep(path_expand_r(path))
 
-  rel_path <- path_dir(proj_rel_path(path))
-  if (!rel_path %in% c("R", "src", "tests/testthat")) {
+  dir <- path_dir(proj_rel_path(path))
+  if (!dir %in% c("R", "src", "tests/testthat")) {
     cli::cli_abort("Open file must be a code or test file.", call = error_call)
   }
 
   file <- path_file(path)
-  if (rel_path == "tests/testthat") {
+  if (dir == "tests/testthat") {
     file <- gsub("^test[-_]", "", file)
   }
   as.character(path_ext_set(file, ext))
--- a/R/r.R
+++ b/R/r.R
@@ -38,30 +38,18 @@
 #'
 #' The [rename_files()] function can also be helpful.
 #'
-#' @param name Either a name without extension, or `NULL` to create the
-#'   paired file based on currently open file in the script editor. If
-#'   the `R/` file is open, `use_test()` will create/open the corresponding
-#'   test file; if the test file is open, `use_r()` will create/open the
-#'   corresponding `R/` file.
+#' @param name Either a string giving a file name (without directory) or
+#'   `NULL` to take the name from the currently open file in RStudio.
 #' @inheritParams edit_file
 #' @seealso The [testing](https://r-pkgs.org/tests.html) and
 #'   [R code](https://r-pkgs.org/r.html) chapters of
 #'   [R Packages](https://r-pkgs.org).
 #' @export
 use_r <- function(name = NULL, open = rlang::is_interactive()) {
-  check_not_empty_file_name(name)
-  name <- name %||% get_active_r_file(path = "tests/testthat")
-  name <- gsub("^test-", "", name)
-  name <- slug(name, "R")
-  check_file_name(name)
-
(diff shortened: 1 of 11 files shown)


R/r.R
```diff
@@ -38,30 +38,18 @@
...
-#'   corresponding `R/` file.
+#' @param name Either a string giving a file name (without directory) or
+#'   `NULL` to take the name from the currently open file in RStudio.
 #' @inheritParams edit_file
 #' @seealso The [testing](https://r-pkgs.org/tests.html) and
 #'   [R code](https://r-pkgs.org/r.html) chapters of
 #'   [R Packages](https://r-pkgs.org).
 #' @export
 use_r <- function(name = NULL, open = rlang::is_interactive()) {
-  check_not_empty_file_name(name)
```
hadley, line 52: I inlined `check_not_empty_filename()` into `check_file_name()` and moved that into `compute_name()`.

R/r.R
```diff
@@ -38,30 +38,18 @@
...
-  name <- gsub("^test-", "", name)
-  name <- slug(name, "R")
-  check_file_name(name)
-
   use_directory("R")
-  edit_file(proj_path("R", name), open = open)
 
-  test_path <- proj_path("tests", "testthat", paste0("test-", name, ".R"))
-  if (!file_exists(test_path)) {
-    ui_todo("Call {ui_code('use_test()')} to create a matching test file")
```
hadley, line 63: I removed this because no other helper includes similar text, but I can restore it if you think it's important.

R/rcpp.R
```diff
@@ -6,28 +6,20 @@
 #'   * May create an initial placeholder `.c` or `.cpp` file
 #'   * Creates `Makevars` and `Makevars.win` files (`use_rcpp_armadillo()` only)
 #'
-#' @param name If supplied, creates and opens `src/name.{c,cpp}`.
-#'
-#' @details
-#'
-#' When using compiled code, please note that there must be at least one file
```
hadley, line 13: This weird wrinkle (implemented in `use_src_example_script())` is no longer needed because we always create a file.

R/r.R
```diff
@@ -179,62 +162,89 @@ rename_files <- function(old, new) {
...
-  invisible(TRUE)
-}
+  ## rstudioapi can return a path like '~/path/to/file' where '~' means
+  ## R's notion of user's home directory
+  path <- proj_path_prep(path_expand_r(path))
 
-valid_file_name <- function(x) {
-  grepl("^[a-zA-Z0-9._-]+$", x)
-}
+  rel_path <- path_dir(proj_rel_path(path))
```
jennybc, line 208: `rel_path` feels like a confusing name for the immediate parent directory within the project, but I guess it's not terribly important.

R/rcpp.R
```diff
@@ -6,28 +6,20 @@
...
 
   use_dependency("Rcpp", "LinkingTo")
   use_dependency("Rcpp", "Imports")
   roxygen_ns_append("@importFrom Rcpp sourceCpp") && roxygen_remind()
 
-  use_src_example_script(name, "cpp")
+  use_src()
+  path <- path("src", compute_name(name, "cpp"))
+  use_template("code.cpp", path)
+  edit_file(path)
```
jennybc, line 22: More bad `proj_path()` vibes.

R/rcpp.R
```diff
@@ -66,11 +58,12 @@ use_rcpp_eigen <- function(name = NULL) {
...
   check_is_package("use_c()")
   check_uses_roxygen("use_c()")
-  check_not_empty_file_name(name)
 
   use_src()
 
-  use_src_example_script(name, "c")
+  path <- path("src", compute_name(name, ext = "c"))
+  use_template("code.c", path)
+  edit_file(path)
```
jennybc, line 66: And again.

R/rcpp.R
jennybc: Actually, is there a reason we aren't just specifying `open = is_interactive()` in `use_template()`?

jennybc (approved):
I didn't do any manual testing, but I assume you have.

R/rcpp.R
hadley: Yeah, because that only opens the file if it changed.

R/r.R
```diff
@@ -38,30 +38,18 @@
...
-
   use_directory("R")
-  edit_file(proj_path("R", name), open = open)
 
-  test_path <- proj_path("tests", "testthat", paste0("test-", name, ".R"))
-  if (!file_exists(test_path)) {
-    ui_todo("Call {ui_code('use_test()')} to create a matching test file")
-  }
+  path <- path("R", compute_name(name))
+  edit_file(path, open = open)
```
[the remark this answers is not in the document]
hadley: Why does `use_template()` automatically adjust the path but `edit_file()` does not?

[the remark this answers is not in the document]
hadley: Oh because `edit_file()` can work outside of the project.

PR #1731, opened 2023-01-17
Also rename files in src/

### LINKED ISSUE #1585: rename_files() should also affect files in src/

hadley:
Fixes #1585

### DIFF
--- a/R/r.R
+++ b/R/r.R
@@ -119,7 +119,7 @@ rename_files <- function(old, new) {
 
   # src/ ------------------------------------------------------------------------
   if (dir_exists(proj_path("src"))) {
-    src_old <- dir_ls(proj_path("src"), glob = glue("*/{old}.*"))
+    src_old <- dir_ls(proj_path("src"), glob = glue("*/src/{old}.*"))
 
     src_new_file <- gsub(glue("^{old}"), glue("{new}"), path_file(src_old))
     src_new <- path(path_dir(src_old), src_new_file)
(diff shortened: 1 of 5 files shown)


tests/testthat/test-r.R
```diff
@@ -37,6 +37,24 @@ test_that("renames R and test and snapshot files", {
   expect_proj_file("tests/testthat/_snaps/bar.md")
 })
 
+test_that("renames src/ files", {
+  create_local_package()
+  git_init()
+
+  dir_create(proj_path("src"))
```
jennybc, line 44:
```suggestion
  use_src()
```

R/r.R
```diff
@@ -108,19 +109,32 @@ rename_files <- function(old, new) {
...
   r_old_path <- proj_path("R", old, ext = "R")
   r_new_path <- proj_path("R", new, ext = "R")
   if (file_exists(r_old_path)) {
     ui_done("Moving {ui_path(r_old_path)} to {ui_path(r_new_path)}")
     file_move(r_old_path, r_new_path)
   }
 
+  # src/ ------------------------------------------------------------------------
+  if (dir_exists(proj_path("src"))) {
+    src_old <- dir_ls(proj_path("src"), glob = glue("*/{old}.*"))
```
jennybc, line 122:
```suggestion
    src_old <- dir_ls(proj_path("src"), glob = glue("*/src/{old}.*"))
```
Feels a bit more consistent with the assumptions I think are in the `gsub()`, i.e. we're only dealing with files in top-level of `src/`.

jennybc (approved):

R/r.R
hadley: Ooooh I thought the glob was relative to the first argument.

PR #1735, opened 2023-01-18
Defunct `use_travis()` and `use_appveyor()`

### LINKED ISSUE #1517: Deprecate travis functions
Fully deprecate everything re: Travis and perhaps beyond, i.e. other services we do not actively use.

hadley:
And mark `use_circleci()` and `use_gitlab_ci()` as questioning.

Fixes #1517


R/ci.R
```diff
@@ -1,141 +1,23 @@
 #' Continuous integration setup and badges
 #'
 #' @description
+#' `r lifecycle::badge("questioning")`
```
jennybc, line 4: With my recent lifecycle / r-pkgs work, I am reminded of this:

https://lifecycle.r-lib.org/articles/stages.html#superseded-stages

> Superseded stages

> We no longer use these stages, but we document them here because we have used them in the past.

> Questioning ...

Has "questioning" been rehabilitated?

(I'm not deeply invested in this but thought I'd mention.)

jennybc (approved):

R/ci.R
hadley: We have used it a bit in dplyr, so maybe it's more useful than we thought, even if it's not super actionable.

PR #1737, opened 2023-01-18
Interactive `use_github_action()`

### LINKED ISSUE #1724: Deprecate `use_github_actions()`?
Since it's just a thin wrapper around `use_github_action_check_release()` with a less informative name. (And that's not a workflow we'd generally recommend anyway, since it only checks on linux)

hadley:
Calling `use_github_action()` without arguments now generates an interactive prompt. This deprecates `use_github_actions()`, `use_github_action_check_release()`, etc since you can now achieve the same result with less typing while getting more information about what the workflow will do. 

Fixes #1724


R/github-actions.R
```diff
@@ -133,7 +95,11 @@ use_github_action <- function(name,
   use_dot_github(ignore = ignore)
 
   if (is.null(save_as)) {
-    save_as <- path_file(url)
+    if (is_check_action(url)) {
```
hadley, line 98: This feels a bit icky — `use_github_action()` was previously an internal helper that happened to have a really good name. And now I've squished in the naming and badge logic from the older helpers.

But I think the net effect of this is neutral since you probably don't want multiple `R CMD check` workflows in one repo. (And if you do, you can use `save_as`)

R/github-actions.R
```diff
@@ -152,15 +118,97 @@ use_github_action <- function(name,
...
+    "Which action do you want to add? (0 to exit)\n",
+    "(See {.url https://github.com/r-lib/actions/tree/v2/examples} for other options)"
+  )
+  # Any changes here also need to be reflected in documentation
+  workflows <- c(
+    "check-standard" = "Run R CMD check in four environments",
+    "test-coverage" = "Compute test coverage and report to https://about.codecov.io",
+    "lint" = "Run lintr::lint_package() and report problems",
+    "pr-commands" = "Add /document and /style commands for pull requests"
+  )
```
hadley, line 147: It would be good to put some thought into what we should list here. Maybe we should have separate lists depending on whether or not `is_package()` is `TRUE `?

R/github-actions.R
```diff
@@ -152,15 +118,97 @@ use_github_action <- function(name,
...
+
+  prompt <- cli::format_inline(
+    "Which action do you want to add? (0 to exit)\n",
+    "(See {.url https://github.com/r-lib/actions/tree/v2/examples} for other options)"
+  )
+  # Any changes here also need to be reflected in documentation
+  workflows <- c(
+    "check-standard" = "Run R CMD check in four environments",
+    "test-coverage" = "Compute test coverage and report to https://about.codecov.io",
+    "lint" = "Run lintr::lint_package() and report problems",
```
hadley, line 145: Need to add to docs, if we decide to keep.

R/github-actions.R
```diff
@@ -1,114 +1,76 @@
-# general GHA setup ------------------------------------------------------------
-
-#' GitHub Actions setup
+#' Set up a GitHub Action
```
jennybc, line 1:
```suggestion
#' Set up a GitHub Actions workflow
```

More inline with GitHub's verbiage and works nicely with the `Workflows` heading below.

R/github-actions.R
```diff
@@ -1,114 +1,76 @@
...
+#'
+#' * `check-standard`: This recommended workflow runs `R CMD check` with
+#'    R-latest on Linux, Mac, and Windows, and with R-devel on Linux. This
+#'    is a good baseline if you plan on submitting your package to CRAN.
+#' * `test-coverage`: Compute test coverage and report to
+#'    <https://about.codecov.io> by calling [covr::codecov()].
+#' * `pr-commands`: This workflow enables the use of two R-specific commands
+#'    in pull request issue comments: `/document` to run
+#'    `roxygen2::roxygenise()` and `/style` to run `styler::style_pkg()`. Both
+#'    will update the PR with any changes once they done.
```
jennybc, line 22:
```suggestion
#'    will update the PR with any changes once they're done.
```

R/github-actions.R
```diff
@@ -1,114 +1,76 @@
...
+#'
+#' If you call `use_github_action()` with out arguments, you'll be prompted to
```
jennybc, line 24:
```suggestion
#' If you call `use_github_action()` without arguments, you'll be prompted to
```

R/github-actions.R
jennybc: pkgdown comes to mind

R/github-actions.R
```diff
@@ -152,15 +118,97 @@ use_github_action <- function(name,
...
+# deprecated --------------------------------------------------------------
+
+#' Use specific Github actions
+#'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
+#' `use_github_actions()` is deprecated because it was just an alias
+#' for [use_github_action_check_release()]. The other variants are deprecated
+#' because [use_github_action()] can  now suggest specific workflows to use.
```
jennybc, line 197:
```suggestion
#' Use specific GitHub Actions workflows
#'
#' @description
#' `r lifecycle::badge("deprecated")`
#'
#' `use_github_actions()` is deprecated because it was just an alias
#' for [use_github_action_check_release()]. The other variants are deprecated
#' because [use_github_action()] can now suggest specific workflows to use.
```

tests/testthat/test-github-actions.R
```diff
@@ -18,7 +19,11 @@ test_that("use_github_action() allows for custom urls", {
   )
   expect_proj_dir(".github")
   expect_proj_dir(".github/workflows")
-  expect_proj_file(".github/workflows/check-full.yaml")
+  expect_proj_file(".github/workflows/R-CMD-check.yaml")
+})
+
+test_that("use_github_action() still errors in non-iteractive environment", {
```
jennybc, line 25:
```suggestion
test_that("use_github_action() still errors in non-interactive environment", {
```

jennybc (approved):
I feel like `use_github_actions()` is the actual best name for the function that triggers the interactive menu, i.e. it's the most natural, given how people talk about GHA. "Oh, you want to do X? You should use GitHub Actions for that, so call `usethis::use_github_actions()` and follow your nose."

R/github-actions.R
hadley: I left it off because I thought we'd want to steer people towards `use_pkgdown_github_pages()`? Either way it's probably worth mentioning that in the docs.

hadley:
I agree that `use_github_actions()` is probably the most natural, but because `use_github_action()` already exists it will be the function suggested be autocomplete. And I can't see an obvious way to do the transition. So unless you have a specific plan, I think this is a situation where we just accept that the optimal function name is no longer possible.

PR #1740, opened 2023-01-19
Look for an .Rproj file anywhere in GitHub repo

### LINKED ISSUE #1680: Should `create_from_github()` support subdirs?
e.g. `create_from_github("https://github.com/apache/arrow/tree/master/r")`

hadley:
Also includes some minor refactoring of `is_rstudio_project()` and `rproj_path()`

Fixes #1680

@jennybc this feels like a pretty minimal fix for motivating problem. I tested it interactively  https://github.com/apache/arrow and it worked the way I wanted (i.e. it uses the `arrow.Rproj` found in the `R/` subdir.)


jennybc:
I fetched this and am playing with it a little. I have to admit it makes me nervous. As in I'm not convinced the upsides outweigh the downsides, just because so many path and project things end up being rather fiddly and are not covered well by tests. But if you really want to do this, I won't oppose it.

One immediate observation is that the RStudio instance that is launched this way is labelled as "r", as opposed to, e.g., "arrow". But that's not out problem, it's an IDE issue.

hadley:
I'm confident that this is a minimal change that won't affect existing behaviour when there is already a `.Rproj` in the root directory, but will make life easier in other access. In any case, the scope is pretty limited since it only affects `create_project_from_github()`.

hadley:
I'm confident that this is a minimal change that won't affect existing behaviour when there is already a `.Rproj` in the root directory, but will make life easier in other access. In any case, the scope is pretty limited since it only affects `create_project_from_github()`.

tests/testthat/test-rstudio.R
```diff
@@ -24,27 +24,30 @@ test_that("use_rstudio() omits package-related config for a project", {
...
+
+  file_touch(path(path, "a.Rproj"))
+  file_touch(path(path, "b.Rproj"))
+  expect_snapshot(rproj_path(path), error = TRUE)
 })
 
+test_that("a non-RStudio project is not recognized", {
+  create_local_package(rstudio = FALSE)
+  expect_false(is_rstudio_project())
+  expect_error(rproj_path(), NA_character_)
```
jennybc, line 47: This doesn't seem right. We *should* get an error, yeah? I actually don't understand why tests and CI are passing.

tests/testthat/test-rstudio.R
```diff
@@ -24,27 +24,30 @@ test_that("use_rstudio() omits package-related config for a project", {
...
 })
 
-test_that("we error for multiple Rproj files", {
-  create_local_package(rstudio = TRUE)
-  file_copy(
-    rproj_path(),
-    proj_path("copy.Rproj")
-  )
-  expect_usethis_error(rproj_path(), "Multiple .Rproj files found", fixed = TRUE)
+test_that("we error if there isn't exactly one Rproj files", {
```
jennybc, line 33:
```suggestion
test_that("we error if there isn't exactly one Rproj file", {
```

jennybc (commented):
My heart's not really in this but I won't oppose it.

R/course.R
```diff
@@ -383,7 +383,7 @@ tidy_unzip <- function(zipfile, cleanup = FALSE) {
   }
 
   if (is_interactive()) {
-    rproj_path <- dir_ls(target, regexp = "[.]Rproj$")
+    rproj_path <- rproj_paths(target)
```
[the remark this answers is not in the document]
hadley: This doesn't change any behaviour; it just centralises the `.Rproj` finding code in one place.

tests/testthat/test-rstudio.R
jennybc: If I run the whole test, it passes. But not if I walk through it line by line.

```
> devtools::load_all(".")
ℹ Loading usethis
> 
> test_that("a non-RStudio project is not recognized", {
+   create_local_package(rstudio = FALSE)
+   expect_false(is_rstudio_project())
+   expect_error(rproj_path(), NA_character_)
+ })
v Setting active project to '/private/tmp/Rtmpo0bGHz/testpkgee6f50e07b7b'
v Restoring original working directory: '/Users/jenny/rrr/usethis/'
v Setting active project to '<no active project>'
v Deleting temporary project: '/tmp/Rtmpo0bGHz/testpkgee6f50e07b7b/'
Test passed 😸
>   create_local_package(rstudio = FALSE)
Setting global deferred event(s).
i These will be run:
  * Automatically, when the R session ends.
  * On demand, if you call `withr::deferred_run()`.
i Use `withr::deferred_clear()` to clear them without executing.
✔ Setting active project to '/private/tmp/Rtmpo0bGHz/testpkgee6f348ba07'
>   expect_false(is_rstudio_project())
>   expect_error(rproj_path(), NA_character_)
Error in if (.p(.cnd, ...)) { : missing value where TRUE/FALSE needed
```

jennybc: Yeah, something does not make sense here. If I change this to `expect_snapshot()`, I definitely get an error.

```
test_that("a non-RStudio project is not recognized", {
  create_local_package(rstudio = FALSE)
  expect_false(is_rstudio_project())
  expect_snapshot(rproj_path())
  #expect_error(rproj_path(), NA_character_)
})
```

which gives

```
> devtools:::test_active_file()
[ FAIL 1 | WARN 0 | SKIP 0 | PASS 19 ]

── Error (test-rstudio.R:47): a non-RStudio project is not recognized ──────────
`rproj_path()` threw an unexpected error.
Message: "testpkgee6f39d74a7" is not an RStudio Project.
Class:   rlang_error/error/condition
Backtrace:

[ FAIL 1 | WARN 0 | SKIP 0 | PASS 19 ]
```

PR #1744, opened 2023-01-19
Add reminder to close milestone in release  issue

### LINKED ISSUE #1642: `use_release_issue()` could remind you to close the milestone
If it exists; I think we can assume the milestone name is the same as the tag name, e.g. `v2.0.4`

hadley:
Fixes #1642

@jennybc The test currently passes when run with `devtools::test()` but fails in `R CMD check` because the checked package is no longer a git repo. I'm happy to figure it out myself but I wondered if you might remember a function with similar needs that I could copy?


jennybc:
👀

tests/testthat/test-release.R
```diff
@@ -94,6 +94,11 @@ test_that("returns empty string if no bullets", {
   expect_equal(news_latest(lines), "")
 })
 
+test_that("can find milestone numbers", {
```
jennybc, line 97: So you seem to be relying on usethis itself being the active project here (?).

We have no other tests that do anything like this.

Given that `gh_milestone_number()` actually hits the GitHub API, we either have to not test this (my recommendation) or refactor it so that one can pass the repo spec in (but then it would still need some skips).

hadley: This was non-trivial to write, so I wanted to include at least one test. If it causes problems in the future, I promise to fix it.

Issue #1379, opened 2021-02-28
Should use_tidy_description() set a config flag?


hadley:
e.g. `Config/usethis/auto_tidy: TRUE` so that afterwards, any operation that touches the description would auto-tidy it?

jennybc:
I've always wanted this #443

hadley:
Moving this to desc

PR #1747, opened 2023-01-20
Improve desc usage

### LINKED ISSUE #1522: Rationalize the use of desc
It would be good to look at all the desc usage holistically and figure out if there are some missing helpers (e.g. I see a lot of `desc::desc(file = project_get())`.
Are we calling desc in places where we should just use the existing `package_data()` helper?
See if we can settle on one of the APIs, i.e. procedural vs. in place.
Look at when/how we trigger writing back to file and the general relationship with `write_over()`.
Are we doing some things in a kludgy way that desc already implements properly (or perhaps should implement)?

hadley:
* New `proj_desc()` helper to get `desc` object
* New `proj_version()` and `proj_deps()` for common needs.
* Eliminate `package_data()

Fixes #1522

Still need to think through `use_description_list()` and `use_description_impl_()` which use `write_over()`


hadley:
Maybe add `description` subclass that provides a `write_over()` method?

hadley:
@jennybc The use of `write_over()` in `use_description_impl_()` makes sense to me, because we're overwriting the complete `DESCRIPTION` file. I'm not so sure about its use in `use_description_list()` because we always use it to append to a list (i.e. `append = TRUE`). This is a non-destructive operation (similar to adding a new field to the `DESCRIPTION`) so I don't think it should require explicit user permission. That would make the behaviour consistent with `use_pkgdown_url()`, which appends the pkgdown url to `URL` without confirmation, and with `write_union()`. 

I'd suggest we rename `use_description_list()` to something like `proj_desc_field_append()` and then use that function in `use_pkgdown_url()`, `use_pkgdown_github_pages()`, `use_article()`, and `use_system_requirement()`. We can also centralise the currently duplicated user messaging code.

Does that make sense?

hadley:
(And maybe `use_description_impl_()` could become `proj_desc_create()`, `use_description_field` -> `proj_field_set()`, `tidy_desc()` -> `proj_desc_tidy()`)

R/data-table.R
```diff
@@ -15,11 +15,15 @@ use_data_table <- function() {
...
+  desc <- proj_desc()
+  deps <- desc$get_deps()
+
   if (any(deps$type == "Depends" & deps$package == "data.table")) {
     ui_warn("data.table should be in Imports or Suggests, not Depends")
     ui_done("Deleting data.table from {ui_field('Depends')}")
-    desc::desc_del_dep("data.table", "Depends", file = proj_get())
+
+    desc$del_dep("data.table", "Depends")
+    desc$write()
```
hadley, line 26: Once this is done I'll tackle #1379 by making `$write()` automatically tidy the description if some flag is set.

R/cpp11.R
```diff
@@ -28,8 +28,8 @@ use_cpp11 <- function() {
 }
 
 get_cpp_register_deps <- function() {
-  res <- desc::desc(package = "cpp11")$get_field("Config/Needs/cpp11/cpp_register")
-  strsplit(res, "[[:space:]]*,[[:space:]]*")[[1]]
+  desc <- desc::desc(package = "cpp11")
+  desc$get_list("Config/Needs/cpp11/cpp_register")
```
jennybc, line 32: Just confirming that the switch from `$get_field()` to `$get_list()` is intentional.

R/description.R
```diff
@@ -193,16 +194,16 @@ use_description_field <- function(name, value, overwrite = FALSE) {
...
 }
 
 use_description_list <- function(key,
                                  values,
-                                 append = TRUE,
-                                 desc = NULL) {
-  desc_provided <- !is.null(desc)
-  desc <- desc %||% desc::desc(file = proj_get())
+                                 append = TRUE) {
+  desc <- desc %||% proj_desc()
```
jennybc, line 206: `desc` still needs to be defined somewhere.

R/helpers.R
```diff
@@ -81,20 +85,18 @@ r_version <- function() {
 
 use_system_requirement <- function(requirement) {
   stopifnot(is_string(requirement))
-  existing_requirements <- desc::desc_get_field("SystemRequirements", default = character(), file = proj_get())
-  existing_requirements <- utils::head(strsplit(existing_requirements, ", ?"), n = 1)
 
-  if (requirement %in% existing_requirements) {
+  desc <- proj_desc()
+  existing <- desc$get_list("SystemRequirements", default = "")
```
jennybc, line 90: I'm getting the sense that maybe `$get_list()` didn't exist when we first started using desc but now it does and it's handy. Let's make sure we're stating a good minimum version for desc.

R/version.R
```diff
@@ -83,16 +83,15 @@ use_version <- function(which = NULL) {
 #' @export
 use_dev_version <- function() {
   check_is_package("use_dev_version()")
-  ver <- desc::desc_get_version(proj_get())
+  ver <- proj_desc()$get_version()
```
jennybc, line 86: Can this be `proj_version()` now?

jennybc (commented):
This is looking great! Yes, your proposals for additional refactoring sound good to me.

R/cpp11.R
hadley: Yup, although I missed the `[[1]]` (fixed momentarily)

R/description.R
hadley: Oops

R/helpers.R
hadley: Yeah, exactly. I've bumped the reqiured version to the latest.

R/proj-desc.R
```diff
@@ -0,0 +1,12 @@
+
+proj_desc <- function(path = proj_get()) {
+  desc::desc(file = path)
+}
+
+proj_version <- function() {
+  proj_desc()$get_field("Version")
```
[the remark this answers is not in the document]
hadley: That returns a `package_version` object which we only want in one place (`use_dev_version()`)

tests/testthat/test-helpers.R
```diff
@@ -65,11 +65,11 @@ test_that("use_dependency() upgrades a dependency", {
   withr::local_options(list(usethis.quiet = FALSE, crayon.enabled = FALSE))
 
   expect_message(use_dependency("usethis", "Suggests"))
-  expect_match(desc::desc_get("Suggests", proj_get()), "usethis")
+  expect_match(desc::desc_get("Suggests"), "usethis")
```
hadley, line 68: Using desc without specific arguments works in tests because `local_package()`/`local_project()` set the working directory. I don't love this interface but converting everything to `proj_desc()$get()` etc didn't seem like a big win to me.

jennybc (approved):

PR #1751, opened 2023-01-24
Refactor use_tidy_github_actions()

hadley:


R/github-actions.R
```diff
@@ -240,7 +233,7 @@ use_tidy_github_actions <- function(ref = NULL) {
     }
   }
 
-  invisible(full_status && pr_status && pkgdown_status && test_coverage_status)
```
hadley, line 243: We spent a lot of code collecting information that's never used anywhere. I think code is clearer without it.

R/github-actions.R
```diff
@@ -212,21 +212,14 @@ use_github_actions_badge <- function(name = "R-CMD-check.yaml",
...
-  # now we inline it here
-  full_status <- use_github_action(
-    "check-full.yaml",
-    ref = ref,
-    save_as = "R-CMD-check.yaml"
-  )
+  use_github_action("check-full.yaml", ref = ref)
   use_github_actions_badge("R-CMD-check.yaml", repo_spec = repo_spec)
 
-  pr_status <- use_github_action_pr_commands(ref = ref)
```
hadley, line 227: Came here specifically to fix this line because it's now deprecated.

jennybc (approved):

PR #1753, opened 2023-01-24
Set a flag to always tidy the description

### LINKED ISSUE #1379: Should use_tidy_description() set a config flag?

hadley:
Fixes #1379

Having implemented this, I'm not sure that usethis is the right place. @gaborcsardi what do you think moving the `$write()` logic into desc itself? That would guarantee every desc operation tidies, not just those that go through `proj_desc()`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,28 +1,19 @@
 Package: usethis
 Title: Automate Package and Project Setup
 Version: 2.1.6.9000
-Authors@R: 
-    c(person(given = "Hadley",
-             family = "Wickham",
-             role = "aut",
-             email = "hadley@rstudio.com",
-             comment = c(ORCID = "0000-0003-4757-117X")),
-      person(given = "Jennifer",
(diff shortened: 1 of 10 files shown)


gaborcsardi:
Yes, that would make sense.

jennybc:
I like the idea to put this in desc instead.

hadley:
@gaborcsardi maybe `Config/desc/tidy: true` ?

gaborcsardi:
That is good, yes.

jennybc:
So, as far as usethis is concerned, this is something we should add to the default `DESCRIPTION` we create, yeah? Or is it something we recommend people configure as a personal preference?

hadley:
We'd add it by default, I think.

hadley:
I'll file a follow up issue once I've added the functionality in desc

PR #1754, opened 2023-01-25
Add reminder to run use_github_links()

### LINKED ISSUE #1704: Remind to add GitHub & pkgdown URLs for first release in `use_release_issue()`
It is getting more and more important to make sure you add your GitHub & pkgdown URLs when submitting your package to CRAN. Yet, if you omit them, all checks will still pass, making it easy to forget it when submitting to CRAN.

I see a large number of packages on CRAN that don't include these links even though they are on GitHub and have a pkgdown website.

For reference:

- Having GitHub & pkgdown URLs in DESCRIPTION is great for discoverability and easy navigation from CRAN

- pkgdown URL is **required** to enable cross-linking from other pkgdown websites

- GitHub URL is **required** by pkgdown to link to source
- GitHub URL is **required** by r-universe to match GitHub & CRAN packages

There could be a check to optionally include the pkgdown URL item if a `_pkgdown.yml` file is found.

Bisaloo:
Fix #1704.

Is this what you had in mind?

If it looks good, I can add tests and a NEWS item.


R/release.R
```diff
@@ -72,6 +72,9 @@ release_checklist <- function(version, on_cran) {
   has_pkgdown <- uses_pkgdown()
   has_readme <- file_exists(proj_path("README.Rmd"))
 
+  repo <- github_url_from_git_remotes()
+  has_github_links <- repo %in% desc::desc_get_urls() && glue::glue("{repo}/issues") %in% desc::desc_get("BugReports")
```
hadley, line 76: I'd extract this out into a `has_github_links()` function that lives in `github.R`, but otherwise the approach looks good.

R/release.R
```diff
@@ -90,6 +93,7 @@ release_checklist <- function(version, on_cran) {
       todo("Check that `Authors@R:` includes a copyright holder (role 'cph')"),
       todo("Check [licensing of included files](https://r-pkgs.org/license.html#code-you-bundle)"),
       todo("Review <https://github.com/DavisVaughan/extrachecks>"),
+      if (!has_github_links) todo("`usethis::use_github_links()`"),
```
hadley, line 96: I think this should be checked on every release, not just the first.

jennybc:
Thanks!

PR #1758, opened 2023-01-30
Implement use_rstudio_preferences()

### LINKED ISSUE #1518: Export `use_rstudio_config()`
Supercedes #1515 per the discussion in #1516

hadley:
Fixes #1518


jennybc (approved):
LGTM, but I have a couple observations / confirmations to make.

From #1518:

> When this gets done, it's a good time to modernize all code where we work with RStudio prefs....
>
> why do we have our own functions for reading/writing RStudio preferences? Why aren't we using rstudioapi::writeRStudioPreference()?
>
> I'm not sure about the original justification, but as I attempt to rewrite these functions using just the API there are some pain points, like it's much harder to test what you're doing, and the RStudio API appears to be missing some functionality, like the ability to remove a setting.

So we decided to go ahead and export a function, but **not** to starting using rstudioapi for working with global RStudio preferences? Because there are more downsides than upsides at this time?

This does feel like a function where, once we export it, we tend to offer a `scope` argument, for global vs local. But there's nothing about this PR that limits our ability to generalize this later.

hadley:
Yeah, I don't think there any major advantages to using the rstudioapi.  

And global vs local is going to be hard todo, since the option names are different 😬

PR #1765, opened 2023-02-14
Minor refactoring of use_github()

### LINKED ISSUE #1757: Cleaner code for detecting default branch in `use_github()`
Can now check `create$default_branch != default_branch`, thanks to improvements in the github API.

hadley:
Thanks to improvements in GitHub API, can now more clearly test if the remote default branch needs to be synched to the local default branch.

Fixes #1757

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -178,8 +178,7 @@ use_github <- function(organisation = NULL,
     verbose = FALSE
   )
 
-  gbl <- gert::git_branch_list(local = TRUE, repo = repo)
-  if (nrow(gbl) > 1) {
+  if (create$default_branch != default_branch) {
     ui_done("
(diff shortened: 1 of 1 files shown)


jennybc (approved):
I think there is some chance I do this, because historically you can't always trust the default branch reported in the payload returned when creating a new repo (e.g. c6757de6f8ebf75c0acd88b8ad6415d4e93d5522, where we recently had to restore a workaround for the fork case). Of course, it's too bad I didn't leave such a comment here.

So, if there's more than 1 local branch, I think I set the default branch on GitHub, just to be safe and it doesn't have much cost.

But if you want to change this, you can.

jennybc:
I don't think it matters much what we do here, but I prefer the *status quo*.

PR #1773, opened 2023-02-23
Remove rhub from release checklist

jennybc:
This came from a discussion @hadley and I had today, while releasing httr and working on the CRAN release chapter of R Packages 2e.

We feel like we get so much value from our ongoing checks via GitHub Actions, that the last-minute pre-release r-hub checks don't really add anything, in most situations. Also, if there is some undiscovered problem, it now shows up in an automated incoming CRAN check and doesn't bother any humans. Then r-hub becomes relevant as part of the troubleshooting toolkit.

Something I considered but did not implement:

We could check if the package has `.github/workflows/R-CMD-check.yaml` and, if it **does not** have such a file, include the checklist items I'm deleting here.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -67,7 +67,6 @@ use_release_issue <- function(version = NULL) {
 release_checklist <- function(version, on_cran) {
   type <- release_type(version)
   cran_results <- cran_results_url()
-  has_src <- dir_exists(proj_path("src"))
   has_news <- file_exists(proj_path("NEWS.md"))
   has_pkgdown <- uses_pkgdown()
   has_lifecycle <- proj_desc()$has_dep("lifecycle")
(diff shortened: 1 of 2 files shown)


gaborcsardi (approved):

R/release.R
```diff
@@ -111,9 +110,6 @@ release_checklist <- function(version, on_cran) {
     todo("`devtools::build_readme()`", has_readme),
     todo("`devtools::check(remote = TRUE, manual = TRUE)`"),
     todo("`devtools::check_win_devel()`"),
-    todo("`rhub::check_for_cran()`"),
-    todo("`rhub::check(platform = 'ubuntu-rchk')`", has_src),
```
DavisVaughan, line 115: rchk (and maybe sanitizers) is one i do find helpful, because IIRC these checks don't get run until _after_ CRAN accepts your package. So it can be accepted and then you can get hit with a 2 week email due to an rchk issue, and this check would have helped you avoid that.

That is how I'm remembering this, but if I'm wrong about that and rchk is run as a pre-check by CRAN, then i wouldn't mind if this was removed

jennybc: I think my personal experience lately has been that I didn't actually get the info I needed from these. Either I got a PREPERROR and no information, r-hub's rchk results don't match CRAN's, or all the rchk results were actually all about cpp11 and not, e.g., readxl's code.

The initial item "check CRAN results" helps a bit with this, because you would presumably see if you had pre-existing rchk issues.

So, yeah, we are technically losing something. But I'm not sure it matters in practice. I could add these back with the condition being `has_src && !on_cran` to partially mitigate.

DavisVaughan: We can just remove them, nbd

Issue #1691, opened 2022-10-17
use_news_md is out of order in use_release_issue()


EmilHvitfeldt:
I was working on my CRAN release of https://github.com/tidymodels/modelenv, and when I got to `use_github_release()` I got the following error because the release candidate didn't have a news file

![Screen Shot 2022-10-17 at 9 19 47 AM](https://user-images.githubusercontent.com/14034784/196230326-cfcffafe-1ee5-4a63-bf3d-dfaa2f96eaaf.png)

Maybe we can move `use_news_md()` up to `Prepare for release:` if a news file haven't been detected when calling `use_release_issue()`?

jennybc:
Is this a duplicate of #1687?

EmilHvitfeldt:
It appears to, sorry for the duplicate!

jennybc:
Did you not have a NEWS file or did you have one with unusual / unexpected format? I'm just wondering why we're getting repeated reports of trouble with NEWS and `use_release_issue()`.

EmilHvitfeldt:
I didn't have a news file, to begin with. Just started with `create_package()`

jennybc:
I'm not able to replicate this. I can't explain why you and @cderv saw this.

Issue #1687, opened 2022-09-29
`use_github_release()` requires a NEWS file - Release issue should reflect that correctly


cderv:
Order of the release todo is the following for 

https://github.com/r-lib/usethis/blob/8ecb7ab8851faa7d8cc56bfb965e1895c326163e/R/release.R#L115-L117

In my first release package, I do not have a NEWS file yet - or at least I did not put one but checklist did not check for one.
https://github.com/r-lib/usethis/blob/8ecb7ab8851faa7d8cc56bfb965e1895c326163e/R/release.R#L69-L80

However, running `usethis::use_github_release()` throws an issue 
````
Error:
! No top-level headings found in 'NEWS.md'
````

I think `usethis::use_news_md()` needs to be run before creating the github release if there is no NEWS file, and even before submitting to CRAN as `use_github_release()` will use the SHA from `CRAN_SUBMISSION` so I would change the order in the checklist to reflect that

malcolmbarrett:
If there's no `NEWS.md`, `use_github_release()` is supposed to say "Initial release"  in the body.

https://github.com/r-lib/usethis/blob/8f045dc0a88eac82a602df56e2dd62df90971a04/R/release.R#L325-L329

It doesn't look like that's changed recently

cderv:
Ok I can't. let's forget about that. Must have been a specific weird situation. Sorry 😓

PR #1775, opened 2023-02-23
Rationalize situation re: GitHub releases and release issues and existence of `NEWS.md`

### LINKED ISSUE #1774: `use_github_release()` fails when CRAN submission didn't include `NEWS.md`
Scenario: I released a package to CRAN without a `NEWS.md` and only committed the news file afterwards. I tried to use `use_github_release()` and got the same error from #1691:

```R
> use_github_release()
✔ 'CRAN-SUBMISSION' file found, from a submission on 2023-02-20
• Release name: 'ggmapinset 0.2.3'
• Tag name: 'v0.2.3'
• SHA: 'cb279aefe096c7195008830c1c79c0f6697bdfb6'
Error: No top-level headings found in 'NEWS.md'
```

I stepped through what `use_github_release()`  is doing internally, and traced the problem to this helper:

https://github.com/r-lib/usethis/blob/c9e8ecb873c36273394fe4bda80e524dd05460d2/R/release.R#L363-L387

Basically it tries to download the contents of `NEWS.md` to a temporary file from github as it was at the commit specified in the `CRAN-SUBMISSION` file, but if that commit doesn't have a `NEWS.md` like in my case, then the temp file looks like:

```
> usethis:::read_utf8(news_path)
[1] "{\"message\":\"Not Found\",\"documentation_url\":\"https://docs.github.com/rest/reference/repos#get-repository-content\"}"
```

So the file was filled with an API error message from github when the code was expecting the contents of `NEWS.md` file. It tries to parse it as a news file and hence throws the original error about missing top-level headings.

The helper ought to check for errors from the github API (I imagine other types of errors are possible too) rather than writing them to the output file, and the code should cope with the fact that the released commit might not have included `NEWS.md`.

jennybc:
`get_release_news()` now uses `read_github_file()`, inside a `tryCatch()`. We no longer error if we fail to find `NEWS.md` (but we do urge the user to create it).

In the absence of `NEWS.md`, the checklist in the release issue now also reminds the maintainer to call `usethis::use_news_md()` before submission (vs. after release).

Closes #1774 (and relates to #1687, #1691)

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -366,24 +366,24 @@ get_release_news <- function(SHA = gert::git_info(repo = git_repo())$commit,
 
   if (HEAD == SHA) {
     news_path <- proj_path("NEWS.md")
-  } else {
-    news_path <- withr::local_tempfile()
-    gh <- purrr::possibly(gh_tr(tr), otherwise = NULL)
-    gh(
-      "/repos/{owner}/{repo}/contents/{path}",
-      path = "NEWS.md", ref = SHA,
-      .destfile = news_path,
-      .accept = "application/vnd.github.v3.raw"
-    )
+    if (file_exists(news_path)) {
+      return(news_latest(read_utf8(news_path)))
+    } else {
+      return("Initial release")
+    }
   }
 
-  if (file_exists(news_path)) {
-    news <- news_latest(read_utf8(news_path))
+  news_lines <- read_github_file(
+    tr$repo_spec,
+    path = "NEWS.md",
+    ref = SHA,
+    host = tr$api_url
+  )
+  if (is.na(lines)) {
+    "Initial release"
   } else {
-    news <- "Initial release"
+    lines
   }
-
-  news
 }
 
 cran_version <- function(package = project_name(),
--- a/R/release.R
+++ b/R/release.R
@@ -379,7 +379,7 @@ get_release_news <- function(SHA = gert::git_info(repo = git_repo())$commit,
     ref = SHA,
     host = tr$api_url
   )
-  if (is.na(lines)) {
+  if (is.null(attr(lines, "error"))) {
     "Initial release"
   } else {
     news_latest(lines)
--- a/R/release.R
+++ b/R/release.R
@@ -86,6 +86,7 @@ release_checklist <- function(version, on_cran) {
     if (!on_cran) c(
       "First release:",
       "",
+      todo("`usethis::use_news_md()`", !has_news),
       todo("`usethis::use_cran_comments()`"),
       todo("Update (aspirational) install instructions in README"),
       todo("Proofread `Title:` and `Description:`"),
@@ -106,6 +107,7 @@ release_checklist <- function(version, on_cran) {
       Check if any deprecation processes should be advanced, as described in \\
       [Gradual deprecation](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation)",
       type != "patch" && has_lifecycle),
+    todo("`usethis::use_news_md()`", on_cran && !has_news),
     todo("[Polish NEWS](https://style.tidyverse.org/news.html#news-release)", on_cran),
     todo("`urlchecker::url_check()`"),
     todo("`devtools::build_readme()`", has_readme),
@@ -133,7 +135,6 @@ release_checklist <- function(version, on_cran) {
     todo("`git push`"),
     todo("`usethis::use_github_release()`"),
     todo("`usethis::use_dev_version()`"),
-    todo("`usethis::use_news_md()`", !has_news),
     todo("`git push`"),
     todo("Finish blog post", type != "patch"),
     todo("Tweet", type != "patch"),
--- a/R/release.R
+++ b/R/release.R
@@ -253,7 +253,8 @@ use_github_release <- function(host = deprecated(),
 
   check_github_has_SHA(SHA = dat$SHA, tr = tr)
 
-  news <- get_release_news(SHA = dat$SHA, tr = tr)
+  on_cran <- !is.null(cran_version())
+  news <- get_release_news(SHA = dat$SHA, tr = tr, on_cran = on_cran)
 
   gh <- gh_tr(tr)
   release <- gh(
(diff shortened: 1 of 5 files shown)


R/use-standalone.R
```diff
@@ -38,6 +38,10 @@ use_standalone <- function(repo_spec, file = NULL, ref = NULL, host = NULL) {
   dest_path <- path("R", paste0("import-", file))
 
   lines <- read_github_file(repo_spec, path = src_path, ref = ref, host = host)
+  if (is.na(lines)) {
+    ui_oops("Failed to retrieve content from {ui_path(src_path)}")
+    rlang::cnd_signal(attr(lines, "error"))
+  }
   lines <- c(standalone_header(repo_spec, src_path), lines)
```
jennybc, line 45: This isn't exactly a canonical way to rethrow the error we got from gh, but it gets the job done. Better ideas? Here's how it looks (the backtrace is displayed more nicely in an interactive session):

``` r
devtools::load_all("~/rrr/usethis")
#> ℹ Loading usethis
use_github_file("r-lib/usethis", path = "NOPE")
#> ✔ Saving 'r-lib/usethis/NOPE' to 'NOPE'
#> ✖ Failed to retrieve content from 'NOPE'
#> Error in `gh_process_response()`:
#> ! GitHub API error (404): Not Found
#> ✖ URL not found: <https://api.github.com/repos/r-lib/usethis/contents/NOPE>
#> ℹ Read more at
#>   <https://docs.github.com/rest/reference/repos#get-repository-content>

#> Backtrace:
#>      ▆
#>   1. └─usethis::use_github_file("r-lib/usethis", path = "NOPE")
#>   2.   └─usethis:::read_github_file(...) at usethis/R/use_github_file.R:73:2
#>   3.     └─(purrr::safely(gh::gh))(...) at usethis/R/use_github_file.R:109:2
#>   4.       ├─purrr:::capture_error(.f(...), otherwise, quiet)
#>   5.       │ └─base::tryCatch(...)
#>   6.       │   └─base (local) tryCatchList(expr, classes, parentenv, handlers)
#>   7.       │     └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
#>   8.       │       └─base (local) doTryCatch(return(expr), name, parentenv, handler)
#>   9.       └─gh (local) .f(...)
#>  10.         └─gh:::gh_process_response(raw)
#>  11.           └─gh:::gh_error(response)
#>  12.             └─cli::cli_abort(...)
#>  13.               └─rlang::abort(...)
```

<sup>Created on 2023-02-23 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

R/use_github_file.R
```diff
@@ -110,7 +115,11 @@ read_github_file <- function(repo_spec, path, ref = NULL, host = NULL) {
     .destfile = tf,
     .accept = "application/vnd.github.v3.raw"
   )
-  read_utf8(tf)
+  if (is.null(res$error)) {
+    read_utf8(tf)
+  } else {
+    structure(NA_character_, error = res$error)
+  }
```
jennybc, line 122: My basic goal is to make this more of a true helper and to let the caller decide what to do about a failure to retrieve file contents.

We need that because, for example, there's no guarantee that `NEWS.md` exists when drafting a release (of course, we think there should be, but it would be nice to not error when `NEWS.md` can't be found).

R/release.R
```diff
@@ -366,24 +366,24 @@ get_release_news <- function(SHA = gert::git_info(repo = git_repo())$commit,
...
-    gh(
-      "/repos/{owner}/{repo}/contents/{path}",
-      path = "NEWS.md", ref = SHA,
-      .destfile = news_path,
-      .accept = "application/vnd.github.v3.raw"
-    )
+    if (file_exists(news_path)) {
+      return(news_latest(read_utf8(news_path)))
+    } else {
+      return("Initial release")
```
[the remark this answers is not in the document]
hadley: Or just error? I don't think we can do anything useful without news bullets?

R/use_github_file.R
```diff
@@ -76,6 +76,11 @@ use_github_file <- function(repo_spec,
     ref = ref,
     host = host
   )
+  if (!is.null(attr(lines, "error"))) {
```
hadley, line 79: This doesn't feel quite right to me, because generally errors should bubble up by default. If you don't want to error, _then_ you need to do something special.

R/use_github_file.R
hadley: This doesn't feel quite right to me, because generally errors should bubble up by default. If you don't want to error, _then_ you need to do something special.

Why not use `tryCatch()` with the existing conditions that gh/httr/httr2 throws?

R/release.R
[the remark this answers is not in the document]
arcresu: In my scenario, I hadn't realised that the initial release of my package to CRAN was supposed to include a NEWS file (it's not obvious that "first release" counts as news, I guess). I wanted to create a github release but got an error about missing news. I tried creating NEWS.md retrospectively but the error persisted until I realised that it was using the repository state from the CRAN release commit.

That leaves me kind of stuck. There's no way to recover because I can't change history and alter the released commit. The only option is to create the release manually on github and double check that the commit matches. That's no big deal, but it's slightly jarring because `usethis` already knows exactly which commit I want to release but is refusing to do it because of something I can't change. I think offering to create an empty draft release would be a slightly friendlier fallback than just erroring.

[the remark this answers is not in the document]
hadley: I think it would be better to handle that problem by prompting earlier in the release checklist to call `use_news_md()`.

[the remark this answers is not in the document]
arcresu: Fair enough. In `devtools::release()` it asks "Have you updated `NEWS.md` file?", but I just said yes because in my mind there was no news since it was an initial release. It didn't complain that the news file was missing, and I didn't realise that it would have consequences later.

R/use_github_file.R
jennybc: OK I'm redoing this PR.

Is this difference in the error condition intentional in the latest gh release? We seem to have lost a lot of information (e.g. the HTTP status code) and it's not marked as an `"rlang_error"` or `"error"` at all.

``` r
# pak::pak("r-lib/gh@v1.3.1")
devtools::load_all("~/rrr/usethis")
#> ℹ Loading usethis
packageVersion("gh")
#> [1] '1.3.1'
x <- rlang::catch_cnd(read_github_file("r-lib/usethis", path = "NOPE"))
class(x)
#> [1] "github_error"   "http_error_404" "rlang_error"    "error"         
#> [5] "condition"
```

``` r
devtools::load_all("~/rrr/usethis")
#> ℹ Loading usethis
packageVersion("gh")
#> [1] '1.4.0'
x <- rlang::catch_cnd(read_github_file("r-lib/usethis", path = "NOPE"))
class(x)
#> [1] "httr2_fetch" "condition"
```

hadley: Do you mind filing an issue in gh, and then I'll tackle it tomorrow?

jennybc: It's a weird situation:

https://github.com/r-lib/gh/issues/179

jennybc: OK I have un-confused myself about the gh error classes (it's in the issue).

hadley: Oooh that makes sense.

R/release.R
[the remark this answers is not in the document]
jennybc: > That leaves me kind of stuck. There's no way to recover ... but is refusing to do it because of something I can't change.

This is a phenomenon we've grappled with elsewhere in usethis (mostly because of Git/GitHub stuff) and, as a rule, I try to eliminate these deadends. This PR should now soldier on with release, but provide lots of signals to create `NEWS.md` for future happiness.

#1691 and #1687 show that many folks seems to bang their shins on this specifically.

R/release.R
```diff
@@ -361,29 +363,34 @@ check_github_has_SHA <- function(SHA = gert::git_info(repo = git_repo())$commit,
...
   }
 
-  if (file_exists(news_path)) {
-    news <- news_latest(read_utf8(news_path))
+  if (is.null(news)) {
+    ui_oops("
+      Can't find {ui_path('NEWS.md')} in the released package source.
+      usethis consults this file for release notes.
+      Call {ui_code('usethis::use_news_md()')} to set this up for the future.")
+    if (on_cran) "-- no release notes --" else "Initial release"
```
jennybc, line 390: This feels like it will cause a productive amount of annoyance for anyone who sees it. Their current release will go through, but they're encouraged to create `NEWS.md`. And the release notes can be edited if they're really unhappy with the text I've made up.

jennybc:
@hadley I'm happy with this, but will pause for a bit, in case you have more feedback.

hadley (approved):

Issue #1723, opened 2023-01-11
Re-consider `install_*()` functions, in light of pak


hadley:
Should the README templates switch to using pak? Are there other places we recommend `devtools::install_github()` that should switch to pak?

jennybc:
Just consulted @gaborcsardi. For now, let's keep `devtools::install_github()` in the template. In terms of orienting more towards pak, we lean towards having `install_github()` switch from remotes to pak in the not-too-distant future.

jennybc:
Another idea from @gaborcsardi: perhaps devtools gains a more generic `install_remote()`-type function where the different sources are differentiated via pak's source DSL instead of via a bunch of different functions.

https://pak.r-lib.org/reference/pak_package_sources.html#shorthands

hadley:
I still think we should consider at least switching to suggest pak for tidyverse packages, just as a way as exposing more people to it.

jennybc:
@gaborcsardi Are you feeling more ready to put `pak::pak()` in usethis's README templates? I am.

PR #1777, opened 2023-03-02
Use `remotes::install_github` instead of `devtools::install_github`

multimeric:
`remotes::install_github` should be preferred to `devtools::install_github`. This is because in particular, `remotes` has many less dependencies than `devtools`. This is evident because `remotes` is a dependency of `devtools`, but also has another 19 dependencies: 
https://github.com/r-lib/devtools/blob/a027efdff97e5a88db5abf6a0c88b6a7b1c41195/DESCRIPTION#L20-L42. Users who simply want to install a new package shouldn't be advised to install the very heavy devtools package just to do this.

There are some advantages of the `devtools` version of this function, but in my view they are not worth it. Devtools uses `pkgbuild::with_build_tools` to verify that build tools are installed, `ellipsis::check_dots_used` to check that all `...` args are correctly used, and seemingly supports an input vector of repos. In any case, these last two points are not relevant for the template that `usethis` provides.

### DIFF
--- a/inst/templates/package-README
+++ b/inst/templates/package-README
@@ -29,7 +29,7 @@ You can install the development version of {{{ Package }}} from [GitHub](https:/
 
 ``` r
 # install.packages("devtools")
-devtools::install_github("{{{ github_spec }}}")
+remotes::install_github("{{{ github_spec }}}")
 ```
 {{/on_github}}
 {{^on_github}}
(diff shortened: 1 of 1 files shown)


malcolmbarrett:
This has been discussed here a lot. Currently, we're waiting to switch to pak. See https://github.com/r-lib/usethis/issues/1723

Issue #1683, opened 2022-09-23
For RStudio packages, add posit updates (if needed) to release checklist


hadley:
* update copyright holder in DESCRIPTION
* double check license file uses "{package} authors" not RStudio 
* update email addresses
* update logo (https://github.com/rstudio/hex-stickers)
* update cheatsheets (once they're updated)

jennybc:
My understanding is that we should use Posit Software, PBC in `cph`.

jennybc:
I think we're leaning more towards "Posit PBC" or maybe "Posit, PBC"

tracykteal:
Final guidance is that we should be using

person("Posit Software, PBC", role = c("cph", "fnd"))

jennybc:
That is already embodied in usethis. I think this issue should have been closed a while ago, actually.

PR #1780, opened 2023-03-03
2023 upkeep checklist

### LINKED ISSUE #1604: `use_tidy_upkeep()` tweaks
This is more motivated by Spring Cleaning: maybe we could add more of our standard pkgdown stuff (e.g. the whole template section, maybe with plausible filled out correctly, dev mode set to auto).

### LINKED ISSUE #1614: `use_tidy_upkeep()` should add bullet for `.r`
i.e. scan `R/*` and `tests/testthat/` for any file ending in `.r` and add a bullet recommending to transform to `.R`. (Or possibly also provide some code to help, since getting the invocation right on case insensitive file systems can be tricky).

### LINKED ISSUE #1692: Add "use `rlang::check_*`" to tidy upkeep checklist
Probably worth writing up some notes on how to write your own check functions somewhere:

* Should have `arg` and `call` arguments.
* When type checking, always report what the object actually is.

ateucher:
This is a start to the spring cleaning checklist for 2023. I think there is probably too much here to be reasonably completed in the time we have alloted so I would love some help paring it down (or not if you think it's reasonable)? For some of the items I will add some conditions to the `todo`s to get (mostly) only the relevant items created

Closes #1614.
Closes #1692.
Closes #1604.


jennybc (commented):
Thanks for this.

If we end up feeling like we should pare down, one helpful criterion is whether the `todo()` offers high-level advice vs. directing folks to call a specific function or very detailed instructions. Any `todo()` that is backed by the latter (i.e. it's easier to act on) seems like more of a keeper to me.

R/tidy-upkeep.R
```diff
@@ -111,6 +111,35 @@ if (year <= 2022) {
...
+     "",
+     todo("Update pkgdown site using instructions at <https://tidytemplate.tidyverse.org>"),
+     todo("Ensure pkgdown `development` is `mode: auto` in pkgdown config"),
+     todo("Submit PR [here](https://github.com/rstudio/aws-main/tree/main/zones) adding your site to the appropriate domain (eg., r-lib, tidyverse, tidymodels) and set url in GitHub Settings > Pages > Custom Domain"),
+     todo("Re-publish released site; see [How to update a released site](https://pkgdown.r-lib.org/dev/articles/how-to-update-released-site.html)"),
+     "",
+     todo("Modernize citation files; see updated `use_citation()`"),
+     todo("Update logo (https://github.com/rstudio/hex-stickers)"),
+     todo('Use `pak::pkg_install("org/pkg") in README'),
+     todo("Consider running `use_tidy_dependencies()`"),
```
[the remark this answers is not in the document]
ateucher: I'm only moderately familiar with this - would this be in addition to, or in place of `use_tidy_dependencies()`?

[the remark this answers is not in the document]
jennybc: One of the things `use_tidy_dependencies()` does is to place a standalone file that implements purrr-like functions. And we've now formalized this sort of maneuver into the new function `use_standalone()` (which can also be called directly, more generally). This also introduced new conventions around file names and tracking the dependencies between standalone files, etc.

So it would be good to publicize this within our group.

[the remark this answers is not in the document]
jennybc: Basically, new calls to `use_tidy_dependencies()` will use the new approach.

And folks with other so-called "compat" files lying around could conceivably want to freshen them up.

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+     todo("Update pkgdown site using instructions at <https://tidytemplate.tidyverse.org>"),
+     todo("Ensure pkgdown `development` is `mode: auto` in pkgdown config"),
+     todo("Submit PR [here](https://github.com/rstudio/aws-main/tree/main/zones) adding your site to the appropriate domain (eg., r-lib, tidyverse, tidymodels) and set url in GitHub Settings > Pages > Custom Domain"),
+     todo("Re-publish released site; see [How to update a released site](https://pkgdown.r-lib.org/dev/articles/how-to-update-released-site.html)"),
+     "",
+     todo("Modernize citation files; see updated `use_citation()`"),
+     todo("Update logo (https://github.com/rstudio/hex-stickers); run `use_tidy_logo()`"),
+     todo('Use `pak::pkg_install("org/pkg") in README'),
+     todo("Consider running `use_tidy_dependencies()` and/or replace compat files with `use_standalone()`"),
+     todo("Use `rlang::check_*` (https://github.com/r-lib/usethis/issues/1692)"),
```
ateucher, line 142: I think I need a bit more explanation as to which context this applies

jennybc: Hopefully @hadley can confirm that I have this right and whether he thinks the current spring cleaning is a good time to nudge on this.

But I think this is about placing rlang's type checking standalone file in our packages:

https://github.com/r-lib/rlang/blob/6a5a1ec1128bfeeb081d3b4815ca8e12c488a303/R/standalone-types-check.R

and calling *those* functions instead of idiosyncratic, homegrown solutions in, e.g., `R/utils.R`.

I think the payoff is getting better and more consistent errors for type errors.

Then, it looks like the remaining meat of #1692 (providing guidance on writing type checkers) might be a good issue for https://design.tidyverse.org/.

Maybe `use_tidy_dependencies()` should also place this file?

To pre-emptively answer a question: you might ask why these functions aren't in rlang itself, which most of our packages have in `Imports` already. I believe it's because we are incubating the functions in the standalone file and when they feel pretty battle-tested and stable, they might move into rlang itself.

R/release.R
```diff
@@ -446,7 +446,7 @@ is_rstudio_person_canonical <- function() {
     "fnd" %in% roles &&
     "cph" %in% roles &&
     attr(roles, "appears_in", exact = TRUE) == "given" &&
-    attr(roles, "appears_as", exact = TRUE) == "RStudio"
+    attr(roles, "appears_as", exact = TRUE) == "Posit, PBC"
```
hadley, line 449: Could we update the function name here?

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
     )
   }
 
+  if (year <= 2023) {
+
+    desc <- proj_desc()
+
+    bullets <- c(bullets,
+     "2023",
```
hadley, line 119: Shouldn't there be an extra space in this indent?

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+     "",
+     "Posit updates:",
+     "",
+     todo('Update copyright holder in DESCRIPTION: `person(given = "Posit, PBC", role = c("cph", "fnd"))`',
```
hadley, line 123: Should look like this https://style.tidyverse.org/syntax.html#long-lines, which is easy to do with http://github.com/lionel-/codegrip

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+          rstudio_pkg && !rstudio_person_ok),
+     todo("Double check license file uses '[package] authors' as copyright holder. Run `use_mit_license()`",
+          grepl("MIT", desc$get_field("License"))),
```
hadley, line 126: That check is for an MIT license, which I don't think is what the bullet is getting at.  I suspect you can just drop this one.

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+     todo("Update email addresses *@rstudio.com -> *@posit.co",
+          any(grepl("rstudio", desc$get_authors()))),
+     todo("`usethis::use_tidy_coc()`"),
+     "",
+     "pkgdown:",
+     "",
+     todo("Update pkgdown site using instructions at <https://tidytemplate.tidyverse.org>"),
+     todo("Ensure pkgdown `development` is `mode: auto` in pkgdown config"),
+     todo("Submit PR [here](https://github.com/rstudio/aws-main/tree/main/zones) adding your site to the appropriate domain (eg., r-lib, tidyverse, tidymodels) and set url in GitHub Settings > Pages > Custom Domain"),
+     todo("Re-publish released site; see [How to update a released site](https://pkgdown.r-lib.org/dev/articles/how-to-update-released-site.html)"),
```
hadley, line 136: Didn't we do all these in the last spring cleaning?

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+     "",
+     todo("Modernize citation files; see updated `use_citation()`"),
```
hadley, line 138: Should check if `inst/CITATION` exists

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+     todo("Update logo (https://github.com/rstudio/hex-stickers); run `use_tidy_logo()`"),
```
hadley, line 139: Does `use_tidy_logo()` exist yet?

R/tidy-upkeep.R
hadley: For this bullet, I I think you could say "`use_standalone("r-lib/rlang", "types-check")` instead of home grown argument checkers"?

R/tidy-upkeep.R
```diff
@@ -111,6 +111,42 @@ if (year <= 2022) {
...
+     todo('Use `pak::pkg_install("org/pkg") in README'),
+     todo("Consider running `use_tidy_dependencies()` and/or replace compat files with `use_standalone()`"),
+     todo("Use `rlang::check_*` (https://github.com/r-lib/usethis/issues/1692)"),
+     todo("Change files ending in `.r` to `.R` in R/ and/or tests/testthat/",
+          lowercase_r()),
+     todo("Add alt-text to pictures, plots, etc; see https://www.rstudio.com/blog/knitr-fig-alt/ for examples"),
```
hadley, line 145: Would be useful to have some code to identify where alt text is missing; that's probably a big enough project to happen later.

R/tidy-upkeep.R
ateucher: That's great context, thanks Jenny and Hadley. I'll update the bullet

R/release.R
ateucher: Yes, definitely for this one... I wasn't sure how far down that hole to go in this PR (e.g., `is_rstudio_pkg`, `is_rstudio_cph_or_fnd`, `get_rstudio_roles`) - though I'm happy to if you'd like!

R/tidy-upkeep.R
ateucher: It was on the list for last year though I understood that maybe not all of them were done, and it was suggested to me that they go in this year to finish them up. But I don't have the history so happy to change/remove!

R/tidy-upkeep.R
ateucher: I have a PR for it here: https://github.com/r-lib/usethis/pull/1781

R/tidy-upkeep.R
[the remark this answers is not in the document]
ateucher: Ok, that makes sense. Thanks!

R/tidy-upkeep.R
ateucher: I thought this was for MIT-licensed packages, as that is the pattern in the LICENSE file - based on [this issue](https://github.com/r-lib/usethis/issues/1683) and [Jenny's comment here](https://github.com/r-lib/usethis/pull/1780#discussion_r1123948264)

R/tidy-upkeep.R
ateucher: Ok, so do you think we should leave this bullet or remove it until we have some tooling to help?

R/tidy-upkeep.R
tracykteal: I'd suggested reviewing the ones for last year on the packages we hadn't 'spring cleaned' last year, since there were a few important ones to update. That could be handled in a different way though, by us saying 'check to see if you did the 2022 things', for instance.

tracykteal (approved):
Seems like we have a few things under discussion still, but overall looks good.

R/tidy-upkeep.R
jennybc: Yeah, we definitely have **not** done the pkgdown refresh for all of our packages at this point. But I think the checklist items in the 2022 section mostly cover pkgdown modernization already, so we could let this ride. The item about making a PR re: the subdomain is novel, though.

ateucher: I wonder if adding the subdomain would be better added to the tidytemplate instructions than in the checklist?

R/tidy-upkeep.R
hadley: Leave it, but I think we should track alt-text tooling in another issue. Do you mind creating it?

R/tidy-upkeep.R
hadley: Yeah, I think tidytemplate is a better home for that advice.

R/tidy-upkeep.R
hadley: Ah, I get it now.

R/release.R
jennybc: I can live with doing a bit of this function renaming in this PR. Or doing none of it now and making a separate PR. Or doing all of it. It's an internal matter.

R/tidy-upkeep.R
jennybc: I just merged #1781.

jennybc (approved):
@ateucher I'm approving this now, so once you're happy with it you can just merge. Realistically, it's the kind of thing we might tweak once we start using it in anger, i.e. it's not going to be written in stone.

R/tidy-upkeep.R
ateucher: Happy to - created here: #1783

R/tidy-upkeep.R
ateucher: PR opened at https://github.com/tidyverse/tidytemplate/pull/76

R/tidy-upkeep.R
```diff
@@ -91,23 +91,71 @@ upkeep_checklist <- function(year = NULL,
...
+
+    desc <- proj_desc()
+
+    bullets <- c(bullets,
+      "2023",
+      "",
+      "Posit updates:",
+      "",
+      todo('
+        Update copyright holder in DESCRIPTION:
```
jennybc, line 124:
```suggestion
        Update copyright holder in DESCRIPTION: \\
```

R/release.R
ateucher: I've done most of this renaming in 5b954cec9. I added the small helper function `author_has_rstudio_email()` which will also apply to #1683.

ateucher:
Thanks very much for the review @tracykteal, @jennybc and @hadley!

PR #1781, opened 2023-03-03
use_tidy_logo

### LINKED ISSUE #1728: `use_tidy_logo()` to retrieve logo from our hex stickers repo
http://github.com/rstudio/hex-stickers/

ateucher:
This adds `use_tidy_logo()` to call `use_logo()` on the appropriate hex sticker file at.

A few design decisions I made, which I would love direction on:

- I explicitly included the arguments to `use_logo()` rather than using `...` - it seemed cleaner especially with only two arguments
- I only used the GitHub api to check the presence of the file, and then manually constructed the raw.githubusercontent.com url. I could use the API response directly to get the download url if you think that's neater.
- I put the function in tidyverse.R rather than logo.R

Closes #1728


jennybc (commented):
Thanks, it's looking good.

> I explicitly included the arguments to `use_logo()` rather than using ... - it seemed cleaner especially with only two arguments
> I put the function in `tidyverse.R` rather than `logo.R`

Both choices seem reasonable to me.

> I only used the GitHub api to check the presence of the file, and then manually constructed the raw.githubusercontent.com url. I could use the API response directly to get the download url if you think that's neater.

Have a look at `use_github_file()`. I think you should be able to use it to grab the PNG file and you can rely on it to error informatively for nonexistence, etc.

jennybc:
Hmm, `use_github_file()` is currently pretty text oriented, now that I think about it.

So it might not be an option *today*. But it probably should be helpful here, in some future version.

ateucher:
> Hmm, `use_github_file()` is currently pretty text oriented, now that I think about it.

I just noticed this too. The file contents in the the json response is base64 encoded - I don't think it's worth adding base64enc as a dependency just for this, so I think it's best to read the file from the download url (raw.githubusercontent.com)

R/tidyverse.R
```diff
@@ -401,3 +404,28 @@ base_and_recommended <- function() {
...
+                         owner = owner, repo = repo)
+
+  pkg_png <- purrr::keep(png_contents, ~ .x[["name"]] == paste0(pkg, ".png"))
+
+  if (length(pkg_png) == 0L) {
+    ui_stop("No png logo available for {pkg} at https://github.com/{owner}/{repo}/")
+  }
+
+  url <- pkg_png[[1]]$download_url
+  use_logo(url, geometry = geometry, retina = retina)
```
jennybc, line 430: What do you think of just directly grabbing the file? E.g.,

``` r
tf <- withr::local_tempfile(fileext = ".png")
gh::gh(
  "/repos/rstudio/hex-stickers/contents/PNG/{pkg}.png/",
  pkg = project_name(),
  .destfile = tf,
  .accept = "application/vnd.github.v3.raw"
)
use_logo(tf)
```

ateucher: I like that. It's the accept heard that I was missing! Do you think then just let it 404 and error that way if the package doesn't have a sticker? Or should I catch it and throw something more informative?

jennybc: gh does a pretty nice job of exposing lots of info upon failure, so I would just let its error bubble up if the file's not there.

PR #1785, opened 2023-03-06
Don't add blank lines between bullets

DavisVaughan:
This turns off the "tight" bullet layout used by GitHub and the bullets for the entire section become too spaced apart

i.e.

Screen Shot 2023-03-06 at 2 22 26 PM


ateucher (approved):
Thanks!

PR #1786, opened 2023-03-06
Update renamed function name usage

DavisVaughan:


ateucher (approved):
Thanks! Missed this one!

Issue #1791, opened 2023-03-07
Upkeep for usethis


ateucher:
2023

Necessary:

* [x] Update copyright holder in DESCRIPTION: `person(given = "Posit Software, PBC", role = c("cph", "fnd"))`
* [x] Double check license file uses '[package] authors' as copyright holder. Run `use_mit_license()`
* [x] Update email addresses *@rstudio.com -> *@posit.co
* [x] Update logo (https://github.com/rstudio/hex-stickers); run `use_tidy_logo()`
* [x] `usethis::use_tidy_coc()`

Optional:

* [x] Review 2022 checklist to see if you completed the pkgdown updates
* [x] Prefer `pak::pak("org/pkg")` over `devtools::install_github("org/pkg")` in README
* [x] ~Consider running `use_tidy_dependencies()` and/or replace compat files with `use_standalone()`~
* [x] #1803
* [x] Add alt-text to pictures, plots, etc; see https://posit.co/blog/knitr-fig-alt/ for examples
* [x] `usethis::use_tidy_github_actions()`

ateucher:
usethis already imports purrr and has all of the dependencies from `use_tidy_dependencies()` so that item is not necessary.

PR #1792, opened 2023-03-07
usethis Upkeep 2023 - necessary

ateucher:
Completes the "necessary" items on #1791:

- updates Copyright holder in DESCRIPTION
- updates maintainer email addresses
- Removes old logos and adds official logo
- Update Code of Conduct with new email address
- Recommends `pak::pak("r-lib/usethis")` in README


malcolmbarrett:
I think it's a little weird for usethis to have pak its own readme but not in the readme template. I understand the purpose of changing them more widely in the tidyverse, but usethis feels like a specific case where it's strange

ateucher:
That's a good point. I think the idea is to use it across the tidyverse to get the community to kick the tires more thoroughly, before being fully comfortable recommending it to package authors.

malcolmbarrett:
Yeah, that's exactly why it feels strange to put it where such advice is encoded without actually encoding it

jennybc (approved):

PR #1793, opened 2023-03-08
Make `lowercase_r()` robust to missing directories for `use_tidy_upkeep_issue()`

ateucher:
During spring cleaning I came across a package that didn't have a tests directory so `fs::dir_ls()` fails, stopping the checklist creation.

### DIFF
--- a/R/tidy-upkeep.R
+++ b/R/tidy-upkeep.R
@@ -181,5 +181,7 @@ tidy_minimum_r_version <- function() {
 
 lowercase_r <- function() {
   path <- proj_path(c("R", "tests/testthat"))
+  path <- path[fs::dir_exists(path)]
   length(fs::dir_ls(path, regexp = "[.]r$")) > 0
 }
+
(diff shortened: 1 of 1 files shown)


R/tidy-upkeep.R
```diff
@@ -180,6 +180,8 @@ tidy_minimum_r_version <- function() {
 }
 
 lowercase_r <- function() {
-  path <- proj_path(c("R", "tests/testthat"))
-  length(fs::dir_ls(path, regexp = "[.]r$")) > 0
+  path <- proj_path(c("R", "tests"))
+  path <- path[fs::dir_exists(path)]
+  any(fs::path_ext(fs::dir_ls(path, recurse = TRUE)) == "r")
```
jennybc, line 185: Is it intentional to shift from `dir_ls()`ing `tests/testthat/` to `tests/`?

Re: `path <- path[fs::dir_exists(path)]`. I think I would find a good ol' `if() { ... }` easier to read.

ateucher: No worries - I probably should have pluralized `path(s)` to make it clearer! Thanks!

PR #1796, opened 2023-03-09
Attempt to bubble up failure message from rprojroot::find_root

### LINKED ISSUE #1795: "not in a project" error could be more helpful
https://github.com/r-lib/usethis/blob/654a96576bca12f33aa81cab98fb19b4d0db1f1b/R/proj.R#L87

I tried running `use_package_doc()` in a directory and got this error. As near as I can tell it's being run in the project directory, so the error is a bit frustrating because it's not actionable.

Ideally, we'd spell out exactly what "inside a project or package" means, but I understand that might require exposing implementation details.

In that case, maybe it would be helpful to just refer to `rprojroot::find_root()` instead?

https://github.com/r-lib/usethis/blob/654a96576bca12f33aa81cab98fb19b4d0db1f1b/R/proj.R#L188

MichaelChirico:
Closes #1795 

Alternatively, we might just re-use this snippet:

https://github.com/r-lib/usethis/blob/654a96576bca12f33aa81cab98fb19b4d0db1f1b/R/proj.R#L221-L222

### DIFF
--- a/R/proj.R
+++ b/R/proj.R
@@ -82,12 +82,10 @@ proj_set <- function(path = ".", force = FALSE) {
 
   check_path_is_directory(path)
   new_project <- proj_find(path)
-  if (inherits(new_project, "error")) {
+  if (is.null(new_project)) {
     ui_stop(c(
       "Path {ui_path(path)} does not appear to be inside a project or package.",
-      "",
-      "rprojroot::find_root() returned the following:",
-      conditionMessage(new_project)
+      'Read more in the help for {ui_code("proj_get()")}.'
     ))
   }
   proj_set(path = new_project, force = TRUE)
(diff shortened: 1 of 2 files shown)


jennybc:
These functions are a real hot spot for introducing bugs, so I'm not eager to change them for something that seems to be a very rare problem.

So, yeah, I'd prefer to handle by suggesting folks read more in the help for `proj_get()` or by executing `rprojroot::find_root()` themselves.

MichaelChirico:
Thanks! Minimal version of that done now.

Another suggestion would be a small refactor to move the error down into `proj_get()` to avoid duplication, or, to keep the internal API as-is, move the error into a wrapper function and call it twice. LMK what you prefer.

jennybc:
Thanks!

MichaelChirico:
thanks for the guidance and tidying up, looks great!

PR #1799, opened 2023-03-09
Add alt text to images in PR Helpers vignette

ateucher:
Part of #1791


vignettes/articles/pr-functions.Rmd
```diff
@@ -190,12 +199,19 @@ pr_view(90)
...
+#|     you mean to add 'formidable'? And can you please add a comma 
+#|     at the end? Thanks!"
 knitr::include_graphics("img/pr-functions-comments.png", dpi = 300)
 ```
 
 Being somewhat new to all this, we try to address one of these comments (fix spelling) and neglect the other (forget to add the comma). We make another change and commit it.
 
 ```{r, echo = FALSE}
+#| fig-alt: A screenshot of the Rstudio Git pane, showing the changed line 
+#|     with "formidabel" changed to "formidabel". The file R/adjective.R
```
jennybc, line 213:
```suggestion
#|     with "formidabel" changed to "formidable". The file R/adjective.R
```

jennybc (approved):
I think I caught one error, but with the errors that are intentionally here, it's hard to tell. So check me!

vignettes/articles/pr-functions.Rmd
ateucher: Nice catch, thanks!

ateucher:
The failure on R-devel here is a [due to a patch](https://developer.r-project.org/blosxom.cgi/R-devel/NEWS/2023/03/15#n2023-03-15) just accepted (by @lionel-) that ensures error messages come from the relevant call - in this case `use_lifecycle_badge()`, not the lower level function that produced it (`tolower()`) [when called with no arguments](https://github.com/r-lib/usethis/blob/a336c671b79df60ca287a6ca60eeab5e9b0092be/tests/testthat/test-badge.R#L11). That means that the error is different in R-devel than current and older versions, so I'm not sure what to do with the snapshot?

lionel-:
You could use the `variant` argument of snapshot, but then you have to maintain two different snapshots and juggle with the corresponding R versions as you update them.

In this case a more convenient fix would be to use `rlang::check_required()` so that the message is the same everywhere.

jennybc:
In this case, the simplest solution was to delete the troublesome expectation. It wasn't doing much. That has already happened on `main`, so I've merged that in here.

PR #1800, opened 2023-03-09
Remind to re-`document()`

jennybc:
A bit late in the week, but still worth it, I think.

### DIFF
--- a/tests/testthat/test-tidy-upkeep.R
+++ b/tests/testthat/test-tidy-upkeep.R
@@ -1,4 +1,16 @@
 test_that("upkeep bullets don't change accidentally", {
+  withr::local_options(
+    usethis.description = list(
+      "Authors@R" = utils::person(
+        "Jane", "Doe",
+        email = "jane@rstudio.com",
+        role = c("aut", "cre")
+      ),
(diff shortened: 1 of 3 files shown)


tests/testthat/test-tidy-upkeep.R
```diff
@@ -1,4 +1,16 @@
...
+    usethis.description = list(
+      "Authors@R" = utils::person(
+        "Jane", "Doe",
+        email = "jane@rstudio.com",
+        role = c("aut", "cre")
+      ),
+      License = "MIT + file LICENSE"
+    )
+  )
+  create_local_package()
```
ateucher, line 12: I encountered this failure too, thanks!

ateucher (approved):
Looks great, thanks!

tests/testthat/test-tidy-upkeep.R
ateucher: I fixed it [here](https://github.com/r-lib/usethis/pull/1798/commits/7f16ed2842912a8fe496b4fbf3276706141159bd) as well but that's not ready anyway...

PR #1746, opened 2023-01-20
Reduce number of manual git release steps

hadley:
* `use_release_issue()` checks default branch and pulls automatically.
* `use_github_release()` automatically pushes, if safe.
* `use_github_release()` just creates the release.
* `use_version()` gains `push` argument.

Fixes #1385. Closes #1678.

@jennybc I did not touch `check_current_branch()`, instead focussing on reducing the places in our release checklist where we previously needed to do manual pushes or puls

Issue #1385, opened 2021-03-08
Should use_github_release() offer to push if needed?


hadley:
Rather than erroring, as currently.

jennybc:
I've definitely thought about this.

https://github.com/r-lib/usethis/blob/29192cf7e485c7bb29b64d6394aac13909a30880/R/utils-git.R#L336

Kind of related to some discussion in https://github.com/r-lib/gert/issues/65 (which lead to `gert::git_ahead_behind()`, which we are already using).

I think I haven't leaped into action, just because I haven't done a full analysis of the situation where both local and remote have some unique commits. Do we ... attempt to pull with rebase, then push? What if there are merge conflicts?

hadley:
Maybe we could just push (without asking) if there are local commits missing from the remote, but if there are remote commits missing from local we'd still require the user to fix it up.

hadley:
Slightly related — should `usethis::use_dev_version()` also offer to push? It's almost always what you want.

PR #1802, opened 2023-03-10
Add the `push` argument to `use_dev_version()`

jennybc:
Completes the intent of #1385 and #1746

Without this PR, this item of the release checklist fails:

https://github.com/r-lib/usethis/blob/50074f88a744affc72e05865b7480c704bac502b/R/release.R#L138


hadley (approved):
Oops

PR #1807, opened 2023-03-15
Fix detection of change in dependency

DavisVaughan:
This should fix the failing main branch

In #1789, support was added for the imports field of `use_standalone()` and internally that used `use_package()` to add the required packages to Imports. `use_package()` called `use_dependency()`, and I think `use_dependency()` had a small bug in it, which I'll explain below.

The changes caused this test to fail because of snapshot changes

https://github.com/r-lib/usethis/blob/a336c671b79df60ca287a6ca60eeab5e9b0092be/tests/testthat/test-tidyverse.R#L15-L23

Specifically it was adding this to the snapshots

```r
Execution halted
  "  v Adding '@importFrom lifecycle deprecated' to 'R/tidydeps-package.R'"
  "  v Writing 'NAMESPACE'"
  "  v Writing 'R/import-standalone-purrr.R'"
+ "  * Refer to functions with `rlang::fun()`"
```

But that is weird, because `use_tidy_dependencies()` already called `use_package("rlang")` up above, and its usage of `use_standalone("r-lib/rlang", "purrr")` should not have caused a "change" in package dependency type that should trigger this `how_to_use()` line in `use_package()`, but it obviously does:

https://github.com/r-lib/usethis/blob/a336c671b79df60ca287a6ca60eeab5e9b0092be/R/package.R#L40-L43

For reference the rlang standalone file does not use a versioned version of rlang
https://github.com/r-lib/rlang/blob/9b450274026baa46aa8f4520fdefd87de02e4565/R/standalone-purrr.R#L6

Turns out `use_dependency()` was actually returning `TRUE` here by mistake, because there isn't a `delta == 0L` branch that exits and returns `FALSE` (i.e. the case of package exists and no change in version). I've tweaked `use_dependency()` to have a `changed` variable that keeps track of whether or not something has changed in the DESCRIPTION file (i.e. `changed <- TRUE` always follows a `desc$write()` call).

This change results in a snapshot change that I actually think is a bug fix, because now 2 calls to `use_package("withr")` results in the 2nd call being completely silent

---

I also sort of question the usage of `use_package()` in `use_standalone()` over the quieter `use_dependency()`, but that is a separate issue, since this one is clearly a bug.


jennybc:
I had already sorted the messaging thing out (b9c4b88f3e2d86f4911110c02d17e24ec112c916), but yes this looks like an improvement in the return value.

PR #1808, opened 2023-03-16
Change color of superseded lifecycle badge to blue.

batpigandme:
Erroneously had it as orange in https://github.com/r-lib/usethis/pull/1663.

### DIFF
--- a/man/figures/lifecycle-superseded.svg
+++ b/man/figures/lifecycle-superseded.svg
@@ -9,7 +9,7 @@
     </clipPath>
     <g clip-path="url(#r)">
         <rect width="55" height="20" fill="#555" />
-        <rect x="55" width="73" height="20" fill="#fe7d37" />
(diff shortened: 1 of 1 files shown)


jennybc:
I've made a second change here. Just in case we do more badge updates in the future I want to point this out. The SVGs here:

`man/figures`

are what usethis uses as a normal R package.

The SVGs here:

`inst/templates`

are what usethis places in other packages (or even itself) when their maintainer calls `use_lifecycle()`.

So typically if you're making a badge update the *second* location is even more important than the first.

A good workflow is probably to make any necessary updates to `inst/templates/lifecycle-*.svg`, do `load_all()`, and then `use_lifecycle()` to see if `man/figures/lifecycle-*.svg` gets the intended updates.

Sidebar: the snapshot test failure is unrelated to this PR which I'm about to merge.

PR #1815, opened 2023-03-31
Add `use_example()` function

### LINKED ISSUE #1814: Feature request: `use_example()`
I love this package and it has helped me so much in package development. I have a feature request that I don't think would be too hard to implement but I am struggling to make it work myself.

In the vein of `use_test()`, which creates a test file for `testthat`, I'm imagining `use_example()`, which 1) creates the folder `inst/examples` if it does not exist (possibly giving the user the option to choose a name), 2) creates an R script `ex-foo.R` in that folder corresponding to the R script currently open (like `use_test()` does), and 3) adds `#' @example inst/examples/ex-foo.R` to the `roxygen2` comments for function `foo()`. 

Because writing examples in `roxygen2` blocks can be a little hard because it lacks some of the capabilities that simply writing R code in a script has, the `#' @example {file.R}` functionality makes it easy to write an example in a separate script. `use_example()` would be a shortcut function to facilitate that process and make it easier to use. Thank you for considering.

rossellhayes:
I saw there was a feature request for a `use_example()` function (#1814) and thought I would contribute a function that I've been using in my own workflow.

This function makes it easy to create paired R and example files, using the convention that the examples for `R/foofy.R` should live in `man/examples/example-foofy.R`. It copies existing examples present in a file's `#' @examples` roxygen tags, and gives the user a notification to add a `#' @example man/examples/example-<file>.R` tag.

I tried to design this function in line with other usethis functions, so please let me know if there are any changes I could make that would help it fit better with the rest of the package.

Closes #1814.


jennybc:
This seems like an interesting idea, but we (usethis maintainers, tidyverse/r-lib team) don't currently use external examples ourselves. Our experience adding usethis features that we don't use personally has not been good. These features are especially vulnerable to bit rot, since we won't notice problems in a timely manner and we lack motivation and context when work is needed. This can then lead to a negative experience for new users.

Thanks again for the suggestion, but we've learned everyone is better off if we identify a lack of fit in advance.

PR #1820, opened 2023-04-12
Implement type checkers from rlang standalone-types-check.R

### LINKED ISSUE #1803: use rlang type-checkers instead of home grown argument checkers

ateucher:
Closes #1803 

Replaces the bespoke type-checking with with the checkers from `use_standalone("r-lib/rlang", "types-check")`.

Most are pretty straightforward, but I did rearrange some to try to simplify the logic - particularly in `use_github_action()`.

I used `check_name()` a lot over `check_string()` as it disallows `""`, which I think is warranted, though the previous checks would have allowed it.

I created a `maybe_name()` wrapper around `check_name(x, allow_null = TRUE)` as there are lots of arguments with NULL as a default value.

This PR replaces #1798, which I closed then rebased this branch as `main` had got quite far ahead.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,7 +33,7 @@ Imports:
     lifecycle (>= 1.0.0),
     purrr,
     rappdirs,
-    rlang (>= 1.0.0),
+    rlang (>= 1.0.6),
     rprojroot (>= 1.2),
     rstudioapi,
     stats,
@@ -58,3 +58,5 @@ Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
+Remotes:  
+    r-lib/rlang
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -58,5 +58,3 @@ Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
-Remotes: 
-    r-lib/rlang
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -33,7 +33,7 @@ Imports:
     lifecycle (>= 1.0.0),
     purrr,
     rappdirs,
-    rlang (>= 1.0.6),
+    rlang (>= 1.1.0),
     rprojroot (>= 1.2),
     rstudioapi,
     stats,
@@ -58,5 +58,5 @@ Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
-Remotes:  
+Remotes: 
     r-lib/rlang
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -70,13 +70,20 @@ use_github_action <- function(name = NULL,
                               open = FALSE,
                               badge = NULL) {
 
+  check_name(name, allow_null = TRUE)
+  check_name(ref, allow_null = TRUE)
+  check_name(url, allow_null = TRUE)
+  check_name(save_as, allow_null = TRUE)
+  check_name(readme, allow_null = TRUE)
+  check_bool(ignore)
+  check_bool(open)
+  check_bool(badge, allow_null = TRUE)
+
   if (is.null(url) && is.null(name)) {
     name <- choose_gha_workflow()
   }
 
   if (is.null(url)) {
-    check_string(name)
-    maybe_string(ref)
 
     if (path_ext(name) == "") {
       name <- path_ext_set(name, "yaml")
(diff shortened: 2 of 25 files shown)


DESCRIPTION
```diff
@@ -58,3 +58,5 @@ Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.2.3
+Remotes: 
+    r-lib/rlang
```
jennybc, line 62: Since rlang 1.1.0 has been released, what is it here that requires installing the dev version?

R/github-actions.R
```diff
@@ -70,13 +70,18 @@ use_github_action <- function(name = NULL,
...
-    name <- choose_gha_workflow()
-  }
+  maybe_name(name)
+  maybe_name(ref)
+  maybe_name(url)
+  maybe_name(save_as)
+  maybe_name(readme)
+  check_bool(ignore)
+  check_bool(open)
+  check_bool(badge, allow_null = TRUE)
```
jennybc, line 80: I haven't done an analysis, but assume you have. Moving these all up here and as a result, softening some `check_`s to `maybe_`s must change behaviour somewhat. But presumably not in any case we're actually worried about.

R/write.R
```diff
@@ -72,7 +74,10 @@ write_union <- function(path, lines, quiet = FALSE) {
 #' @param overwrite Force overwrite of existing file?
 #' @export
 write_over <- function(path, lines, quiet = FALSE, overwrite = FALSE) {
-  stopifnot(is.character(lines), length(lines) > 0)
+  check_name(path)
+  check_character(lines)
```
jennybc, line 78: I think we lost the check that lines has at least one element FWIW.

tests/testthat/test-course.R
```diff
@@ -154,7 +154,7 @@ test_that("create_download_url() works", {
 })
 
 test_that("normalize_url() prepends https:// (or not)", {
-  expect_error(normalize_url(1), "is\\.character.*not TRUE")
+  expect_error(normalize_url(1), "must be a valid name")
```
jennybc, line 157: Let's go ahead and make this a snapshot test.
```suggestion
  expect_snapshot(normalize_url(1), error = TRUE)
```
Usually, when I touch existing regex-y tests of error messages, I convert them to a snapshot test (vs. updating the regex). Most of them originate from a time before snapshot tests.

tests/testthat/test-tutorial.R
```diff
@@ -2,8 +2,8 @@ test_that("use_tutorial() checks its inputs", {
   skip_if_not_installed("rmarkdown")
 
   create_local_package()
-  expect_error(use_tutorial(), "no default")
-  expect_error(use_tutorial(name = "tutorial-file"), "no default")
+  expect_error(use_tutorial(), "must be a valid name")
+  expect_error(use_tutorial(name = "tutorial-file"), "must be a valid name")
```
jennybc, line 6: Would also convert these to snapshots.

(Probably all 3 of these are far from being "load-bearing" expectation and could also be considered for deletion. But let's err on the side up updating over deleting.)

jennybc (approved):
I left a few comments, but I trust you to see it through.

DESCRIPTION
ateucher: Whoops - `use_standalone("r-lib/rlang", "types-check")` did that, it slipped by me.

R/write.R
ateucher: Hmmm good point. Should `check_character()` perhaps have a `length` or `min_length` argument? Or should there be a separate `check_length()` function?

R/github-actions.R
ateucher: I've looked at this one pretty carefully and I'm fairly certain it doesn't meaningfully change the behaviour - In most cases it just consolidates the logic:

- `name` and `ref` are only used when `url` is `NULL` and the case when `name` is `NULL` is caught on line 84, and `ref` is `NULL` on line 90. Otherwise `maybe_name()` ensures they are length 1 character vectors.
- `maybe_name(url)` combined with `if (is.null(url))` on line 82 ensures that if `url` is not `NULL` it will be a length 1 character vector.
- The cases where `save_as` and `readme` being null are already caught on lines 100 and 120 respectively.

R/write.R
jennybc: I think I've looked for that myself, when converting other packages, and I did not find anything.

ateucher: I'll open an issue in `rlang`, I think it would be useful

ateucher: https://github.com/r-lib/rlang/issues/1618

ateucher: Added back a `stopifnot()` for now

ateucher:
I think this is ready now @jennybc - would you like me to merge this one before @lionel-'s #1821?

PR #1821, opened 2023-04-13
Support prefixed standalone files

lionel-:
- Allow passing in prefixed standalone file names like `aaa-standalone-foo`
- While I was in there I added type-checking for `file` using the rlang standalone file.

### DIFF
--- a/R/use_standalone.R
+++ b/R/use_standalone.R
@@ -70,7 +70,7 @@ use_standalone <- function(repo_spec, file = NULL, ref = NULL, host = NULL) {
   }
 
   src_path <- path("R", file)
-  dest_path <- path("R", paste0("import-", file))
+  dest_path <- path("R", as_standalone_dest_file(file))
 
   lines <- read_github_file(repo_spec, path = src_path, ref = ref, host = host)
   lines <- c(standalone_header(repo_spec, src_path), lines)
@@ -149,6 +149,10 @@ as_standalone_file <- function(file) {
     file <- paste0("standalone-", file)
   }
   file
+}
+
+as_standalone_dest_file <- function(file) {
+  gsub("standalone-", "import-standalone-", file)
 }
 
 standalone_header <- function(repo_spec, path) {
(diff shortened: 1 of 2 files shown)


ateucher:
Just a heads up - this overlaps somewhat with #1820 - really just the addition of the standalone files, and the modification of `maybe_string()`

lionel-:
oops, then let's merge yours first and I'll rebase here.

ateucher:
No worries. Mine is wider ranging (implementing the type-checkers) so I'm happy to wait for this to be merged and then rebase mine... whatever works best for you and @jennybc

ateucher:
@lionel- my PR was merged, so you should be good to rebase here...

jennybc:
Could @lionel- and/or @ateucher advance this along? I want to release usethis soon, but now I have to wait for testthat (planned for May 1), so there's a bit of a pause. So we should be able to get this into the usethis release. But I have to immerse myself in R Packages for a couple days, doing copy edits.

R/use_standalone.R
```diff
@@ -53,6 +53,9 @@
 #' }
 use_standalone <- function(repo_spec, file = NULL, ref = NULL, host = NULL) {
   check_is_project()
+  check_string(file, allow_null = TRUE)
+  check_string(ref, allow_null = TRUE)
+  check_string(host, allow_null = TRUE)
```
ateucher, line 58: I think the checks for `file` and `host` could be `maybe_name()`, allowing `NULL` but not empty or `NA`? I think `ref` can allow `""`

ateucher (approved):

ateucher:
I think this looks great - but just for my education what is the rationale/use case for the  prefixed standalone files?

R/use_standalone.R
lionel-: I've used `maybe_name()` for all three to keep things simple.

lionel-:
> what is the rationale/use case for the prefixed standalone files?

This is for standalone files that implement build-time dependencies for other files in the repo. To avoid having to specify a collation order for the files, you can use the `aaa-` prefix since files are sourced in alphabetical order.

lionel-:
Sorry for the delay! Thanks for reviewing @ateucher

ateucher:
Makes perfect sense - thanks!

PR #1825, opened 2023-04-13
Public upkeep checklist

### LINKED ISSUE #1794: Create `use_upkeep_issue()` to facilitate Spring Cleaning by package developers
Create `use_upkeep_issue()` based on `use_tidy_upkeep_issue()` for broader use, containing a few generic items, and enable/encourage the use of custom upkeep bullets similar to how `use_release_issue()` allows users to provide custom release bullets.

ateucher:
This adds a public function `use_upkeep_issue()` similar to `use_tidy_upkeep_issue()` for general (i.e. non-tidyverse) use.

I factored out the internals of `use_tidy_upkeep_issue()` into `make_upkeep_issue()` which is called by both exported functions. 

I created a default checklist for `use_upkeep_issue()` with some good practices... I would love some feedback on what should be included there. Many of them are one-time only fixes, but most are conditionally populated so should only show up when needed.

Next step would be to add the ability to add custom upkeep tasks similar to `use_release_checklist()`; I can just modify `release_extra_bullets()` for that. The simpler approach would be to just advise in the documentation of `use_upkeep_issue()` to modify the issue once it's opened.

Closes #1794.


R/tidy-upkeep.R
```diff
@@ -30,16 +68,67 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+      !uses_testthat_edition(3)
+    ),
+    todo("
+        Align the names of `R/` files and `test/` files for workflow happiness.
+        The docs for `usethis::use_r()` include a helpful script.
+        `usethis::rename_files()` may be be useful."),
+    todo("usethis::use_github_action('check-standard')", !uses_github_actions()),
+    todo(
+      "Consider changing default branch from `master` to `main`",
+      git_default_branch() == "master"
```
ateucher, line 101: I'm wondering if the machinations required in the tests to make this work are a code smell, and we should just put this bullet in unconditionally?

R/tidy-upkeep.R
```diff
@@ -3,9 +3,43 @@
...
+#'   of them will generally make your package better,
+#'   easier to maintain, and more enjoyable for your
+#'   users. Some of the tasks are meant to be performed
+#'   only once (and once completed shouldn't show up in
+#'   subsequent lists), and some should be reviewed
+#'   periodically. The tidyverse team uses a similar
+#'   function [use_tidy_upkeep_issue()] for our annual
+#'   package Spring Cleaning.
+#'
+#' @param year Optional year you are performing the upkeep
```
jennybc, line 25: Maybe elaborate a bit more on the impact of `year`. In fact, I'm actually not entirely clear on how `year` is going to play out in the non-tidy version of this function.

R/tidy-upkeep.R
```diff
@@ -30,16 +68,67 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+upkeep_checklist <- function(year = NULL) {
+
+  bullets <- c(
+    year,
+    "",
+    todo("`usethis::use_readme_rmd()`", !file_exists("README.Rmd")),
+    todo("`usethis::use_roxygen_md()`", !is_true(uses_roxygen_md())),
+    todo("`usethis::use_github_links()`", !has_github_links()),
+    todo("`usethis::use_pkgdown_github_pages()`", !uses_pkgdown()),
+    todo("`urlchecker::url_check()`"),
```
jennybc, line 80: This one seems out of place (url checking), in that it's something you do repeatedly vs. (almost) only once. And it's also in the release checklist. I'm not sure if the answer is relocation or removal, but one of those seems appropriate.

R/tidy-upkeep.R
```diff
@@ -30,16 +68,67 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+    todo(
+      "
```
jennybc, line 82: Nitpick: I think the `todo()` below ("Align the names ...") uses styling more consistent with what's elsewhere in usethis.

R/tidy-upkeep.R
```diff
@@ -30,16 +68,67 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+        `usethis::use_package_doc()`
+        Consider letting usethis manage your `@importFrom` directives here.
+        `usethis::use_import_from()` is handy for this.",
+      !has_package_doc()
+    ),
+    todo(
+      "
+        `usethis::use_testthat(3)` and upgrade to 3e, \\
+        [testthat 3e vignette](https://testthat.r-lib.org/articles/third-edition.html)",
+      !uses_testthat_edition(3)
```
jennybc, line 92: I haven't done any analysis of this but ... what's going to happen to someone who just has no tests at all?

tests/testthat/_snaps/tidy-upkeep.md
```diff
@@ -69,3 +69,51 @@
...
+      writeLines(upkeep_checklist())
+    Output
+      
+      * [ ] `usethis::use_readme_rmd()`
+      * [ ] `usethis::use_github_links()`
+      * [ ] `usethis::use_pkgdown_github_pages()`
+      * [ ] `urlchecker::url_check()`
+      * [ ] `usethis::use_package_doc()`
+      Consider letting usethis manage your `@importFrom` directives here.
+      `usethis::use_import_from()` is handy for this.
```
jennybc, line 84: Does this mean it's going to look wonky in the GitHub issue too?

tests/testthat/test-tidy-upkeep.R
```diff
@@ -12,6 +12,41 @@ test_that("upkeep bullets don't change accidentally", {
...
+    tidy_upkeep_checklist(posit_pkg = TRUE, posit_person_ok = FALSE)
   ))
 })
+
+test_that("upkeep bullets don't change accidentally",{
+  skip_if_no_git_user()
+  withr::local_options(usethis.description = NULL)
+  create_local_package()
+
+  # Need to add git so can detect presence of master branch
```
jennybc, line 24: I would use the new `testthat::local_mocked_bindings()` to eliminate the fussiness. I.e. just fake having a git repo with a `master` branch.

https://testthat.r-lib.org/reference/local_mocked_bindings.html

tests/testthat/test-tidy-upkeep.R
```diff
@@ -12,6 +12,41 @@ test_that("upkeep bullets don't change accidentally", {
...
+  use_git()
```
jennybc, line 25: Even if you stick with this (which I think we should not), you'd have to go to even more effort to make sure the default branch is `master`. Locally this fails for me because I have configured my default for the default branch to be `main`. More reason to design this away.

tests/testthat/test-tidy-upkeep.R
```diff
@@ -12,6 +12,41 @@ test_that("upkeep bullets don't change accidentally", {
...
+
+  expect_snapshot(writeLines(
+    upkeep_checklist()
+  ))
+
+  # Add some files to test conditional todos
+  use_code_of_conduct("jane.doe@foofymail.com")
+  use_citation()
+  use_testthat(3)
+  desc::desc_set_dep("lifecycle")
```
jennybc, line 38: `use_package()`? In the spirit of always using the usethis function to do X, if usethis can do X.

jennybc (commented):
I made some comments to move things along and start some discussion.

I'm not sure if `tidy-upkeep.R` is the right file for this anymore, although I'm not sure it's not. But worth thinking about.

R/tidy-upkeep.R
jennybc: Below I suggest using some mocks in the tests.

R/tidy-upkeep.R
```diff
@@ -3,9 +3,43 @@
...
 #'   the default, will give you a full set of actions to perform.
 use_tidy_upkeep_issue <- function(year = NULL) {
-  check_is_package("use_tidy_upkeep_issue()")
+  make_upkeep_issue(year = year, tidy = TRUE)
+}
+
+#' Open an issue in GitHub with a checklist of upkeep
+#' tasks
+#'
+#' @description This opens an issue in your package
```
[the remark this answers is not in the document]
ateucher: Sounds good. Weird, I had my margin width set to 58...??

R/tidy-upkeep.R
```diff
@@ -16,7 +50,11 @@ use_tidy_upkeep_issue <- function(year = NULL) {
     }
   }
 
-  checklist <- upkeep_checklist(year)
+  checklist <- if (tidy) {
+    tidy_upkeep_checklist(year)
+  } else {
+    upkeep_checklist(year)
+  }
```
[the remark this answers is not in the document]
ateucher: Yeah I always go back and forth on this as well. I could _just_ fit it on one line, or switch to the internal assignment?

R/tidy-upkeep.R
ateucher: I was just thinking about this as well. It could default to the current year, and maybe put the year in the issue title, and not in the body? Or remove entirely... I don't think it's serving much of a purpose as is

R/tidy-upkeep.R
ateucher: I think since it's in the release checklist I'll remove it

R/tidy-upkeep.R
ateucher: It will return a `FALSE` and thus the bullet will appear - I did think about having two separate bullets - one to tell people to set up tests if they didn't have any at all, and one to upgrade testthat if it is < 3, but I thought this one would capture both sufficiently.

tests/testthat/test-tidy-upkeep.R
ateucher: Oh great, I haven't used this yet!

R/tidy-upkeep.R
```diff
@@ -3,9 +3,43 @@
 #' @param year Approximate year when you last touched this package. If `NULL`,
 #'   the default, will give you a full set of actions to perform.
 use_tidy_upkeep_issue <- function(year = NULL) {
-  check_is_package("use_tidy_upkeep_issue()")
+  make_upkeep_issue(year = year, tidy = TRUE)
+}
+
+#' Open an issue in GitHub with a checklist of upkeep
+#' tasks
```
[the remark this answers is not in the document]
ateucher: Better, thanks!

tests/testthat/_snaps/tidy-upkeep.md
ateucher: Here's an [example issue](https://github.com/ateucher/rmapshaper/issues/144). I used the same formatting of the those bullets as in the original `tidy_upkeep_checklist()` - some have line breaks, and some have all on one line...

ateucher:
> I made some comments to move things along and start some discussion.
> 
> I'm not sure if `tidy-upkeep.R` is the right file for this anymore, although I'm not sure it's not. But worth thinking about.

I was thinking of just doing `rename_files("tidy-upkeep", "upkeep"). Or do you want to separate the tidy checklist from the public checklist? It looks like currently there are some (most?) `use_tidy_*()` functions in `tidyverse.R`, and some with the parent functions (e.g., `use_tidy_github_labes()`.

R/tidy-upkeep.R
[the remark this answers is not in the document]
jennybc: In this case, it's fine either way, I think.

tests/testthat/test-tidy-upkeep.R
ateucher: That's really nice - thanks for the pointer!

R/tidy-upkeep.R
jennybc: It probably wouldn't be too hard to distinguish "uses testthat (but not 3e)" from "doesn't use testthat" and customize the bullet accordingly.

These are pretty different audiences, I think, so maybe worth giving them their own bullet point, hyperlink, etc.

tests/testthat/_snaps/tidy-upkeep.md
jennybc: That all looks fine to me.

jennybc:
As long as you give the file(name) issue some thought, I'm OK with what you choose. Let's just make sure we aren't introducing some new convention. Do what seems most consistent with what we've done for other functions with generic and tidy versions. (This could happen for `use_release_issue()` at some point as well.)

R/tidy-upkeep.R
ateucher: Ok, great, I'll flesh that out.

ateucher: See fcc9ffd and 6992786

R/tidy-upkeep.R
jennybc: I don't think it makes much sense to treat year the way we currently do in the tidy upkeep issue. Yes, I think it makes sense to timestamp the issue in some sense, either in its title or the body, with the date and possible the usethis version.

R/upkeep.R
```diff
@@ -30,16 +54,76 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
 }
 
-upkeep_checklist <- function(year = NULL,
-                             posit_pkg = is_posit_pkg(),
-                             posit_person_ok = is_posit_person_canonical()) {
-  year <- year %||% 2000
+upkeep_checklist <- function() {
+
+  bullets <- c(
+    todo("`usethis::use_readme_rmd()`", !file_exists("README.Rmd")),
```
jennybc, line 60:
```suggestion
    todo("`usethis::use_readme_rmd()`", !file_exists(proj_path("README.Rmd"))),
```

R/upkeep.R
```diff
@@ -30,16 +54,76 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+    todo("
+         `usethis::use_testthat()`. \\
+         Learn more about testing at https://r-pkgs.org/tests.html",
+         !uses_testthat()
+    ),
+    todo("
+        `usethis::use_testthat(3)` and upgrade to 3e, \\
+        [testthat 3e vignette](https://testthat.r-lib.org/articles/third-edition.html)",
+        uses_old_testthat_edition(current = 3)
+    ),
```
jennybc, line 79: Yeah, I like this.

jennybc (approved):
It's not quite done, but I think we're in agreement on what "done" means.

This should also get a NEWS bullet.

R/tidy-upkeep.R
ateucher: For the non-tidy function, I removed it from the body and put it in the issue title, defaulting to the current year. I like the idea about tagging it with the usethis version as well. For the tidy upkeep issue, the year is baked into the format quite heavily... though perhaps we should also default to the current year rather than all years as it is now? We also talked about using a custom field in the `DESCRIPTION` to tag the latest upkeep for tidyverse packages, is that something to look at now?

ateucher: Ok, year now always appears in upkeep issue title, defaults to current year

* for `use_tidy_upkeep_issue()`, if `year` is specified it will be used in the title and used to select the subset of checklist items included (as before). If `year` is `NULL`, the current year will be in the issue title, but all current and previous checklists will be included.
* for `use_upkeep_issue()`, `year` is only used in the title, as there are not multiple year-based checklists.

I added a footer to the issue, ala reprex: f7cce2298d8. To avoid spurious changes in snapshots, I mocked the `checklist_footer()` function in the tests... I'm not sure if I'm now abusing `local_mocked_bindings()`!

ateucher:
I think this is almost done now @jennybc. 

An example of a non-tidy issue: https://github.com/ateucher/rmapshaper/issues/149
And a tidy issue:https://github.com/r-lib/usethis/issues/1827

R/upkeep.R
```diff
@@ -16,30 +39,90 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
-  year <- year %||% 2000
+upkeep_checklist <- function() {
 
+  bullets <- c(
+    todo("`usethis::use_readme_rmd()`", !file_exists(proj_path("README.Rmd"))),
+    todo("`usethis::use_roxygen_md()`", !is_true(uses_roxygen_md())),
+    todo("`usethis::use_github_links()`", !has_github_links()),
+    todo("`usethis::use_pkgdown_github_pages()`", !uses_pkgdown()),
+    todo("
+        `usethis::use_package_doc()`.
```
jennybc, line 64:
```suggestion
        `usethis::use_package_doc()`
```
Spotted this when looking at one of the example issues that you linked.

R/upkeep.R
```diff
@@ -16,30 +39,90 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+    todo(
+      "Consider changing default branch from `master` to `main`",
+      git_default_branch() == "master"
+    ),
+    todo("`usethis::use_code_of_conduct()`", !has_coc()),
+    todo(
+      "Modernize citation files; see `usethis::use_citation()`",
+      has_citation_file()
+    ),
+    todo("Remove check environments section from `cran-comments.md`"),
```
jennybc, line 93: I wonder if a typical community member will know what this is about?

Also, should it be conditional on `file_exists(proj_path("cran-comments.md"))`?

Perhaps, if `cran-comments.md` exists, we say something like: "Remove any description of check environments from `cran-comments.md`.

R/upkeep.R
```diff
@@ -193,3 +281,36 @@ lowercase_r <- function() {
...
+  env <- env %||% safe_pkg_env()
+
+  if (env_has(env, "upkeep_bullets")) {
+    c(paste0("* [ ] ", env$upkeep_bullets()), "")
+  } else {
+    ""
+  }
+}
+
+checklist_footer <- function(tidy) {
```
jennybc, line 310: This looks nice in the example issues!

jennybc (approved):

R/upkeep.R
ateucher: Yeah, I like that - thanks!

tests/testthat/test-upkeep.R
```diff
@@ -10,6 +10,8 @@ test_that("tidy upkeep bullets don't change accidentally", {
     )
   )
   create_local_package()
+  # mock test footer so doesn't change with date and usethis version
+  local_mocked_bindings(checklist_footer = function(tidy) "test footer")
```
jennybc, line 14: This is one way to do this and it seems fine to me.

You can also use the `transform =` argument of `expect_snapshot()`. If you search "our" code for that, you should be able to surface many examples of what that sort of usage looks like.

ateucher: Oh I like that! I'll do that even if it's just for learning (but I also think it's a bit neater)

tests/testthat/_snaps/upkeep.md
```diff
@@ -68,4 +68,45 @@
...
+# upkeep bullets don't change accidentally
+
+    Code
+      writeLines(upkeep_checklist())
+    Output
+      * [ ] `usethis::use_readme_rmd()`
+      * [ ] `usethis::use_github_links()`
+      * [ ] `usethis::use_pkgdown_github_pages()`
+      * [ ] `usethis::use_package_doc()`
+      Consider letting usethis manage your `@importFrom` directives here. `usethis::use_import_from()` is handy for this.
```
hadley, line 82: Is this indented correctly?

tests/testthat/_snaps/upkeep.md
```diff
@@ -68,4 +68,45 @@
...
+      * [ ] `usethis::use_testthat()`. Learn more about testing at https://r-pkgs.org/tests.html
```
hadley, line 83: Needs `<>`?

R/upkeep.R
```diff
@@ -16,30 +38,97 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+      "
+      `usethis::use_testthat(3)` and upgrade to 3e, \\
+      [testthat 3e vignette](https://testthat.r-lib.org/articles/third-edition.html)",
+      uses_old_testthat_edition(current = 3)
+    ),
+    todo("
+      Align the names of `R/` files and `test/` files for workflow happiness. \\
+      The docs for `usethis::use_r()` include a helpful script. \\
+      `usethis::rename_files()` may be be useful."),
+    todo("`usethis::use_github_action('check-standard')`", !uses_github_actions()),
```
hadley, line 84:
```suggestion
    todo("`usethis::use_github_action('check-standard')`"),
```

Just because you probably want to update it when doing spring cleaning

R/upkeep.R
```diff
@@ -16,30 +38,97 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+    todo(
+      "Consider changing default branch from `master` to `main`",
+      git_default_branch() == "master"
+    ),
+    todo("`usethis::use_code_of_conduct()`", !has_coc()),
+    todo(
+      "Modernize citation files; see `usethis::use_citation()`",
+      has_citation_file()
+    ),
```
hadley, line 93: I don't think this needs to be a bullet since `R CMD check` now warns about it?

R/upkeep.R
```diff
@@ -16,30 +38,97 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+    todo(
+      "Remove any description of check environments from `cran-comments.md`",
+      file_exists(proj_path("cran-comments.md"))
+    ),
```
hadley, line 97: Could we make this a little more precise?

R/upkeep.R
```diff
@@ -16,30 +38,97 @@ use_tidy_upkeep_issue <- function(year = NULL) {
...
+    todo(
+      "
+      Use lifecycle instead of artisanal deprecation messages, as described \\
+      in [Communicate lifecycle changes in your functions](https://lifecycle.r-lib.org/articles/communicate.html)",
+      !proj_desc()$has_dep("lifecycle")
+    ),
```
hadley, line 103: This feels a bit more tidyverse-y than the other bullets.

ateucher: Do you mean the suggestion to use lifecycle itself, or the way it's worded?

tests/testthat/_snaps/upkeep.md
ateucher: It does render correctly in the issue (starts a newline as part of the same bullet): https://github.com/ateucher/rmapshaper/issues/155

R/upkeep.R
jennybc: I suspect he means that this is arguably too opinionated for the public checklist.

Do we really mean to condition on the _absence_ of lifecycle?

You could imagine conditioning on the _presence_ of lifecycle, i.e. target only folks who have clearly opted in to the lifecycle lifestyle. And making it more about checking if anything needs to be advanced.

ateucher: Ah yes, I like the idea of conditioning on the presence of lifecycle. There is then overlap with `use_release_issue()`, though is there a case to be made that upkeep is a better place for it than as you're preparing for a release?  Happy to remove it from here though if just residing in the release checklist is the way to go.

hadley: I think remove it. Opt-in into the lifecycle lifestyle is not something for every package (IMO most packages don't need that level of deprecation handling)

R/upkeep.R
hadley: We should check for the wording used in `cran-comments.md`

R/upkeep.R
ateucher: Sounds good to me.

R/upkeep.R
ateucher: Great, done.

PR #1830, opened 2023-04-20
use_news_md() adds a development header if it is a dev version

### LINKED ISSUE #1708: Should use_news_md() add a development header if it is a dev version?
Currently, if using `use_news_md()` to create a NEWS file for a package with a development version, one gets the header "package-name package-version". I wonder whether that header should actually be "package-name (development version)"

``` r
withr::local_options(usethis.quiet = TRUE)
temp_dir <- withr::local_tempdir()
dir.create(file.path(temp_dir, "package"))
usethis::create_package(file.path(temp_dir, "package"))

withr::with_dir(file.path(temp_dir, "package"), {
  usethis::use_news_md()
  readLines("NEWS.md")
})
#> [1] "# package 0.0.0.9000"                                     
#> [2] ""                                                         
#> [3] "* Added a `NEWS.md` file to track changes to the package."
```

<sup>Created on 2022-12-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'd expect the header to be "# package (development version)".

ateucher:
Closes #1708.

This modifies `use_news_md()` to set the version in the header to `"(development version)"` if `proj_version()` looks like a development version.

It also addresses @hadley's comment [here](https://github.com/r-lib/usethis/issues/1708#issuecomment-1398420607) and sets the first bullet to "Initial submission" when it looks like a "new" package - i.e., the package is not on CRAN and it has an early version number.


R/news.R
```diff
@@ -9,13 +9,23 @@
 #' @export
 use_news_md <- function(open = rlang::is_interactive()) {
   check_is_package()
-  desc <- proj_desc()
+
+  ver <- proj_version()
+  on_cran <- !is.null(cran_version())
+  version_string <- if (is_dev_version(ver)) "(development version)" else ver
+
+  if (!on_cran && package_version(ver) <= package_version("0.1.0")) {
```
ateucher, line 17: This choice of version number was a bit arbitrary... maybe we could just condition on whether or not it's on CRAN? I know that's not everybody's final destination but I think most of this and related tooling is CRAN-oriented...

R/version.R
```diff
@@ -152,3 +151,8 @@ use_c_version <- function(ver) {
 
   write_utf8(version_path, lines)
 }
+
+is_dev_version <- function(version) {
```
jennybc, line 155:
```suggestion
is_dev_version <- function(version = proj_version()) {
```

Maybe sensible/handy?

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # usethis (development version)
 
+* `use_news_md()` now places "(development version)" in the header of `NEWS.md`
```
jennybc, line 3: Does this need a bit more qualification / rewording? `use_news_md()` did already place "(development version)" in most relevant scenarios, I think, so the news is that we're making it more robust.

jennybc (approved):

NEWS.md
jennybc: Oh wait, I am thinking of subsequent edits to `NEWS.md`. Got it. Disregard.

R/news.R
ateucher: Great, I like all of that

NEWS.md
ateucher: Yep, I had to keep remembering that too!

PR #1831, opened 2023-04-20
modernize `use_rscloud_badge()`

### LINKED ISSUE #1670: Update expected paths in `use_rscloud_badge()`
I recently learned that the default paths for RStudio Cloud projects, both for individual projects and for projects inside shared spaces, have been changed such that the word `content` is now used instead of `project` in the path. As far as I can tell, this is an [undocumented](https://rstudio.cloud/learn/whats-new) change.

```html
# old
https://rstudio.cloud/project/<project-id>
https://rstudio.cloud/spaces/<space-id>/project/<project-id>

# new
https://rstudio.cloud/content/<project-id>
https://rstudio.cloud/spaces/<space-id>/content/<project-id>
```

This is a breaking change for the function I contributed in #1584 and #1589, `use_rscloud_badge()`, which expects that paths passed via the `url` argument will include `project`. While valid paths that use `project` will be redirected by RStudio Cloud to the same path using `content` instead, all new paths obtained from RStudio Cloud, via the dropdown `(...) > Share Project Link` or from copy/pasting the URL, will use `content` and thus will cause `use_rscloud_badge()` to return an error:

``` r
library(httr2)
library(usethis)
library(withr)

# 1. define helper function ----
check_rscloud_url <- function(url) {
  req_perform(request(url))
}

# 2. confirm 'project' containing url is a valid url ----
project_url <- "https://rstudio.cloud/project/3584129"
check_rscloud_url(project_url)
#> <httr2_response>
#> GET https://rstudio.cloud/project/3584129
#> Status: 200 OK
#> Content-Type: text/html
#> Body: In memory (643 bytes)

# 3. confirm 'content' containing url is a valid url ----
content_url <- "https://rstudio.cloud/content/3584129"
check_rscloud_url(content_url)
#> <httr2_response>
#> GET https://rstudio.cloud/content/3584129
#> Status: 200 OK
#> Content-Type: text/html
#> Body: In memory (643 bytes)

# 4. confirm use_rscloud_badge(content_url) returns an error ----
with_tempdir({
  create_project(tempdir())
  use_readme_md()
  use_rscloud_badge(content_url)
})
#> ✔ Setting active project to '/private/var/folders/qf/
#> yt27gctx33x4pvc255t0w9zc0000gp/T/RtmpggYsT7'
#> ✔ Creating 'R/'
#> ✔ Writing a sentinel file '.here'
#> • Build robust paths within your project via `here::here()`
#> • Learn more at <https://here.r-lib.org>
#> ✔ Setting active project to '/private/var/folders/qf/yt27gctx33x4pvc255t0w9zc0000gp/T/RtmpggYsT7'
#> ✔ Writing 'README.md'
#> Error: `usethis::use_rscloud_badge()` requires a link to an existing RStudio Cloud project of the form 'https://rstudio.cloud/project/<project-id>' or 'https://rstudio.cloud/spaces/<space-id>/project/<project-id>'.
```

<sup>Created on 2022-09-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

This would be a relatively easy fix and I'd be happy to submit a PR if helpful. One question for the maintainers would be, since paths that contain `project` are still technically valid URLs, should we soft deprecate the use of paths containing `project` or should we require that all paths use `content` moving forward?

ateucher:
Closes #1670.

This modernizes the paths to Posit Cloud projects used in creating the badge in `use_rscloud_badge()`. I think it makes sense to deprecate that function in favour of `use_posit_cloud_badge()`.

One open question (bigger than this PR) is that of the logo in the badge. shields.io uses icons from https://simpleicons.org/ - there isn't currently one for posit (but is for RStudio). Currently I've just made the badge look like this:

[Launch Posit Cloud](https://posit.cloud/content/5821761)

I've used the blue from Posit's colour palette; green and orange are pretty common for other badges, though the wine colour could also be nice:

[Launch Posit Cloud](https://posit.cloud/content/5821761)


R/badge.R
```diff
@@ -136,28 +138,38 @@ use_binder_badge <- function(ref = git_default_branch(), urlpath = NULL) {
...
+      'https://posit.cloud/spaces/<space-id>/content/<project-id>'.")
   }
 
   invisible(TRUE)
 }
 
+#' @rdname badges
+#' @export
+use_rscloud_badge <- function(url) {
+  lifecycle::deprecate_warn("2.2.0", "use_rscloud_badge()",
```
ateucher, line 168: I made an assumption on the package version here...

R/badge.R
```diff
@@ -136,28 +138,40 @@ use_binder_badge <- function(ref = git_default_branch(), urlpath = NULL) {
...
+  spaces_url <- "posit[.]cloud/spaces"
   if (grepl(project_url, url) || grepl(spaces_url, url)) {
-    img <- "https://img.shields.io/badge/launch-cloud-75aadb?style=flat&logo=rstudio"
-    use_badge("Launch RStudio Cloud", url, img)
+    # TODO: Get posit logo hosted at https://github.com/simple-icons/simple-icons/
+    # and add to end of img url as `?logo=posit` (or whatever slug we get)
+    img <- "https://img.shields.io/badge/launch-posit%20cloud-447099?style=flat"
+    use_badge("Launch Posit Cloud", url, img)
   } else {
     ui_stop("
```
jennybc, line 155: You could switch to `usethis_abort()` since you're touching this. But it's fine if you don't.

tests/testthat/_snaps/badge.md
```diff
@@ -7,3 +7,43 @@
       ! `stage` must be one of "experimental", "stable", "superseded", or "deprecated", not "eperimental".
       i Did you mean "experimental"?
 
+# use_rscloud_badge() handles bad and good input
+
+    Code
+      use_posit_cloud_badge()
+    Condition
+      Error in `is.factor()`:
+      ! argument "url" is missing, with no default
```
jennybc, line 16: This is sort of gross. Maybe we should throw in a `check_name(url)`.

jennybc (approved):
LGTM. I made a few observations, but I'm happy with however you resolve.

R/badge.R
jennybc: Seem like the most likely version number, yes.

jennybc:
BTW I like the blue badge.

tests/testthat/_snaps/badge.md
jennybc: Actually, R 4.3 automatically does something much better, which is nice.

ateucher: That is nicer - but then the snapshot will be different for different versions. `check_name()` is a good idea

R/badge.R
ateucher: Done. The cli inline markup is very nice!

PR #1834, opened 2023-04-26
Allow appending urls in `use_github_links()`

### LINKED ISSUE #1805: Should `use_github_links()` really be on the release checklist?
It comes with `usethis::use_release_issue()`

It tries to change `URL: https://vctrs.r-lib.org/` to `URL: https://github.com/r-lib/vctrs`, which is not what we wanted.

@hadley said this is likely a more community focused function, so we probably don't need it in our release checklist?

ateucher:
Closes #1805.

Allows appending of the GitHub repo url to existing urls in the `URL` `DESCRIPTION` field. Appending is the default behaviour (i.e., when `overwrite = FALSE`). When `overwrite = TRUE` it will overwrite the old value with the GitHub repo url. 

This is slightly different behaviour than before as there is now no option to leave the `URL` field as-is: it will either be replaced, or appended to. I think this is ok, since why would you use the function if you didn't want to update the field? The alternative would be to allow three options: append, overwrite, or leave as is...

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -150,13 +150,10 @@ tidy_desc <- function(desc) {
   try(desc$normalize(), silent = TRUE)
 }
 
-# 2021-10-10, while adding use_description_list(), I moved this helper here
-#
-# this helper feels out-of-sync with current usethis practices around active
-# project and how overwrite is handled
-#
-# I won't change use_description_field() now, but use_description_list() is
-# implemented differently, more in keeping with our current style
+# 2023-05-03
+# TODO: Rationalize this with proj_desc_field_append(); being careful
+# about the interactions between overwrite and append behaviour,
+# https://github.com/r-lib/usethis/pull/1834
 use_description_field <- function(name, value, overwrite = FALSE) {
   # account for `value`s produced via `glue::glue()`
   value <- as.character(value)
--- a/R/description.R
+++ b/R/description.R
@@ -176,7 +176,7 @@ use_description_field <- function(name, value, overwrite = FALSE) {
   }
 
   ui_done("Setting {ui_field(name)} field in DESCRIPTION to {ui_value(value)}")
-  desc$set(name, value)
+  desc$set_list(name, value)
   desc$write()
 
   invisible()
--- a/R/github.R
+++ b/R/github.R
@@ -225,8 +225,14 @@ use_github_links <- function(auth_token = deprecated(),
   res <- gh("GET /repos/{owner}/{repo}")
 
   desc <- proj_desc()
-  if (!res$html_url %in% desc$get_urls()) {
-    use_description_field("URL", res$html_url, overwrite = overwrite)
+  existing_urls <- desc$get_urls()
+  if (!res$html_url %in% existing_urls) {
+    if (overwrite) {
+      urls <- res$html_url
+    } else {
+      urls <- paste(existing_urls, res$html_url, sep = ", ", collapse = ", ")
+    }
+    use_description_field("URL", urls, overwrite = TRUE)
   }
 
   use_description_field(
--- a/R/github.R
+++ b/R/github.R
@@ -219,22 +219,20 @@ use_github_links <- function(auth_token = deprecated(),
   }
 
   check_is_package("use_github_links()")
-  tr <- target_repo(github_get = TRUE)
 
-  gh <- gh_tr(tr)
-  res <- gh("GET /repos/{owner}/{repo}")
+  gh_url <- github_url_from_git_remotes()
 
   desc <- proj_desc()
   existing_urls <- desc$get_urls()
 
-  if (!res$html_url %in% existing_urls) {
-    urls <- c(existing_urls, res$html_url)
+  if (!gh_url %in% existing_urls) {
+    urls <- c(existing_urls, gh_url)
     use_description_field("URL", urls, overwrite = overwrite)
   }
 
   use_description_field(
     "BugReports",
-    glue("{res$html_url}/issues"),
+    glue("{gh_url}/issues"),
     overwrite = overwrite
   )
 
--- a/R/github.R
+++ b/R/github.R
@@ -226,14 +226,10 @@ use_github_links <- function(auth_token = deprecated(),
 
   desc <- proj_desc()
   existing_urls <- desc$get_urls()
-  
+
   if (!res$html_url %in% existing_urls) {
-    if (overwrite || length(existing_urls) == 0) {
-      urls <- res$html_url
-    } else {
-      urls <- paste(existing_urls, res$html_url, sep = ", ", collapse = ", ")
-    }
-    use_description_field("URL", urls, overwrite = TRUE)
+    urls <- c(existing_urls, res$html_url)
+    use_description_field("URL", urls, overwrite = overwrite)
   }
 
   use_description_field(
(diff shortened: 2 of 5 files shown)


R/github.R
```diff
@@ -225,8 +225,15 @@ use_github_links <- function(auth_token = deprecated(),
...
-    use_description_field("URL", res$html_url, overwrite = overwrite)
+  existing_urls <- desc$get_urls()
+  
+  if (!res$html_url %in% existing_urls) {
+    if (overwrite || length(existing_urls) == 0) {
+      urls <- res$html_url
+    } else {
+      urls <- paste(existing_urls, res$html_url, sep = ", ", collapse = ", ")
+    }
+    use_description_field("URL", urls, overwrite = TRUE)
```
ateucher, line 236: This kind of short-circuits the overwrite argument but has the intended effect, without rewriting `use_description_field()`

tests/testthat/test-github.R
```diff
@@ -0,0 +1,45 @@
+test_that("use_github_links populates empty URL field", {
+  create_local_package()
+  use_git()
```
ateucher, line 3: It seems like a lot to activate git in every test... I could instead mock `target_repo()`?

jennybc (commented):
In `use_github_links()` and in the tests, I suggest leaning more into desc, since we're dealing with a list-y field and it has a specific helper for this (i.e. `desc$set_urls()`). That way we can just deal with the URLs as a character vector and let desc deal with commas and whitespace.

I totally see where you're coming from re: appending even if `overwrite = FALSE` ("why would you use the function if you didn't want to update the field?"). And yet it feels confusing if we start to let `overwrite` mean different things in different contexts. It seems most consistent with what we do elsewhere to ask if we can update the `URL` field if `overwrite = FALSE`.

tests/testthat/test-github.R
jennybc: You are going where no one's gone before re: testing these functions!

Yeah, if you can simplify the mocking that would be nice. This is why these functions haven't had tests before, which is also a position we could retreat to.

tests/testthat/test-github.R
```diff
@@ -0,0 +1,45 @@
+test_that("use_github_links populates empty URL field", {
```
jennybc, line 1: Similar tests start with:

```
skip_on_cran()
skip_if_no_git_user()
local_interactive(FALSE)
```

I'm not sure there's any reason to skip on CRAN, if you're mocking the GitHub calls, so maybe that's not needed.

`skip_if_no_git_user()` is needed if `use_git()` stays (some CRAN machines don't have git or a git user? anyway, this skip exists for a reason 😬 ).

`local_interactive(FALSE)` is just something I leave behind in tests where that is helpful during interactive tinkering.

tests/testthat/test-github.R
```diff
@@ -0,0 +1,45 @@
...
+  create_local_package()
+  use_git()
+  local_mocked_bindings(
+    gh_tr = function(tr) {
+        function(endpoint, ...) list(html_url = "https://github.com/USER/REPO")
+      },
+    target_repo = function(github_get) NULL
+  )
+  # when an existing field, and overwrite = FALSE, should append
+  use_description_field("URL", "https://existing.url", overwrite = TRUE)
```
jennybc, line 41: You could have 2 URLs to start with, then let `use_github_links()` add a 3rd. There's a whole bunch of packages with 3+ URLs (including bcdata and gapminder): the pkgdown site, some URL about the data source or API wrapped, the GitHub URL. So those are good real-world inspiration.

So, combined with leaning into desc, the setup would be something like:

```
d <- proj_desc()
d$set_urls(c("https://bcgov.github.io/bcdata/", "https://catalogue.data.gov.bc.ca/"))
d$write()
```

R/github.R
```diff
@@ -225,8 +225,15 @@ use_github_links <- function(auth_token = deprecated(),
...
   desc <- proj_desc()
-  if (!res$html_url %in% desc$get_urls()) {
-    use_description_field("URL", res$html_url, overwrite = overwrite)
+  existing_urls <- desc$get_urls()
+  
+  if (!res$html_url %in% existing_urls) {
+    if (overwrite || length(existing_urls) == 0) {
+      urls <- res$html_url
+    } else {
+      urls <- paste(existing_urls, res$html_url, sep = ", ", collapse = ", ")
```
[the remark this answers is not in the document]
ateucher: Whoops. Unfortunately his won't work either if `existing_urls` is greater than length 1. More fodder for the argument to use `desc`

tests/testthat/test-github.R
ateucher: I've taken your advice on using the various `skip_` and `local_` helpers. I'm not sure how to simplify the mocking other than making more helpers?

ateucher:
Thanks for the review @jennybc. One other thing I noticed while poking around is that `use_description_field()` and `proj_desc_field_append()` do almost exactly the same things except the former has the `overwrite` argument (which errors when it's `FALSE` and there is an existing value in the field. I could try to rationalize them, or open as a separate issue? I believe `use_github_links()` is actually the only function that calls `use_description_field`.

tests/testthat/test-github.R
ateucher: I found the existing `github_url_from_git_remotes()`, which does the job of several lines, and is easy to mock https://github.com/r-lib/usethis/pull/1834/commits/db041b19074462ae56770148d0a5a0d2b2ca501f

R/description.R
```diff
@@ -176,7 +176,7 @@ use_description_field <- function(name, value, overwrite = FALSE) {
   }
 
   ui_done("Setting {ui_field(name)} field in DESCRIPTION to {ui_value(value)}")
-  desc$set(name, value)
+  desc$set_list(name, value)
```
jennybc, line 179: So this changes `use_description_field()` to always use the list form? :thinking:

In which case, have you taken a look at `use_description_list()`, which I see mentioned in a comment above, just to contemplate this pair of functions together?

ateucher:
@jennybc I think rationalizing `use_description_field()` and `proj_desc_field_append()` warrants its own PR as they are both used widely and the interaction between append and overwrite behaviour needs some thought - I'm happy to start it. I think it might be worth merging this now and then I'll follow on shortly with that PR - what do you think?

jennybc:
Yeah, it was good to make #1837 as a separate PR. I would lean towards merging that then rebasing/merging here.

tests/testthat/test-github.R
```diff
@@ -0,0 +1,42 @@
+test_that("use_github_links populates empty URL field", {
+  local_interactive(FALSE)
+  create_local_package()
+  use_git()
+  skip_if_no_git_user()
```
jennybc, line 5: I would place the skip at the very top. I think that's a general principle, i.e. if you're going to skip, get out as early as possible.
And, in this specific case, definitely before `use_git()`.

tests/testthat/test-github.R
```diff
@@ -0,0 +1,42 @@
...
+    proj_desc()$get_field("BugReports"),
+    "https://github.com/OWNER/REPO/issues"
+    )
+})
+
+test_that("use_github_links() aborts or appends URLs when it should", {
+  local_interactive(FALSE)
+  create_local_package()
+  use_git()
+  skip_if_no_git_user()
```
jennybc, line 24: Ditto here

jennybc (approved):

tests/testthat/test-github.R
ateucher: 🤦

PR #1835, opened 2023-04-27
Tweak the new mocking helpers

jennybc:
I'm asking for review mostly to have a discussion, to get us on the same page and to start to germinate ideas on recommended mocking practices. I converted usethis over to testthat's mocking functions yesterday and this is a bit of polishing.


tests/testthat/helper-mocks.R
```diff
@@ -0,0 +1,52 @@
+mock_cran_version <- function(version, .env = caller_env()) {
```
jennybc, line 1: We now have mocking helpers for mocks that come up in multiple places.

I think if you have a bunch of these, it probably does make sense to give them their own helper file.

I'm a bit ambivalent about some of these helpers that are super simple (i.e. fit on one line). Would it be better to have `local_mocked_binding()` calls in the tests themselves, in terms of "keeping the reader in the test"? OTOH if the helper has a good name, the reader probably doesn't feel the need to go read the definition.

ateucher: I think if they are used commonly I like having the helper functions if they are clearly and consistently named (see my comment below). I don't think it takes the reader out of the test as it's clearly obvious what they're doing.

tests/testthat/helper-mocks.R
```diff
@@ -0,0 +1,52 @@
...
+}
+
+mock_check_is_package <- function(.env = caller_env()) {
+  local_mocked_bindings(
+    check_is_package = function(...) invisible(),
+    .env = .env
+  )
+}
+
+mock_rstudio_not_available <- function(.env = caller_env()) {
```
ateucher, line 16: This is the only helper that isn't of the pattern `mock_[function_name]()`. I think having them all fit that pattern helps the reader understand exactly what function is being mocked. What about `mock_rstudio_available_false()`? Are there any cases where it's mocked to be `TRUE`? It could then be:

```r
mock_rstudio_available <- function(is_available) {
  local_mocked_bindings(
    rstudio_available = function(...) is_available, 
    .env = .env
  )
}
```

tests/testthat/helper-mocks.R
ateucher: I also like having them in their own helper file, it makes discovering them much easier.

ateucher (approved):
High level comment - I love the mocking infrastructure, it's really intuitive and easy to use - so much so that it might be easy to overuse/abuse in some cases. I wonder if some simple guidance on when and when not to mock would be useful (though that probably would belong in testthat rather than here).

Two types of cases I can see guidance being useful are:

1. When should I be mocking a function vs setting up some local infrastructure (e.g., [here](https://github.com/r-lib/usethis/pull/1834#discussion_r1178480726).
2. Are there cases where I am mocking when I should be actually testing the thing? I.e., to the extent that it can be tested, make sure it tested properly elsewhere.

A couple of comments/suggestions inline below

tests/testthat/helper-mocks.R
jennybc: Yeah it started as `mock_rstudio_available()` and it made `rstudio_available()` always return `FALSE`.

Then I thought I should expose the hard-wired value as the argument (`FALSE`, not available, vs. `TRUE` available).

But then I realized we _never_ needed the `TRUE` version.

So then I settled on `mock_rstudio_not_available()`.

ateucher: That makes sense, it just stood out to me as different - `mock_check_installed()`, `mock_ui_yeah()` etc all have a hard-wired return value but don't signal it in the function name. 

Maybe the difference is that that some mockers are used to breeze by tricky bits (like UI menus, checking github repos etc) so you always want them to return a value that allows the code to run, but some might to be emulate different environments?

tests/testthat/helper-mocks.R
hadley: I like them being in a helper function, but I don't love the names because they only tell you what function is being mocked, not what the mocked function will do. I think they also should have `local` in their name? 

Brainstorming:
* `local_cran_version()`
* `local_disable_install_check()`
* `local_pretend_is_package()`
* ...

hadley (approved):

jennybc:
I've renamed them all like `local_*()`.

I took @ateucher's suggestion re: `local_rstudio_available().

PR #1836, opened 2023-05-02
Add some upkeep bullets

ateucher:
* use_tidy_description()
* Explain use_github_action() usage

### DIFF
--- a/R/upkeep.R
+++ b/R/upkeep.R
@@ -58,6 +58,7 @@ upkeep_checklist <- function() {
     todo("`usethis::use_roxygen_md()`", !is_true(uses_roxygen_md())),
     todo("`usethis::use_github_links()`", !has_github_links()),
     todo("`usethis::use_pkgdown_github_pages()`", !uses_pkgdown()),
+    todo("`usethis::use_tidy_description()`"),
     todo(
       "
       `usethis::use_package_doc()`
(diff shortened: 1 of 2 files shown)


R/upkeep.R
```diff
@@ -81,7 +82,9 @@ upkeep_checklist <- function() {
       Align the names of `R/` files and `test/` files for workflow happiness. \\
       The docs for `usethis::use_r()` include a helpful script. \\
       `usethis::rename_files()` may be be useful."),
-    todo("`usethis::use_github_action('check-standard')`"),
+    todo("
+      `usethis::use_github_action('check-standard')`
+      Set up or update GitHub Actions for automated package checking."),
```
jennybc, line 87: This is just a "nice to have", but the same would apply to, say, a pkgdown workflow. Empirically, it feels like a big chunk of "my GHA is not working" ends up being resolved by downloading the latest workflow config.

jennybc (approved):

R/upkeep.R
ateucher: I definitely find that as well, that was the motivation for adding this sentence. I feel like the check actions are the most susceptible to breaking more frequently, but easy enough to add a bullet for `use_github_action('pkgdown')` to update that workflow, conditional on using pkgdown. Maybe `use_github_action('test-coverage')` as well (especially since a lot of R CMD check workflows have test coverage in them, and that would be lost with the newer check action).

PR #1837, opened 2023-05-03
Rationalize use_description_field() & proj_desc_field_append()

ateucher:
`use_description_field()` and `proj_desc_field_append()` do very similar things. This combines them into one function with arguments `overwrite` and `append` to allow the existing behaviour of the original two functions:

* `proj_desc_field_append()` -> `proj_desc_field_update()`
* remove `use_description_field()`
* add overwrite and append arguments to `proj_desc_field_update()`
* replace use of `use_description_field()` and `proj_desc_field_append()` with `proj_desc_field_update()`, setting overwrite and append arguments as appropriate
* update tests

NB: There will be merge conflicts after #1834 is merged.

### DIFF
--- a/R/description.R
+++ b/R/description.R
@@ -149,36 +149,3 @@ tidy_desc <- function(desc) {
   # Wrap in a try() so it always succeeds, even if user options are malformed
   try(desc$normalize(), silent = TRUE)
 }
-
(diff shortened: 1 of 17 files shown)


jennybc (approved):
Thanks! This looks like a really good rationalization of this little family of helpers.

Do we merge this first, then rebase and adjust #1834?

PR #1843, opened 2023-05-19
fix: make use_testthat() work for brand-new package

maelle:
Coming from https://github.com/r-lib/desc/issues/117#issuecomment-1554499421

I understand why this doesn't happen in usethis test where testthat edition field is already set by `create_local_package()`; after digging a bit I see that prior to https://github.com/r-lib/usethis/commit/747931bdee400be9969b7ab75af390c05ff2e7a9 there was `as.character()`. Now I am at peace, having the whole story :jigsaw: :relieved: 

@krlmlr

### DIFF
--- a/tests/testthat/_snaps/proj-desc.md
+++ b/tests/testthat/_snaps/proj-desc.md
@@ -1,4 +1,4 @@
-# proj_desc_field_append() only messages when adding
+# proj_desc_field_update() only messages when adding
 
     Code
       proj_desc_field_update("Config/Needs/foofy", "alfa", append = TRUE)
@@ -9,4 +9,11 @@
       proj_desc_field_update("Config/Needs/foofy", "bravo", append = TRUE)
     Message
       v Adding 'bravo' to Config/Needs/foofy
+
+# proj_desc_field_update() works with multiple values
+
+    Code
+      proj_desc_field_update("Config/Needs/foofy", c("alfa", "bravo"), append = TRUE)
+    Message
+      v Adding 'alfa', 'bravo' to Config/Needs/foofy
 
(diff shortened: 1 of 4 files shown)


krlmlr (approved):

ateucher (approved):
Thanks @maelle this is great. Along with this change I think it's also worth making `proj_desc_field_update()` stricter [here](https://github.com/r-lib/usethis/blob/e05947f0e49244d7dcb13105a48e32718eb59db5/R/proj-desc.R#L34) so this would be handled there rather than in `desc` (probably using `check_string(value)`).

Happy to make that change myself if you'd prefer.

jennybc:
Should `proj_desc_field_update()` be in charge of coercing to character?

ateucher:
I did certainly wonder about that but thought it may end up receiving something unexpected that goes awry when trying to coerce, whereas the calling functions would be more suited to formatting the value properly since they know about them. Right now `use_testthat_impl()` is the only function that calls it with a non-character value.

jennybc:
OK I'm happy for you decide. So, yes, I think it would be good to check for character input, as you suggest above.

ateucher:
I'm going to merge this - thanks @maelle!

tests/testthat/_snaps/proj-desc.md
```diff
@@ -10,3 +10,10 @@
     Message
       v Adding 'bravo' to Config/Needs/foofy
 
+# proj_desc_field_update() works with multiple values
+
+    Code
+      proj_desc_field_update("Config/Needs/foofy", c("alfa", "bravo"), append = TRUE)
+    Message
+      v Adding 'alfa', 'bravo' to Config/Needs/foofy
```
jennybc, line 18: I'm not sure this would be worth the fiddliness. But arguably the set logic should be applied *before* we message. Maybe just worth keeping in mind if we touch this again.

PR #1845, opened 2023-05-26
Have `use_release_issue()` remind tidy packages to bump minimum R version

### LINKED ISSUE #1762: Can `use_release_issue()` remind us if min R version needs a bump?
(for packages that follow the tidyverse versioning policy)

ateucher:
Suggest bumping minimum R version in release issue when package minimum R version is older than current tidy. 

Fixes #1762.


R/release.R
```diff
@@ -77,6 +77,7 @@ release_checklist <- function(version, on_cran) {
   has_readme <- file_exists(proj_path("README.Rmd"))
   has_github_links <- has_github_links()
   is_posit_pkg <- is_posit_pkg()
+  tidy_min_r_version <- numeric_version(tidy_minimum_r_version())
```
jennybc, line 80: Do you think `tidy_minimum_r_version()` should return an instance of `"numeric_version"`? 🤔 Caveat: I haven't done a survey of where else this is used.

jennybc (approved):

R/release.R
ateucher: Yes, that makes sense. It is only otherwise used in `tidy_upkeep_checklist()` where it doesn't hurt. 

On that note, I didn't add the conditional check to the bullet in `tidy_upkeep_checklist()` as it's just buried in the 2021 checklist, and since it's now in the release checklist maybe we won't also do it during Spring Cleaning?

PR #1846, opened 2023-05-26
Documentation for usethis options

### LINKED ISSUE #1768: Document all of usethis's options in one place
usethis needs a help topic like these:

* https://gargle.r-lib.org/reference/gargle_options.html
* https://tibble.tidyverse.org/reference/tibble_options.html

(Some of) the options to document there:

* `usethis.destdir`
* `usethis.description`
* `usethis.protocol`
* `usethis.quiet`
* `usethis.overwrite`

ateucher:
Closes #1768.

There is duplication with the [setup article](https://usethis.r-lib.org/articles/usethis-setup.html#store-default-values-for-description-fields-and-other-preferences), however I think it's worth it to also have a man page for them.


R/usethis-package.R
```diff
@@ -10,6 +10,68 @@
...
+#' If the built-in defaults don't suit you, set one or more of these options.
+#' Typically, this is done in the `.Rprofile` startup file, which you can open
+#' for editing with [edit_r_profile()] - this will set the specified options for
+#' all future R sessions. Your code will look something like:
+#'
+#' ```
+#' options(
+#'   usethis.description = list(
+#'     `Authors@R` = 'person("Jane", "Doe", email = "jane@example.com",
+#'       role = c("aut", "cre"), comment = c(ORCID = "YOUR-ORCID-ID"))',
```
jennybc, line 28: You may be copying this from somewhere else, in which this comment applies to that too. Are we sure this line break is OK, i.e. it doesn't hurt anything?

In my `.Rprofile` I do it like so and I seem to remember that line length was part of the motivation:

```
options(
  ...,
  usethis.description = list(
    "Authors@R" = utils::person(
      "Jennifer", "Bryan",
      email = "jenny@posit.co",
      role = c("aut", "cre"),
      comment = c(ORCID = "0000-0002-6983-2759")
     ),
     License = "MIT + file LICENSE"
    ),
  ...
)
```

R/usethis-package.R
```diff
@@ -10,6 +10,68 @@
...
+#'     License = "MIT + file LICENSE",
+#'     Version = "0.0.0.9000"
```
jennybc, line 30: Is this `Version` the default? If so, I'd remove it.

If you're looking around for more examples of stuff to set, I think `usethis.overwrite = TRUE` is a good one to publicize.

R/usethis-package.R
```diff
@@ -10,6 +10,68 @@
...
+#'   ),
+#'   usethis.destdir = "/path/to/folder/", # for use_course(), create_from_github()
+#'   usethis.protocol = "ssh", # Use ssh git protocol
+#'   usethis.quiet = TRUE, # suppress user-facing messages
```
jennybc, line 34: Ooh, this makes me a little nervous, in terms of people copy/pasting, because I think it's pretty rare that someone would want this.

R/usethis-package.R
```diff
@@ -10,6 +10,68 @@
...
+#' )
+#' ```
+#'
+#' @section Options for the usethis package:
+#'
+#' - `usethis.description`: customize the default content of new `DESCRIPTION`
+#'   files by setting this option to a named list.
+#'   If you are frequent package developer, it is worthwhile to pre-configure
```
jennybc, line 42:
```suggestion
#'   If you are a frequent package developer, it is worthwhile to pre-configure
```

jennybc (approved):

R/usethis-package.R
ateucher: Yeah, I blindly copied it directly from the `use_description()` docs - I'll fix in both places.

R/usethis-package.R
ateucher: I'll swap out `usethis.overwrite = TRUE` for `usethis.quiet`

R/usethis-package.R
ateucher: This definitely looks nicer

ateucher: Oh actually it's similar in R Packages too: https://r-pkgs.org/setup.html#personal-startup-configuration

PR #1847, opened 2023-05-27
Move up "add link to blog post" release bullet

### LINKED ISSUE #1761: `use_release_issue()` bullet for `Add link to blog post in pkgdown news menu` isn't in a useful order
Pulling from dplyr's 1.1.0 release bullet, we had

```
Wait for CRAN...

* [x] Accepted :tada:
* [x] `git push`
* [x] `usethis::use_github_release()`
* [x] `usethis::use_dev_version()`
* [x] `git push`
* [x] Finish blog post
* [x] Tweet
* [x] Add link to blog post in pkgdown news menu
```

That last bullet doesn't seem to be in the right place because by that point you are already on the "dev" pkgdown site, so you won't ever see a link to that release's blog post until the _next_ release of the package

ateucher:
Closes #1761. 

As pointed out by @DavisVaughan in the linked issue, there is no perfect place for this bullet in the checklist. I made the assumption that the user won't be pushing the released version to GitHub prior to CRAN acceptance, so I moved the bullet to after CRAN acceptance and before `use_github_release()`, which will push to GitHub and thus build the released version of the pkgdown site, with the link to the blog post in the pkgdown news. The link will 404 until the post is published but that shouldn't usually be a long delay.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -137,7 +137,7 @@ release_checklist <- function(version, on_cran) {
     "Wait for CRAN...",
     "",
     todo("Accepted :tada:"),
-    todo("Add link to blog post in pkgdown news menu", type != "patch"),
+    todo("Add preemptive link to blog post in pkgdown news menu", type != "patch"),
     todo("`usethis::use_github_release()`"),
     todo("`usethis::use_dev_version(push = TRUE)`"),
     todo("`usethis::use_news_md()`", !has_news),
--- a/R/release.R
+++ b/R/release.R
@@ -137,12 +137,12 @@ release_checklist <- function(version, on_cran) {
     "Wait for CRAN...",
     "",
     todo("Accepted :tada:"),
+    todo("Add link to blog post in pkgdown news menu", type != "patch"),
     todo("`usethis::use_github_release()`"),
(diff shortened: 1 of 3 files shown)


R/release.R
```diff
@@ -137,12 +137,12 @@ release_checklist <- function(version, on_cran) {
     "Wait for CRAN...",
     "",
     todo("Accepted :tada:"),
+    todo("Add link to blog post in pkgdown news menu", type != "patch"),
```
DavisVaughan, line 140:
```suggestion
    todo("Add preemptive link to blog post in pkgdown news menu", type != "patch"),
```

Maybe a little keyword to hint at the fact that the link won't work yet?

DavisVaughan (approved):
I guess we are technically guessing a little at what the blog post will be titled, but that seems reasonable

jennybc (approved):
I won't be surprised if we end up discussing this again, because there really is no great solution. Which ever way we rig it, there's probably someone who prefers it the other way. But if @DavisVaughan is happy/happier, then I am happy.

ateucher:
Maybe at some point we could build some tooling to make republishing a released site a little less cumbersome

PR #1851, opened 2023-06-05
Do not add covr to Suggests in `use_coverage()`

### LINKED ISSUE #1844: Make `use_coverage()` add covr to `Config/Needs/coverage` rather `Suggests`
Currently, `use_coverage()` adds covr to `Suggests`. 

Since covr is only used during package development, I would prefer to include it in `Config/Needs/coverage`. I believe this is consistent with the fact that, e.g., rcmdcheck is not listed in `Suggests` either. Additionally, the [`test-coverage` workflow](https://github.com/r-lib/usethis/blob/e05947f0e49244d7dcb13105a48e32718eb59db5/.github/workflows/test-coverage.yaml) already installs covr, as well as any dependency listed in `Config/Needs/coverage`.

Bisaloo:
## Context

At the moment, `use_coverage()` adds covr to `Suggests`. This was not always the case but it was added in https://github.com/r-lib/usethis/commit/2698067772018000e8c49453956c04fd6a036a12, apparently to facilitate coverage computation in travis CI(?).

usethis and the R community have now largely moved away from travis in favour of GitHub Actions. The [`test-coverage` GHA workflow provided by usethis](https://github.com/r-lib/usethis/blob/e05947f0e49244d7dcb13105a48e32718eb59db5/.github/workflows/test-coverage.yaml) already installs covr in all cases, without needing it to be explicitly listed in `Suggests`.

https://github.com/r-lib/usethis/blob/e05947f0e49244d7dcb13105a48e32718eb59db5/.github/workflows/test-coverage.yaml#L24-L27

## Homogenizing 

This PR proposes to stop automatically including covr in Suggests when running `use_coverage()` (fix #1844). This is in line with what `use_pkgdown()` does: pkgdown is not added to Suggests there. Again, it's not necessary because as in the `test-coverage` workflow, pkgdown is automatically installed without the need to l

### Differences with `use_pkgdown()`

`use_pkgdown()` does however use `check_installed("pkgdown")`. I don't believe it's necessary in the case of `use_coverage()`. The difference is that `use_pkgdown()` needs to know which pkgdown version is installed to determine the structure of the default `_pkgdown.yml` config file.

https://github.com/r-lib/usethis/blob/00ee94781f9f2552f5dc2a6c710b58803e8dba58/R/pkgdown.R#L50-L52

(Note that `check_installed("pkgdown")` could probably be removed here as well though. It was strictly necessary in the past when usethis used pkgdown functions (https://github.com/r-lib/usethis/commit/87a2cb5d7e7ed9234bee4a4be92c70d959e9a2a7) but it's not the case anymore as far as I can tell. Version checking could be done conditionally and the bootstrap 5 template could be used by default when pkgdown is not yet installed. It's a different question though and I can open a separate issue if you wish)


jennybc:
Thanks!

PR #1855, opened 2023-06-06
Use CC licenses, all of them

pachadotdev:


pachadotdev:
I will move this to a PR from a dedicated branch

PR #1864, opened 2023-06-17
Make compute_name() work when `ext` is not "R"

### LINKED ISSUE #1863: Error in `compute_name()` when extension is not "R"
Thanks for the great package!

I believe I've encountered a bug.  When a user does `usethis::use_rcpp("code.cpp")`, one gets the error

```
`name` must have extension "cpp", not "cpp"
```

I believe the problem is that "R" is hardcoded into this line of code:

https://github.com/r-lib/usethis/blob/97b2891a838fe0138543a9a93e8b2b7f5cf86677/R/r.R#L81

Here's a reprex demonstrating the problem:

``` r
usethis:::compute_name("code.cpp", ext="cpp")
#> Error:
#> ! `name` must have extension "cpp", not "cpp".
#> Backtrace:
#>     ▆
#>  1. └─usethis:::compute_name("code.cpp", ext = "cpp")
#>  2.   └─cli::cli_abort(...)
#>  3.     └─rlang::abort(...)
```

<sup>Created on 2023-06-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'm about to make a pull request with a proposed fix and an added test demonstrating the problem.

jacobbien:
Closes #1863

Please see #1863 for a description of the issue that this PR is aiming to address.

### DIFF
--- a/R/r.R
+++ b/R/r.R
@@ -78,7 +78,7 @@ compute_name <- function(name = NULL, ext = "R", error_call = caller_env()) {
 
     if (path_ext(name) == "") {
       name <- path_ext_set(name, ext)
-    } else if (path_ext(name) != "R") {
+    } else if (path_ext(name) != ext) {
       cli::cli_abort(
         "{.arg name} must have extension {.str {ext}}, not {.str {path_ext(name)}}.",
         call = error_call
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks! I tweaked the test.

FYI, I think `name` is mostly envisioned to be extension-free here, but I agree with your change.

Also, I'm editing your PR description to use the Magic Words that will cause the merge to close the associated issue.

https://code-review.tidyverse.org/author/submitting.html#closes

PR #1867, opened 2023-06-21
Update man page for git_vaccinate()

jakobarendt:
I updated the man page for the git_vaccinate() command such that it documents **all** of the adjustments made to the `.gitignore`. I also reordered the list of files and directories that the command adds to the `.gitignore` such that it fits the order in the output returned after calling the command.

Sorry if my line breaks do not follow your standard conventions. I was a little unsure about that.

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -506,7 +506,7 @@ default_branch_sitrep <- function() {
 
 #' Vaccinate your global gitignore file
 #'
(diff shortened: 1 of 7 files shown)


jennybc:
Can you implement this using markdown syntax, i.e. the backticks that were previously there, as opposed to `\code{}`? And also run `devtools::document()` to regenerate the `.Rd` file.

jakobarendt:
Done. Thanks for the feedback!

jennybc:
Thanks!

PR #1879, opened 2023-08-01
In `use_package()` and others, defer check for installed package

krlmlr:
I might want to add a package dependency first, and worry about installing later.


carlajarchuleta1-a11y:
f-check-installed-later

PR #1894, opened 2023-09-06
`use_github()` - allow repo creation when repo is a redirect

### LINKED ISSUE #1893: `use_github()` fails with "Repo already exists" after renaming a repo
If I create a repo on GitHub, then rename it and then try to use `use_github()` with the old name, it errors with `"Repo [reponame] already exists on 'github.com'"`, even though that name should be now be available after the renaming.

To reproduce:

On GitHub:
1. Create a repo (called for example "test123")
2. Rename the repo to "test123.old"

Locally:
1. `create_project("~/Desktop/test123")`
2. `use_git()` (accept all the prompts)
3. `use_github()` - this gives:
    ```
    ℹ Defaulting to 'https' Git protocol
    ✔ Setting active project to '/Users/andy/dev/test123'
    Error: Repo 'ateucher/test123' already exists on 'github.com'
    ```

This occurs because of the [redirect that GitHub does on renamed repos](https://docs.github.com/en/enterprise-cloud@latest/repositories/creating-and-managing-repositories/renaming-a-repository), so the API call in `check_no_github_repo()` [here](https://github.com/r-lib/usethis/blob/467ff57f92afb34878dc411c46e3f0de029ceea4/R/github.R#L273-L277) does not fail because it finds the renamed repo (`test123.old`).

ateucher:
This changes `check_no_github_repo()` to check the name of the response from the GitHub API against the provided spec (user/repo) - If they are different it assumes it's a redirect and then proceeds and allows repo creation. It retains the behaviour where if request to the GitHub API returns a 404 it similarly passes and allows the creation of the new repo.

Unfortunately the response returned from the GitHub API doesn't appear to explicitly say it's been redirected (i.e., it's a 200 HTTP code, not a 301).

Fixes #1893.

I didn't add any tests - though could add to the manual tests if you like?

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -268,6 +268,7 @@ check_no_origin <- function() {
 }
 
 check_no_github_repo <- function(owner, repo, host) {
+  spec <- glue("{owner}/{repo}")
   repo_found <- tryCatch(
     {
       repo_info <- gh::gh(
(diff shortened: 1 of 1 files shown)


jennybc:
Is this time sensitive, i.e. needed for conf?

ateucher:
Nope! It was for my own prep, but I just used my local branch.

jennybc:
Thanks!

PR #1898, opened 2023-10-04
Add `use_package_doc()` details for package and function by the same name

### LINKED ISSUE #1170: use_package_doc() leads to an R-CMD-CHECK warning if package shares name with a function
`use_package_doc()` leads to an R-CMD-CHECK warning if the package contains an exported function with the same name as the package itself.

From my package `ipa`, which exports the function `ipa()`:

```
checking Rd metadata ... WARNING
##[warning]Rd files with duplicated alias 'ipa':
  ‘ipa-package.Rd’ ‘ipa.Rd’
```

Is it possible to make package documentation still accessible from `package?foo` and ``?`foo-package` `` even if `?foo` is already used by function documentation?

SimonDedman:
resolves #1170

### DIFF
--- a/R/documentation.R
+++ b/R/documentation.R
@@ -1,28 +1,30 @@
 #' Package-level documentation
 #'
-#' Adds a dummy `.R` file that will prompt roxygen to generate basic
+#' Adds a dummy `.R` file that will cause roxygen2 to generate basic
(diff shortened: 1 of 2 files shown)


jennybc:
Thanks!

PR #1899, opened 2023-10-07
Tweak blog post action order

hadley:
I assume the intent here is to get the blog post into the newly built site, but I found the order action a bit confusing, especially given that I like to be absolutely sure I'm getting the right url by copying it from the live site.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -137,11 +137,11 @@ release_checklist <- function(version, on_cran) {
     "Wait for CRAN...",
     "",
     todo("Accepted :tada:"),
-    todo("Add preemptive link to blog post in pkgdown news menu", type != "patch"),
+    todo("Finish & publish blog post", type != "patch"),
+    todo("Add link to blog post in pkgdown news menu", type != "patch"),
     todo("`usethis::use_github_release()`"),
     todo("`usethis::use_dev_version(push = TRUE)`"),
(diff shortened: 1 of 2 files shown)


jennybc:
As previously foretold, here we go:

https://github.com/r-lib/usethis/pull/1847#pullrequestreview-1451829213

I'm going to ask @DavisVaughan to review.

DavisVaughan (approved):
I don't think this counters anything from the previous PR, so it looks good

I mentioned a chicken and egg problem here:
https://github.com/r-lib/usethis/issues/1761#issuecomment-1419564358

But I think this workflow gets us around it nicely, for example:

- Release blog post for 0.2.0 (blog post typically links back to pkgdown site, which will be at 0.1.0)
- Add link to blog post for 0.2.0 in pkgdown news menu
- Create a 0.2.0 commit that builds pkgdown site for 0.2.0 (this includes the blog post news menu item, and now the link in the blog post points to a 0.2.0 site too)
- Market the blog post

The one thing you won't be able to easily do when publishing the blog post is link _directly_ to the 0.2.0 NEWS section, i.e. like https://slider.r-lib.org/news/index.html#slider-030, since the 0.2.0 site won't be up when you publish the blog post. But that is ok, we can just link to the NEWS page itself instead, i.e. without the direct header link https://slider.r-lib.org/news/index.html

PR #1375, opened 2021-02-24
Drop test environments from cran-comments

hadley:
Fixes #1365

PR #1901, opened 2023-10-18
remove unused parameter in use_cran_comments()

AshesITR:
The template was changed two years ago to no longer use `rversion`, so no need to provide it.


jennybc:
Note: the template change was in #1375.

Thanks!

Issue #1857, opened 2023-06-08
`cran_version()` should be more robust


ateucher:
Several functions (`use_release_checklist()`, `use_news_md()`) call `cran_version()` to check if a package is on CRAN, which fails if a CRAN mirror is not set.

`cran_version()` calls `available.packages()` which in turn calls `contrib.url()` with `repos = getOption("repos")`. If the CRAN mirror is unset (the default value is `"@CRAN@"` in a factory fresh installation) it will prompt a user to choose in an interactive session, or fail otherwise:

``` r
withr::with_options(
  list(repos = c(CRAN = "@CRAN@")), 
  usethis:::cran_version()
)
#> Error in contrib.url(repos, type): trying to use CRAN without setting a mirror
```

This is showing up in revdep checks and CRAN checks for some usethis reverse depenencies which use `use_news_md()`. E.g., https://github.com/cynkra/fledge/issues/683

`cran_version()` should not fail if there is no internet, or if there is no CRAN mirror set. In the former case, it should likely return `NULL`, which it currently returns if the package is not on CRAN. That seems like a reasonable fallback. If there is no CRAN mirror set we should try to set one locally, and also wrap that in a `tryCatch()`, returning `NULL` again if still unsuccessful.

This was surfaced in revdepchecks for #1849, was at first thought to be a false positive, and did not cause issues on incoming checks, but is causing failures for revdeps on CRAN.

ateucher:
If `getOption(repos)["CRAN"]` is `NULL` or `NA` (I don't know under what circumstances it would be set this way), `cran_version()` currently returns `NULL`. By locally setting a CRAN mirror we should be able to have success in more of these cases as well.

krlmlr:
Are you aiming for a point release anytime soon? Thanks!

jennybc:
I don't have concrete plans because this is literally the only thing that would be in it. Since we just released 1 year's worth of maintenance, I suspect a few more smallish things will come up in the medium term. So maybe once there's a couple little things, we'll do a patch release.

jennybc:
I ended up needing to do a patch release anyway, so this fix will presumably be in usethis 2.2.1.

PR #1908, opened 2023-10-31
Speed up tests

hadley:
* Enable parallel testing. For me, this dropped runn time down from 30s to 11s, 10s of which is the release tests.
* Refactor release tests to avoid expensive CRAN downloads. This has equivalent test coverage but reduces test time from 10s to 4s, bringing total test time down to around 8 seconds.
* Start the slowest tests (github-actions and release) first. This brings total test time down to around 6s


R/release.R
```diff
@@ -412,7 +412,7 @@ cran_version <- function(package = project_name(), available = NULL) {
   if (is.null(available)) {
     # Guard against CRAN mirror being unset
     available <- tryCatch(
-      utils::available.packages(repos = default_cran_mirror()),
+      available.packages(repos = default_cran_mirror()),
```
jennybc, line 415: Just mildly curious. Does this make a difference?

Updated: oh I see you want to mock it.

jennybc (approved):
The new form of testing the problem from #1857 is pretty different, but if you're happy that it still gets the job done, then I'm happy!

hadley:
Abstractly I think the turns the tests around 90°; they still cover the same stuff but in a rather different way.

PR #1909, opened 2023-10-31
More upkeep polishing

hadley:


hadley:
@olivroy this is specifically for our internal upkeep process, where I don't think unneeded dependencies tend to be much of a problem.

R/release.R
```diff
@@ -112,8 +112,6 @@ release_checklist <- function(version, on_cran) {
       Check if any deprecation processes should be advanced, as described in \\
       [Gradual deprecation](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation)",
       type != "patch" && has_lifecycle),
-    todo("Bump required R version in DESCRIPTION to {tidy_min_r_version}",
```
hadley, line 115: Now in upkeep, rather than release.

tests/testthat/test-upkeep.R
```diff
@@ -1,56 +1,49 @@
...
-      License = "MIT + file LICENSE"
-    )
-  )
   create_local_package()
+  use_mit_license()
 
-  expect_snapshot(
-    writeLines(tidy_upkeep_checklist(posit_pkg = TRUE, posit_person_ok = FALSE)),
-    transform = scrub_checklist_footer
+  local_mocked_bindings(
```
hadley, line 5: I probably should've done this in a separate PR, but using mocking for everything makes the tests quite a bit simpler.

R/release.R
jennybc: I kinda like it as part of the release checklist, but not a hill I will die on.

jennybc (approved):

R/release.R
hadley: I don't have strong feelings either except that it should be in one or the other.

PR #1919, opened 2023-11-10
 Use CC licenses, all of them (follow up to #1855)

pachadotdev:


hadley:
We very much appreciate your time and enthusiasm, but as I mentioned in the related issue, we'd prefer to provide key CC licenses not all of them at once.

PR #1921, opened 2023-11-10
cpp11 dynlib

pachadotdev:
Hi @hadley

These minimal changes allow the user to just write functions of the form

```
#include <cpp11.hpp>

using namespace cpp11;

[[cpp11::register]] int hello(int x) {
    return x + 1;
}
```

Immediately after using `use_cpp11()`.

With this change the user can run `devtools::load_all()` and `hello(1)` will work. Otherwise it says `pkgname_hello` not found.

Semi-related to https://github.com/r-lib/cpp11/pull/341


hadley:
I think it would be better to follow a similar process to `use_import_from()`, i.e. use `roxygen_ns_append()` to write to the package doc, and then run `roxygen_update_ns()` to automatically update the namespace and reload the package.

pachadotdev:
> I think it would be better to follow a similar process to `use_import_from()`, i.e. use `roxygen_ns_append()` to write to the package doc, and then run `roxygen_update_ns()` to automatically update the namespace and reload the package.

ready!
good suggestion, I just tried with `usethis::create_package()` and it works

R/cpp11.R
```diff
@@ -40,3 +42,16 @@ check_cpp_register_deps <- function() {
...
+
+get_cpp_dynlib <- function() {
+  desc <- desc::desc(package = NULL)
+  desc$get_list("Package")[[1]]
+}
+
+check_cpp_dynlib <- function() {
+  pkgname <- get_cpp_dynlib()
+  roxygen_ns_append(
+    sprintf("@useDynLib %s, .registration = TRUE", pkgname)
```
hadley, line 54: You could use `project_name()` here instead of a custom function.

And then I think it's short enough that it would be clearer to inline directly in `use_cpp11()`.

hadley (commented):
Can you please also add a news bullet and a test? (Probably something in "use_cpp11() creates files/dirs, edits DESCRIPTION and .gitignore")

.vscode/settings.json
```diff
@@ -0,0 +1,59 @@
+{
```
hadley: Can you please ignore this file?

NEWS.md
```diff
@@ -5,6 +5,10 @@
 * Implicit usage of `numeric_version()` via comparison now always provides
   character input. This is in response to a request from CRAN to anticipate
   future solutions to <https://bugs.r-project.org/show_bug.cgi?id=18548>.
+* `use_cpp11()` creates/edits DESCRIPTION and NAMESPACE, allowing to use C++
+  code directly.
+* `.gitignore` now ignores `.vscode` for those who contribute to `usethis` from
```
hadley, line 10: This doesn't need a news bullet.

pachadotdev: removed

pachadotdev:
@hadley it should be ok now

hadley:
Thanks!

jennybc:
I have reverted this because I need to release usethis under a CRAN deadline and CI has been failing ever since this PR was merged. This change creates a test-time dependency on cpp11 and decor, neither of which usethis has a direct or indirect dependency on. I've fiddled around with the failing tests a bit, but there's no easy way, e.g. via a simple mock, to patch this up quickly, because the dependencies are needed for the `roxygenize()` step which is precisely the point. So @pachadotdev or @hadley if you care a lot about this, we need to either give up on testing the feature or invest time in more refactoring and mocking.

```
── Error ('test-cpp11.R:19:3'): use_cpp11() creates files/dirs, edits DESCRIPTION and .gitignore ──
<packageNotFoundError/error/condition>
Error in `loadNamespace(x)`: there is no package called 'cpp11'
Backtrace:
     ▆
  1. ├─usethis::use_cpp11() at test-cpp11.R:19:3
  2. │ └─usethis:::roxygen_update_ns() at usethis/R/cpp11.R:31:3
  3. │   ├─utils::capture.output(...) at usethis/R/roxygen.R:105:3
  4. │   │ └─base::withVisible(...elt(i))
  5. │   ├─base::suppressMessages(roxygen2::roxygenise(proj_get(), "namespace")) at usethis/R/roxygen.R:105:3
  6. │   │ └─base::withCallingHandlers(...)
  7. │   └─roxygen2::roxygenise(proj_get(), "namespace") at usethis/R/roxygen.R:105:3
  8. │     └─roxygen2 (local) load_code(base_path)
  9. │       └─pkgload::load_all(path, helpers = FALSE, attach_testthat = FALSE)
 10. │         └─pkgbuild::compile_dll(path, quiet = quiet)
 11. │           └─pkgbuild:::update_registration(...)
 12. └─base::loadNamespace(x)
 13.   └─base::withRestarts(stop(cond), retry_loadNamespace = function() NULL)
 14.     └─base (local) withOneRestart(expr, restarts[[1L]])
 15.       └─base (local) doWithOneRestart(return(expr), restart)
```

jennybc:
I suspect this should just be `roxygen_ns_append(...) && roxygen_remind()`, as done elsewhere in the package. Initiating use of cpp11 is a once-per-package event, so it's OK for it to involve a couple of steps.

PR #1925, opened 2023-11-29
Replace references to Lastpass with Bitwarden

rogerssam:
There were a couple of left over references to LastPass after PR #1725 which I have replaced with Bitwarden. Happy to just remove references to LP if you prefer.


jennybc:
Thanks! (Looks the like CI failures were introduced in #1921, not here.)

PR #1928, opened 2023-12-06
Use cli in messages

olivroy:
fix #956 

This is a rough draft. I would understand you'd rather take this on yourself, but in case it could be of any use, I thought I'd share!

Edit: CI won't cooperate. but anyway. I will try this branch of usethis installed locally and will update if something goes wrong.

Edit2 : feel free to take anything that you like from this!

## List of problems

* Can't figure out how to [have red bullets](https://cli.r-lib.org/articles/usethis-ui.html#how-to).

## Nice to have?

ui_code that is meant to be run and other that is not meant?

## Results

Impact of changes in be8a0f0

image

## not covered by tests

* `git_sitrep()`: may be a problem as it uses `ui_todo with multiple lines input extensively, and is not really tested)
* `pr_push()`

PR #1935, opened 2024-01-17
Minor test lints

olivroy:
This is just to remove unrelated stuff from #1928


jennybc:
Thanks! The CI failures are because P3M is having a bad day.

PR #1940, opened 2024-01-29
Indicate argument type

MikeJohnPage:
On several occasions now I have tried to add a minimum package version using a `double` like so:

```r
usethis::use_package("my_package", min_version = 1.1.0)
```

Each time I get an error, as I incorrectly guessed that `min_version` would require a numeric type. While I think the error message is helpful enough (`Error: unexpected numeric contstant in "usethis...`), I think the documentation could be improved.

Currently, it is only clear that `min_version` should be a character vector when looking at the examples. The parameter documentation is not clear in and of itself.


jennybc:
Thanks!

PR #1945, opened 2024-02-14
follow @jennybc advice

pachadotdev:
this is a new version of https://github.com/r-lib/usethis/pull/1921

@jennybc  I took some time to type with 1 hand and I used the logic you proposed here https://github.com/r-lib/usethis/pull/1921#issuecomment-1941927554


R/cpp11.R
```diff
@@ -9,9 +9,9 @@
 #'
 #' @export
 use_cpp11 <- function() {
-  check_installed("cpp11")
```
jennybc, line 12: I think we want to keep this check.

R/cpp11.R
```diff
@@ -24,6 +24,10 @@ use_cpp11 <- function() {
 
   check_cpp_register_deps()
 
+  roxygen_ns_append(
+    sprintf("@useDynLib %s, .registration = TRUE", project_name())
+  ) && roxygen_remind()
+
```
jennybc, line 30: It appears that `use_src()` does exactly this already? Can you help me understand why `use_cpp11()` needs to do that again?

https://github.com/r-lib/usethis/blob/93f4410ab93cb6e19cd6a49cd4ce33f93055b38b/R/rcpp.R#L74-L75

tests/testthat/test-cpp11.R
```diff
@@ -5,12 +5,14 @@ test_that("use_cpp11() requires a package", {
...
 test_that("use_cpp11() creates files/dirs, edits DESCRIPTION and .gitignore", {
+  skip_if_not_installed("cpp11")
+
   create_local_package()
   use_roxygen_md()
+  use_package_doc()
 
   local_interactive(FALSE)
   local_check_installed()
-  local_mocked_bindings(check_cpp_register_deps = function() invisible())
```
jennybc, line 16: In terms of skips and mocks, I would expect this to look very much like the same test for `use_rcpp()`:

https://github.com/r-lib/usethis/blob/a91943706708f7ab41105cc3d5edf6ee61bbfb15/tests/testthat/test-rcpp.R#L6-L21

pachadotdev: I changed it to reflect this. The test "use_rcpp_armadillo() creates Makevars..." is something I can implement later for cpp11armadillo.

jennybc:
Whew, we got there!

PR #1954, opened 2024-02-28
Add option for `proj_activate()` in same session

olivroy:
I use a wrapper that allows me to switch projects only by typing the name (as opposed to giving the full path)

https://olivroy.github.io/reuseme/reference/proj_switch.html

Looks at given options in Rprofile to activate project.

```r
# in r profile
options(reuseme.reposdir = c("~/Documents/github", "~/Documents/packages", "~/Desktop/rrr"))
```

let's say you want to use project in "~/Documents/github/usethis"

```r
proj_switch("usethis")
```

works and gives helpful error if there is a typo. (uses `rlang::arg_match()`.

```r
proj_switch("usethiss")
#> Did you mean "usethis"?
```

So the goal of this PR would be to add the `new_session` to `usethis::proj_activate()` to eventually add it to `reuseme::proj_switch()`.

Cheers.


olivroy:
I will just use the rstudioapi package directly

PR #1956, opened 2024-02-29
Use cli

jennybc:
This is a huge PR, most of which is mind-numbingly repetitive. Recommended way to review:

* Read the new developer-facing article `vignettes/articles/ui-cli-conversion.Rmd`, which is glorified notes-to-self. I may create a blog post about this and these notes would be helpful. This will also be helpful to link to in future PRs that introduce unconventional UI. It is probably worth doing `pr_fetch(1956)` and rendering this because then you can enjoy the asciicast chunks that really show the before and after.
* Look for the comments I sprinkled around, where I'll draw attention to a few interesting things.
* `utils-ui.R` is the most important file, because that's where the new UI functions live.

I plan to merge this soon, but am happy to receive feedback. But I really want to get friends and family using this soon so we can put a few miles on it be fore I submit to CRAN.

Closes #956, closes #1809

PR #1959, opened 2024-03-06
Cli tweaks

olivroy:
Follow-up to #1956 with mostly minor tweaks, but 3 more notable changes

* `use_logo()` now adds a link to README if it exists.a5ca3b88d9729fb10d4a06cd8bc4cc9f78b430fd
* `check_is_package()` adds a link to help is `whos_asking` is provided. 269d2a2763302adcccf4e773ba6fba66266d1d0e
* Clickable hyperlink to Rprofile

### DIFF
--- a/R/pr.R
+++ b/R/pr.R
@@ -224,7 +224,7 @@ pr_init <- function(branch) {
   config_key <- glue("branch.{branch}.created-by")
   gert::git_config_set(config_key, value = "usethis::pr_init", repo = repo)
 
-  ui_bullets(c("_" = "Use {.run usethis::pr_push} to create a PR."))
+  ui_bullets(c("_" = "Use {.fun usethis::pr_push} to create a PR."))
   invisible()
 }
 
--- a/R/pr.R
+++ b/R/pr.R
@@ -251,10 +251,9 @@ pr_resume <- function(branch = NULL) {
   check_string(branch)
 
   if (!gert::git_branch_exists(branch, local = TRUE, repo = repo)) {
-    code <- glue('pr_init("{branch}")')
     ui_abort(c(
       "x" = "No branch named {.val {branch}} exists.",
-      "_" = "Call {.code {code}} to create a new PR branch."
+      "_" = "Call {.run usethis::pr_init('{branch}')} to create a new PR branch."
     ))
   }
 
@@ -264,7 +263,7 @@ pr_resume <- function(branch = NULL) {
   gert::git_branch_checkout(branch, repo = repo)
   git_pull()
 
-  ui_bullets(c("_" = "Use {.fun usethis::pr_push} to create or update PR."))
+  ui_bullets(c("_" = "Use {.run usethis::pr_push()} to create or update PR."))
   invisible()
 }
 
@@ -302,7 +301,7 @@ pr_fetch <- function(number = NULL, target = c("source", "primary")) {
 
   if (is.na(pr$pr_repo_owner)) {
     ui_abort("
-      The repo or branch where PR #{pr$pr_number} originates seems to have been
+      The repo or branch where {.href [PR #{pr$pr_number}]({pr$pr_https_url})} originates seems to have been
       deleted.")
   }
 
@@ -1001,8 +1000,9 @@ check_pr_branch <- function(default_branch = git_default_branch()) {
       "i" = "The {.code pr_*()} functions facilitate pull requests.",
       "i" = "The current branch ({.val {gb}}) is this repo's default branch, but
              pull requests should NOT come from the default branch.",
-      "i" = "Do you need to call {.fun pr_init} (new PR)? Or {.fun pr_resume} or
-             {.fun pr_fetch} (existing PR)?"
+      "i" = "Do you need to call {.fun usethis::pr_init} (new PR)?
+             Or {.fun usethis::pr_resume} or
+             {.fun usethis::pr_fetch} (existing PR)?"
     )
   )
 }
--- a/R/proj.R
+++ b/R/proj.R
@@ -214,8 +214,10 @@ check_is_package <- function(whos_asking = NULL) {
 
   message <- "Project {.val {project_name()}} is not an R package."
   if (!is.null(whos_asking)) {
+    # remove brackets, so that it is recognized as a help link
+    whos_asking_fn <- sub("\\(\\)", "", whos_asking)
     message <- c(
-      "i" = "{.code {whos_asking}} is designed to work with packages.",
+      "i" = "{.topic [{whos_asking}](usethis::{whos_asking_fn})} is designed to work with packages.",
       "x" = message
     )
   }
--- a/R/rprofile.R
+++ b/R/rprofile.R
@@ -40,7 +40,7 @@ use_devtools <- function() {
 use_rprofile_package <- function(package) {
   check_installed(package)
   ui_bullets(c(
-    "_" = "Include this code in {.path .Rprofile} to make {.pkg {package}}
+    "_" = "Include this code in {.href [.Rprofile](file://~/.Rprofile)} to make {.pkg {package}}
            available in all interactive sessions:"
   ))
   ui_code_snippet("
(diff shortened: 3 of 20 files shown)


R/pr.R
```diff
@@ -224,7 +224,7 @@ pr_init <- function(branch) {
   config_key <- glue("branch.{branch}.created-by")
   gert::git_config_set(config_key, value = "usethis::pr_init", repo = repo)
 
-  ui_bullets(c("_" = "Use {.run usethis::pr_push} to create a PR."))
+  ui_bullets(c("_" = "Use {.fun usethis::pr_push} to create a PR."))
```
jennybc, line 227: Can you remove this change? I just made this change yesterday and it was in response to using the `pr_*()` functions and wishing this had been runnable.

R/pr.R
```diff
@@ -251,10 +251,9 @@ pr_resume <- function(branch = NULL) {
   check_string(branch)
 
   if (!gert::git_branch_exists(branch, local = TRUE, repo = repo)) {
-    code <- glue('pr_init("{branch}")')
     ui_abort(c(
       "x" = "No branch named {.val {branch}} exists.",
-      "_" = "Call {.code {code}} to create a new PR branch."
+      "_" = "Call {.run usethis::pr_init('{branch}')} to create a new PR branch."
```
jennybc, line 256: What's the reasoning for fiddling with this?

R/proj.R
```diff
@@ -214,8 +214,10 @@ check_is_package <- function(whos_asking = NULL) {
 
   message <- "Project {.val {project_name()}} is not an R package."
   if (!is.null(whos_asking)) {
+    # remove brackets, so that it is recognized as a help link
```
jennybc, line 217: You can remove the comment (or you need to say "parentheses", but better to just remove it).

R/proj.R
```diff
@@ -214,8 +214,10 @@ check_is_package <- function(whos_asking = NULL) {
...
+    whos_asking_fn <- sub("\\(\\)", "", whos_asking)
```
jennybc, line 218:
```suggestion
    whos_asking_fn <- sub("()", "", whos_asking, fixed = TRUE)
```

I avoid backslashes in regexes whenever I can.

https://xkcd.com/1638

R/rprofile.R
```diff
@@ -40,7 +40,7 @@ use_devtools <- function() {
 use_rprofile_package <- function(package) {
   check_installed(package)
   ui_bullets(c(
-    "_" = "Include this code in {.path .Rprofile} to make {.pkg {package}}
+    "_" = "Include this code in {.href [.Rprofile](file://~/.Rprofile)} to make {.pkg {package}}
```
jennybc, line 43: This hard-wired `.Rprofile` path makes me nervous and I think it's inconsistent with what we do elsewhere in usethis re: this path.

R/rprofile.R
```diff
@@ -54,7 +54,7 @@ use_rprofile_package <- function(package) {
 #' @export
 use_partial_warnings <- function() {
   ui_bullets(c(
-    "_" = "Include this code in {.path .Rprofile} to warn on partial matches:"
+    "_" = "Include this code in {.href [.Rprofile](file://~/.Rprofile)} to warn on partial matches:"
```
jennybc, line 57: Ditto

R/spelling.R
```diff
@@ -24,5 +24,5 @@ use_spell_check <- function(vignettes = TRUE,
   spelling::spell_check_setup(
     pkg = proj_get(), vignettes = vignettes, lang = lang, error = error
   )
-  ui_bullets(c("_" = "Run {.run devtools::check()} to trigger spell check."))
+  ui_bullets(c("_" = "Run {.run devtools::spell_check()} to trigger spell check."))
```
jennybc, line 27: I just allow spell checking to happen during `check()`.

R/test.R
```diff
@@ -28,11 +28,7 @@ use_testthat <- function(edition = NULL, parallel = FALSE) {
 }
 
 use_testthat_impl <- function(edition = NULL, parallel = FALSE) {
-  check_installed("testthat")
-  if (utils::packageVersion("testthat") < "2.1.0") {
-    ui_abort("
-      {.pkg testthat} 2.1.0 or greater needed. Please install before re-trying")
-  }
+  check_installed("testthat", version = "2.1.0")
```
jennybc, line 31: Can you remove this change?

R/tidyverse.R
```diff
@@ -86,14 +86,15 @@ create_tidy_package <- function(path, copyright_holder = NULL) {
...
   ui_bullets(c(
-    "i" = "In the new package, remember to do:",
-    "_" = "{.run usethis::use_git()}",
-    "_" = "{.run usethis::use_github()}",
-    "_" = "{.run usethis::use_tidy_github()}",
-    "_" = "{.run usethis::use_tidy_github_actions()}",
-    "_" = "{.run usethis::use_tidy_github_labels()}",
-    "_" = "{.run usethis::use_pkgdown_github_pages()}"
-  ))
+    "i" = "In the new package, remember to do:"))
```
jennybc, line 89: `))` should be on their own line

tests/testthat/_snaps/github-actions.md
```diff
@@ -36,5 +36,5 @@
     Condition
       Error in `check_uses_github_actions()`:
       x Cannot detect that package {TESTPKG} already uses GitHub Actions.
-      i Do you need to run `use_github_action()`?
+      i Do you need to run `(use_github_action())[usethis::use_github_action()]`?
```
jennybc, line 39: Is this really how this should appear in a snapshot?

tests/testthat/test-proj.R
```diff
@@ -36,7 +36,12 @@ test_that("check_is_package() errors for non-package", {
 
 test_that("check_is_package() can reveal who's asking", {
   create_local_project()
-  expect_usethis_error(check_is_package("foo"), "foo")
+  expect_snapshot(
+    error = TRUE, {
+      check_is_package("foo()")
+    },
```
jennybc, line 42: I don't think these brackets are necessary?

tests/testthat/test-logo.R
```diff
@@ -1,8 +1,19 @@
...
+
+test_that("use_logo() shows a clickable path with README", {
+  skip_if_not_installed("magick")
+  skip_on_os("solaris")
+
+  create_local_package()
+  use_readme_md()
+  img <- magick::image_write(magick::image_read("logo:"), "logo.png")
+  withr::local_options(usethis.quiet = FALSE)
+  expect_snapshot(use_logo("logo.png"),  transform = scrub_testpkg)
```
jennybc, line 18: Is there a way to actually test for "clickable path"? Because it doesn't feel perceivable from the current snapshot.

jennybc (commented):
Thanks for (most of) this 😅 

It would make things easier for me to review and merge if the PR was less of a grab bag. Because some of it's no-brainer stuff I'm definitely looking for. But other changes are more opinions and seem to be where we have different opinions.

tests/testthat/_snaps/github-actions.md
olivroy: I made a mistake in the order I did (text)[url] instead of  [text](url).  noob.

tests/testthat/test-logo.R
olivroy: Well the single quotes in the snapshot indicates that some markup was applied (and that it could actually find README.md) 'README.md'.

R/rprofile.R
olivroy: The only problem with the current approach is that it either creates a clickable output. However, clicking on it will do nothing..

 (that feels confusing to me.) So if you don't think this is safe enough, maybe using {.val} would be better?

R/pr.R
olivroy: To create a run link. I tried 2a932519c929d8201c451fce44e584269ba22c25, which seems less invasive?

R/rprofile.R
jennybc: I'm worried about the path to `.Rprofile`. It's not necessarily/always `~/.Rprofile`. One has to honor the `R_PROFILE_USER` env var, for example. And you have to worry about whether the whole stack has the same idea of "home directory" on Windows.

https://github.com/r-lib/usethis/blob/682f4bd92879367bebc1cab9ac2c8077733048a4/R/edit.R#L113

olivroy: I have decided to revert as you responded :)

jennybc:
Thanks!

Issue #1961, opened 2024-03-08
`use_zip` error in unzipping directory with a different name from the zipfile


burnsal:
I am running into a missing value error trying to download and unzip large files with the `use_zip` function. 

I was able to isolate the issue to an incompatibility with the `top_directory`  function for extracting the head directory of the zip files in the helper function `tidy_unzip`. Here is the reprex with output included:

```r
## usethis::use_zip() issue reprex

# Goal: download and unzip large files stored on a Figshare platform

# long link: https://agdatacommons.nal.usda.gov/ndownloader/files/44576230
# bitly link: https://bit.ly/ctdv2sp

# attempt with existing function
> usethis::use_zip("https://bit.ly/ctdv2sp",
                 destdir = "C:/temp")
✔ Downloading from 'https://bit.ly/ctdv2sp'
Downloaded: 312.74 MB  (100%)
✔ Download stored in 'C:/temp/species_v2_3.zip'
Error in if (length(unique_top) > 1 || !is_directory) { : 
  missing value where TRUE/FALSE needed

# separate process into its helper functions to source the error
> usethis:::tidy_download("https://bit.ly/ctdv2sp",
                        destdir = "C:/temp")
Downloaded: 312.74 MB  (100%)

> usethis:::tidy_unzip("C:/temp/species_v2_3.zip")
Error in if (length(unique_top) > 1 || !is_directory) { : 
  missing value where TRUE/FALSE needed


## Dig deeper into `tidy_unzip`

# ABnote: first, define zipfile downloaded with `tidy_download`
> zipfile <- "C:/temp/species_v2_3.zip"
> file.exists(zipfile)
[1] TRUE

> base_path <- path_dir(zipfile)
> print(base_path)
[1] "C:/temp"

> filenames <- utils::unzip(zipfile, list = TRUE)[["Name"]]
> print(filenames) # you can see that the zipfile has a single-folder directory with a different name than the zip file
[1] "species_v2/label_encoder.txt" "species_v2/model_arch.pt"     "species_v2/model_weights.pth"

# dropbox particularites do not apply here, skip next two lines

> td <- top_directory(filenames)
Error in if (length(unique_top) > 1 || !is_directory) { : 
  missing value where TRUE/FALSE needed

## AH HA! Look at `top_directory` function
> in_top <- path_dir(filenames) == "."
[1] FALSE FALSE FALSE
> unique_top <- unique(filenames[in_top])
character(0)
[...]

PR #1962, opened 2024-03-08
Fix Issue #1961: tidy unzip directory

burnsal:
Fix for issue #1961 to allow use_zip to successfully unzip file with different directory structure


jennybc:
Can you create a small `.zip` file that displays the behaviour you're targetting? And add it as a test fixture, similar to `yo-loose-regular.zip` and the like, all explained here https://github.com/r-lib/usethis/tree/main/tests/testthat/ref.

Then you can add a new test for your scenario, along these lines:

https://github.com/r-lib/usethis/blob/fdb4a7244df3a229dbf2b9af1027e8b962b8a59a/tests/testthat/test-course.R#L86-L107

You can put that in a new test, i.e. a new `test_that()` block inside `test-course.R`.

R/course.R
```diff
@@ -529,10 +529,14 @@ top_directory <- function(filenames) {
   in_top <- path_dir(filenames) == "."
   unique_top <- unique(filenames[in_top])
   is_directory <- grepl("/$", unique_top)
-  if (length(unique_top) > 1 || !is_directory) {
-    NA_character_
+  if(length(unique(path_dir(filenames)))==1 & length(unique_top) == 0){
```
jennybc, line 532:
```suggestion
  if (length(unique(path_dir(filenames))) == 1 && length(unique_top) == 0) {
```

burnsal:
Test fixture added. I was not able to find the `test_file` function, but the test works for me if I change that command to point to the test file in the testthat/ref folder, the test script runs smoothly.

burnsal:
Resolved typos on previous request and updated test to be compatible with current main branch

PR #1963, opened 2024-03-11
Error chaining tweaks

olivroy:
Just a little proof of concept. Addresses https://github.com/r-lib/usethis/pull/1956#discussion_r1508347797

With the usage of cli, the error messages have become a little bit more verbose.

However, I believe that usethis error messages are usually very clear, and the error trace is rarely needed.

This PR adds a default of [`call = NULL`](https://cli.r-lib.org/reference/cli_abort.html#arguments) for usethis errors.

I overrode this default where I felt like it would be useful.

Also used `call = caller_env()` where it felt appropriate.

So for example, before if you called `use_package_doc()` on a non-package project. You'd get Error in `check_is_package()`

```r
# before with project
use_package_doc()
#> Error in `check_is_package()`
#> Error message...

# with this PR
use_package_doc()
#> Error in `use_package_doc()`
#> Error message ...
```

I still consider this a bit incomplete, but thought I'd share in case you had some work started on this.

I haven't tested it that much since I have other usethis branches installed, but from my observations in the past months, I still think that some calls don't help understanding the error, so this route is probably worth pursuing


olivroy:
not worth it

PR #1993, opened 2024-05-18
Add PR link to pr_finish

olivroy:
Sometimes, if you reopen a project later, you may want to be able to jump back to the PR URL

### DIFF
--- a/R/pr.R
+++ b/R/pr.R
@@ -967,7 +967,7 @@ pr_branch_delete <- function(pr) {
 
   if (is.null(pr_ref)) {
     ui_bullets(c(
-      "i" = "{.href [PR #{pr$pr_number}]({pr$pr_html_url})} originated from branch {.val {pr_remref}},
+      "i" = "PR {.href [{pr$pr_string}]({pr$pr_html_url})} originated from branch {.val {pr_remref}},
              which no longer exists."
     ))
     return(invisible(FALSE))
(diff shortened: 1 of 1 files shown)


olivroy:
I had never tried `pr_finish()` on an unmerged PR, not convinced deleting the local branch is actually desirable.

Instead, maybe it could nudge towards:

1. `pr_pause()` (to switch back to main)
2. Close the pr on github + `pr_finish()` / `pr_forget()`
3. `pr_forget()` to cancel the PR? (though `pr_forget()` possibly does this same thing

I am also finding the git challenge message too wordy. Maybe this could be revisited:

Currently:

```r
> pr_pause()
✔ Setting active project to "~/Documents/rrr-forks/usethis".
✔ Switching back to the default branch.
! There are uncommitted changes, which may cause problems or be lost when we push,
  pull, switch, or compare branches
  Do you want to proceed anyway?

1: Negative
2: Yup
3: No
```

What I propose:



```r
> pr_pause()
✔ Setting active project to "~/Documents/rrr-forks/usethis".
✔ Switching back to the default branch.
! Uncommitted changes may cause problems or be lost when switching branch.

I want to:
1: proceed with branch switching and possibly lose uncommited changes to files.
2: abort and inspect {.file {uncommited_files}} manually and commit or ignore changes as needed.
```

Possibly a resource with the risks would be great. I can't quite figure out when this will actually create something I don't want, or if all is fine.

PR #1996, opened 2024-05-23
Nudge towards building favicons in `use_logo()` and `use_pkgdown()`

### LINKED ISSUE #1995: Should `use_logo()` call or suggest  `pkgdown::build_favicons()`?

olivroy:
fixes #1995.

I think usethis could run it automatically, but since `build_favicons()` has an openssl dependency, it may better to let users do it.

Over time, I have encountered many examples of packages that don't use the standard man/figures/logo.png that pkgdown recognizes. 

With https://github.com/r-lib/pkgdown/pull/2563, they will no longer be built automatically on gh actions, so they have to be initialized manually by users.

cc: @hadley if you have some views of what should happen here!

### DIFF
--- a/tests/testthat/test-pkgdown.R
+++ b/tests/testthat/test-pkgdown.R
@@ -93,6 +93,7 @@ test_that("use_pkgdown() nudges towards use_logo() if the package seems to have 
 })
 
 test_that("use_pkgdown() nudges towards build_favicons().", {
+  skip_if_not_installed("magick")
   skip_on_os("solaris")
 
   create_local_package()


R/pkgdown.R
```diff
@@ -154,6 +168,21 @@ pkgdown_config_path <- function() {
   )
 }
 
+has_logo <- function() {
+  !is.null(
+    path_first_existing(c("man/figures/logo.png", "man/figures/logo.svg"))
+  )
+}
+
+potentially_has_logo <- function() {
```
olivroy, line 177: This idea came to me, because of PRs I opened in various repos of package devs who didn't seem to be aware of this. see #1995

tests/testthat/test-pkgdown.R
```diff
@@ -77,6 +77,35 @@ test_that("pkgdown_url() returns correct data, warns if pedantic", {
...
+  local_mocked_bindings(pkgdown_version = function() "1.9000")
+
+  img <- magick::image_write(magick::image_read("logo:"), "hex-sticker.svg")
+  withr::local_options("usethis.quiet" = FALSE)
+  expect_snapshot({
+    use_pkgdown()},  transform = scrub_testpkg)
+})
+
+test_that("use_pkgdown() nudges towards build_favicons().", {
+  skip_on_os("solaris")
```
olivroy, line 96:
```suggestion
  skip_if_not_installed("magick")
  skip_on_os("solaris")
```

olivroy:
I just realized that pkgdown is not a usethis dependency, so running the code may cause R CMD CHECK issues after all..

olivroy:
r-lib/pkgdown#2611 is a better solution!

PR #2001, opened 2024-06-09
Include file names in `challenge_uncommited_changes()`

olivroy:
Proof of concept to make the message shorter, and more focused on current situation.

Adding a `which` argument that conflicts with `msg`.

Rephrase message to be shorter.

Add a link to the uncommited files with `{.file`. `format_inline()` will shorten automatically as needed. 

I think it is worth adding files in the message, since it can facilitate the choice.

Let me know what you think!

This PR:

image

Before 

image

1. If I know which files are not commited, I spend more time figuring out if I should press 1, 2, or 3 for yes or no is than actually looking if I'm okay with continuing.
2. If I don't know what files, I have to go look elsewhere to figure out which files are not commited, and then go to do step 1.

---

Yields the following message (instead of the correctly formatted  with ! before.

```
    There are uncommitted changes and we're about to create and push to a new \
    GitHub repo 
```


jennybc:
I think this is a good pattern to reuse in this scenario:

```
> usethis::use_dev_version(push = TRUE)
✔ Adding "3.0.0.9000" to Version.
✔ Adding new heading to NEWS.md.
ℹ There are 2 uncommitted files:
• DESCRIPTION
• NEWS.md
! Is it ok to commit and push them?
...
```

This is ultimately coming from `git_ask_commit()`. I suspect there might be some refactoring that would help to share this code in more situations.

PR #2005, opened 2024-06-11
Add advice to view existing branches in `pr_resume()`

olivroy:
Let's say I call `pr_resume("git")`, but branch `git2` exists. `pr_resume()` will error but there is no hint that branch `git2` exists already.

This adds a runnable link to view local branches


jennybc:
`pr_resume()` (called with *no* input) already has a pretty nice experience, in terms of offering choices. I view that as the solution to "Hmmm, I don't know the exact name of the branch I want to checkout".

PR #2010, opened 2024-06-27
Tickle git pane with `pr_finish()`

### LINKED ISSUE #1992: RStudio git pane can get out of sync when checking out branch with the `pr_*()` functions
I'd suggest to add `rstudio_git_tickle()` in pr_resume()`

https://github.com/r-lib/usethis/blob/da18e11fa311ba7034ab58e86bb7b7e0977d1d26/R/pr.R#L265-L266

Also in `pr_finish()`, at the end of `pr_clean()`

I didn't follow the path of usage, but it feels harmless to add it in a couple places.

If you agree, I can open a PR.

olivroy:
don't know if this will work, but will test after merged

closes #1992

### DIFF
--- a/R/pr.R
+++ b/R/pr.R
@@ -532,6 +532,7 @@ pr_forget <- function() pr_clean(mode = "forget")
 pr_clean <- function(number = NULL,
                      target = c("source", "primary"),
                      mode = c("finish", "forget")) {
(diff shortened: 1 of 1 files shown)


jennybc:
I still don't recall experiencing a problem, but this seems to make sense / not do any harm in theory, so let's do it.

PR #2014, opened 2024-07-02
`git_sitrep()` works again without error if token is expired

olivroy:
@jennybc this is a fix to a regression that is caused by the switch to cli.

Basically, this removes some elements to the message object, since it contains many things that should not be printed.

```r
usethis::git_sitrep()
```

usethis 2.2.3
image
Current main
image
This PR
image
Doesn't show `Error in gh()`:

While working on this workaround fix, I saw another issue (not fixed).

The bullets are not properly formatted if a token is missing, but couldn't figure out how to fix this. 

I'd recommend you running `usethis::git_sitrep()` without a valid token to notice it.

Issue I am noticing (not addressed)
image

### DIFF
--- a/R/github_token.R
+++ b/R/github_token.R
@@ -168,6 +168,13 @@ pat_sitrep <- function(host = "https://github.com",
       message,
       "i" = maybe_who$error
     )
+    # Workaround cli eval (to remove extra information) #2014
+    message$i.trace <- NULL
+    message$i.response_headers <- NULL
+    message$i.call <- NULL
+    message$i.response_content <- NULL
+    message$i.rlang <- NULL
+    message$i.use_cli_format <- NULL
     ui_bullets(message)
     return(invisible(FALSE))
   }


jennybc:
> While working on this workaround fix, I saw another issue (not fixed).

Can you open a separate issue about that, so we don't lose track of it?

jennybc:
I can't seem to reproduce the problem.

I just installed from `main` and deleted my PAT from the credential store. Here's what I see:

```
> git_sitrep()

── Git global (user) 
• Name: "Jenny Bryan"
• Email: "jenny.f.bryan@gmail.com"
• Global (user-level) gitignore file: ~/.gitignore_global
• Vaccinated: TRUE
• Default Git protocol: ℹ Defaulting to "https" Git protocol. "https"
• Default initial branch name: "main"

── GitHub user 
• Default GitHub host: "https://github.com"
• Personal access token for "https://github.com": <unset>
☐ To create a personal access token, call `create_github_token()`.
☐ To store a token for current and future use, call `gitcreds::gitcreds_set()`.
ℹ Read more in the Managing Git(Hub) Credentials article.

...
```

Do you understand why we are having such a different experience?

olivroy:
I have ` rlang::global_entrace()` enabled in my .Rprofile as suggested by @lionel- in r-lib/pkgload#282.

jennybc:
So do I:

```
if (requireNamespace("rlang", quietly = TRUE)) {
    options(error = rlang::entrace, rlang_backtrace_on_error = "reminder")
}
```

Are you sure this isn't about scopes for the token? I don't see how this can be about the token being absent.

olivroy:
@jennybc 

I encountered this again. And to be clear, this only happens when the token is expired, not absent. (i.e. received multiple emails about his and ignored them) 
I create my token with usethis, so the permissions should be quite standard.

I updated my PR's title.

I guess that to experience this yourself, the best way would be to do.

Set the expiration date to tomorrow.
image

Add this token with `gitcreds::gitcreds_set()`

And call `usethis::git_sitrep()` tomorrow.

jennybc:
Ok yeah absent and expired are pretty different!

jennybc:
I found a simpler solution to this. I think this was just a think-o that probably predated the switch to cli.

jennybc:
What you see now with an expired token:

```
── GitHub user 
• Default GitHub host: "https://github.com"
• Personal access token for "https://github.com": <discovered>
✖ Can't get user information for this token.
ℹ The token may no longer be valid or perhaps it lacks the "user" scope.
ℹ GitHub API error (401): Bad credentials
```

Issue #2023, opened 2024-07-19
Which repo to target is asked many times without more context (e.g in `use_release_issue()`)


cderv:
Here is an example 
````r
> usethis::use_release_issue()
✔ Setting active project to "C:/Users/chris/Documents/DEV_R/quarto-r".
Which repo should we target? 

1: cderv/quarto-r = "origin"
2: quarto-dev/quarto-r = "upstream"

Sélection : 2
Current version is 1.4.3.9000.
What should the release version be? (0 to exit) 

1: major --> 2.0.0
2: minor --> 1.5.0
3: patch --> 1.4.4

Sélection : 3
Which repo should we target? 

1: cderv/quarto-r = "origin"
2: quarto-dev/quarto-r = "upstream"

Sélection : 2
Which repo should we target? 

1: cderv/quarto-r = "origin"
2: quarto-dev/quarto-r = "upstream"
````

First time ask to choose a repo, which makes sense for the action to where to open the issue. 

Though after it ask two more times, without context on the choice to make. Usually I select the same repo two times more. 

So either this is expected question, and maybe more context should be added, or the first choice could be cached and applied for next choices. 

I figured I would report as is - I did not look in code about what workflow is really happening in this case.

hadley:
Is this now fixed?

cderv:
Yes it is I believe - no more repetition of the question now. 👍 

````
> usethis::use_release_issue()
✔ Setting active project to "C:/Users/chris/Documents/DEV_R/rticles".
Which repo should we target? 

1: cderv/rticles = "origin"
2: rstudio/rticles = "upstream"

Sélection : 2
Current version is 0.27.3.
What should the release version be? (0 to exit) 

1: major --> 1.0.0
2: minor --> 0.28.0
3: patch --> 0.27.4

Sélection : 2
✔ Opening URL <https://github.com/rstudio/rticles/issues/575>.
````

jennybc:
Ha! My Positron training to never close an issue with a PR made me uses "addresses" instead of "fixes" or "closes" 😬

PR #2024, opened 2024-07-22
Stop repeating "which repo?" question in use_release_issue()

jennybc:
Addresses #2023 

We have to be careful around calling functions inside the `(release|upkeep)_checklist()` helpers that need to know which GitHub repo we're targetting. We're at risk of asking the user to choose the repo over and over again, with no context, if the user happens to have a fork of the source repo. We already _know_ the target repo at this point, so we propagate that info now.

### DIFF
--- a/R/release.R
+++ b/R/release.R
@@ -136,10 +136,6 @@ release_checklist <- function(version, on_cran, target_repo = NULL) {
 }
 
 gh_milestone_number <- function(target_repo, version, state = "open") {
-  if (!curl::has_internet()) {
-    return(NA)
-  }
-
   gh <- gh_tr(target_repo)
   milestones <- tryCatch(
     gh("/repos/{owner}/{repo}/milestones", state = state),
--- a/R/release.R
+++ b/R/release.R
@@ -56,7 +56,7 @@ use_release_issue <- function(version = NULL) {
   }
 
   on_cran <- !is.null(cran_version())
-  checklist <- release_checklist(version, on_cran)
+  checklist <- release_checklist(version, on_cran, tr)
 
   gh <- gh_tr(tr)
   issue <- gh(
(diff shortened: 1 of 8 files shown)


R/release.R
```diff
@@ -104,9 +97,7 @@ release_checklist <- function(version, on_cran) {
     "Prepare for release:",
     "",
     todo("`git pull`"),
-    if (!is.na(milestone_num)) {
-      todo("[Close v{version} milestone](../milestone/{milestone_num})")
-    },
+    todo("[Close v{version} milestone](../milestone/{milestone_num})", !is.na(milestone_num)),
```
jennybc, line 100: I think this was just a case of forgetting that `todo()` can be called this way.

R/release.R
```diff
@@ -144,11 +135,15 @@ release_checklist <- function(version, on_cran) {
   )
 }
 
-gh_milestone_number <- function(repo_spec, version, state = "open") {
-  milestones <- gh::gh(
```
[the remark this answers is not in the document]
hadley: Do we need this specifiy `curl::has_internet()` check? Won't it be caught below when `gh()` errors?

tests/testthat/test-release.R
```diff
@@ -130,12 +130,18 @@ test_that("returns empty string if no bullets", {
 test_that("can find milestone numbers", {
   skip_if_offline("github.com")
 
+  tr <- list(
```
hadley, line 133: I think it would be worth adding a test that `gh_milestone_number()` returns `NA` when `gh()` errors (e.g. because you've mocked it to always error)

hadley (approved):

R/release.R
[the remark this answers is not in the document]
jennybc: OK I removed that check. It does mean you can be exposed to some reporting coming out of gh around retrying, in the absence of an internet connection, but that doesn't seem like a big deal.

tests/testthat/test-release.R
jennybc: done

PR #2038, opened 2024-08-15
Dots in rename

### LINKED ISSUE #1969: rename_files breaks with . in old or new
I'm not sure how to create a complete reprex for this. I definitely recommend making sure everything is committed in whatever directory you're using if you're in a real project.

I was trying to organize a directory better, and had the idea of using "." within some filenames. It wouldn't have actually sorted how I want, but I would have seen that when I tried.

Except it truncated at the `.`, so 

``` r
rename_files("generate_security", "generate_pkg.security")
```

was treated as

``` r
rename_files("generate_security", "generate_pkg")
```

Once I ran it it made sense, but I wasn't expecting it, and it made a mess of some things. I think I've recovered everything, but it would definitely be possible to lose work.

I'd recommend either requiring a confirmation if anything other than `.R` would be truncated, and/or maybe requiring that all work be checked in before running (rather than just checking for using git).

I can probably PR if you'd like to implement this. Let me know if you have a preference for what to implement.

jonthegeek:
Check for uncommitted changes before renaming files, and only remove `.R` file extensions in `old` and `new`.

Fixes #1969.

I *think* this is the least-surprising approach to take. It's probably best to just not use `.` in filenames, but this will stop things from breaking in confusing ways when people do.

### DIFF
--- a/R/rename-files.R
+++ b/R/rename-files.R
@@ -16,6 +16,12 @@
 #' @export
 rename_files <- function(old, new) {
   check_uses_git()
+  challenge_uncommitted_changes(
+    msg = "
+    There are uncommitted changes and we're about to bulk-rename files. It is \\
+    highly recommended to get into a clean Git state before bulk-editing files",
+    untracked = TRUE
+  )
 
   old <- path_ext_remove(old)
   new <- path_ext_remove(new)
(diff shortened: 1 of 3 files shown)


jonthegeek:
Oops, I forgot I copy/pasted that!

jennybc:
Thanks!

PR #2039, opened 2024-08-15
suppress git_protocol message in git_sitrep

### LINKED ISSUE #2034: Funky bullet in `git_sitrep()`
The line for "Default Git protocol" is wonky. I suspect we need to suppress the message 'i Defaulting to "https" Git protocol'.

```
> git_sitrep()

── Git global (user) 
• Name: "Jenny Bryan"
• Email: "jenny.f.bryan@gmail.com"
• Global (user-level) gitignore file: [~/.gitignore_global](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
• Vaccinated: TRUE
• Default Git protocol: ℹ Defaulting to "https" Git protocol. "https"
• Default initial branch name: "main"
```

thomascwells:
Fixes #2034 

If the default git communication protocol isn't set, `git_protocol()` sets a default *and* messages the user about the chosen default. That message looks confusing in the output bullets for `git_sitrep()`. This PR suppresses that message.

Putting `suppressMessages()` directly in the `kv_line()` call didn't seem to work (maybe something about the environment?), so I split it into two steps. 

These chunks were useful for testing changes since I had some of these options set in my user-level .Rprofile:
```r

devtools::load_all()

options("usethis.protocol" = NULL)
git_protocol()

options("usethis.protocol" = NULL)
git_sitrep()
```

### DIFF
--- a/R/git.R
+++ b/R/git.R
@@ -357,7 +357,8 @@ git_sitrep <- function(tool = c("git", "github"),
     if (!vaccinated) {
       ui_bullets(c("i" = "See {.fun usethis::git_vaccinate} to learn more."))
     }
-    kv_line("Default Git protocol", git_protocol())
+    protocol <- suppressMessages(git_protocol())
+    kv_line("Default Git protocol", protocol)
     kv_line("Default initial branch name", init_default_branch)
   }
 
--- a/R/git.R
+++ b/R/git.R
@@ -357,8 +357,7 @@ git_sitrep <- function(tool = c("git", "github"),
     if (!vaccinated) {
       ui_bullets(c("i" = "See {.fun usethis::git_vaccinate} to learn more."))
     }
-    protocol <- suppressMessages(git_protocol())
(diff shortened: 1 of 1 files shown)


R/git.R
```diff
@@ -357,7 +357,8 @@ git_sitrep <- function(tool = c("git", "github"),
     if (!vaccinated) {
       ui_bullets(c("i" = "See {.fun usethis::git_vaccinate} to learn more."))
     }
-    kv_line("Default Git protocol", git_protocol())
+    protocol <- suppressMessages(git_protocol())
```
jennybc, line 360: I think `ui_silence()` is probably our official way of doing this. And it may be better to do that, so that we only silence output coming from usethis itself. It probably doesn't really matter in practice, but it is what I would do here, I think, so let's try it.

jennybc:
Thanks!

PR #2042, opened 2024-08-15
Only print revdep checklist items if revdeps exist for the current package

### LINKED ISSUE #1817: Suggest checking reverse dependencies only if they exists
When using `use_release_issue()`, it always warns to run reverse dependency checks  via https://github.com/r-lib/usethis/blob/04ebe9e1210a44b3d59c3c2be5371706bb31d106/R/release.R#L158 

I think it might be easy to make it conditionally to if there is any dependency in CRAN. 
Something along:

```r
if (lengths(tools::package_dependencies(pkg, which = "all", reverse = TRUE)) == 0) {
    return(NULL)
}
```

For new releases and new package developers it will be easier to go through one less item.
It can be a nice warning to developers when package get their first reverse dependency. 

I see usethis is already hitting the 20 dependencies limit. Maybe `tools` can be "hidden" as a Suggests or use one of the current dependencies for the same functionality.

seankross:
Fixes #1817

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -39,6 +39,7 @@ Imports:
     rprojroot (>= 1.2),
     rstudioapi,
     stats,
+    tools,
     utils,
     whisker,
     withr (>= 2.3.0),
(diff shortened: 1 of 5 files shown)


jennybc:
Thanks!

PR #2043, opened 2024-08-15
Let `use_package()` decrement a package dependency.

### LINKED ISSUE #1957: use_package() `min_version` will not decrement
When `min_version` is provided and it is a smaller semver than the one currently in the DESCRIPTION file, it will not be decremented to the lower version. 

```r
usethis::use_package("dplyr", min_version = "1.0.0")
usethis::use_package("dplyr", min_version = "0.8.0")
```

jplecavalier:
Fixes #1957.

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -63,11 +63,11 @@ use_dependency <- function(package, type, min_version = NULL) {
              {.field {existing_type}} in DESCRIPTION; no change made."
     ))
   } else if (delta == 0 && version_spec(version) != version_spec(existing_version)) {
-    # change version
     if (version_spec(version) > version_spec(existing_version)) {
       direction <- "Increasing"
+    } else {
+      direction <- "Decreasing"
     }
-    else direction <- "Decreasing"
 
     ui_bullets(c(
       "v" = "{.drn {direction}} {.pkg {package}} version to {.val {version}} in
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -62,19 +62,19 @@ use_dependency <- function(package, type, min_version = NULL) {
       "!" = "Package {.pkg {package}} is already listed in
              {.field {existing_type}} in DESCRIPTION; no change made."
     ))
-  } else if (delta == 0 && !is.null(min_version)) {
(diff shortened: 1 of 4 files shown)


R/helpers.R
```diff
@@ -62,19 +62,19 @@ use_dependency <- function(package, type, min_version = NULL) {
...
-      ui_bullets(c(
-        "v" = "Increasing {.pkg {package}} version to {.val {version}} in
-               DESCRIPTION."
-      ))
-      desc$set_dep(package, type, version = version)
-      desc$write()
-      changed <- TRUE
-    }
+    direction <- ifelse(version_spec(version) > version_spec(existing_version),
+                        "Increasing", "Decreasing")
```
jennybc, line 68: Can you change this to an `if () ... else`? I generally reserve `ifelse()` for vectorized situations.

R/helpers.R
```diff
@@ -62,19 +62,19 @@ use_dependency <- function(package, type, min_version = NULL) {
...
+
+    ui_bullets(c(
+      "v" = "{.drn {direction}} {.pkg {package}} version to {.val {version}} in
```
jennybc, line 71: I'm not familiar with the `.drn` inline markup. Can you help me understand that?

jennybc (commented):
Looking good! I had a couple comments/questions.

R/helpers.R
jplecavalier: Sure!

R/helpers.R
jplecavalier: I just tried to replicate the same logic as in `{.pkg {package}}` and`{.val {version}}`. I tried to remove `.drn` and then `{direction}` would not resolve with the string in the object `direction`. Any hint?

jennybc: Something like `{.pkg {package}}` does interpolation of `package` through its surrounding brackets. Then `.pkg` and the second layer of brackets applies some inline styling.

For pure interpolation (which is the goal with `direction`), we just need one pair of brackets.

You can learn more about this in the cli package: https://cli.r-lib.org/reference/inline-markup.html.

jplecavalier: Thanks for pointing this out, and thanks again for your time and patience today!

PR #2044, opened 2024-08-15
Bump default serialization version in use_data() to 3

### LINKED ISSUE #1966: Bump the default version argument in `use_data()` to 3
Now we're a few years past the R 3.6 release is it worth bumping the default `version` argument for `use_data()` from 2 to 3 (matching R's current default)?

laurabrianna:
Fixes #1966  - The grand sum of this update is one character (plus a news bullet) Note that this change did not break anything in any test (should a test be added?)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* `use_data()` now uses serialization version 3 by default. (@laurabrianna, #2044)
+
 # usethis 3.0.0
 
 ## Transition to cli package for UI
--- a/R/data.R
+++ b/R/data.R
@@ -38,7 +38,7 @@ use_data <- function(...,
                      internal = FALSE,
                      overwrite = FALSE,
                      compress = "bzip2",
-                     version = 2,
+                     version = 3,
                      ascii = FALSE) {
   check_is_package("use_data()")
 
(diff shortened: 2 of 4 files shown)


jennybc (commented):
Looking good!

Can you take a look at the docs for the `version` argument? They state that the default is 2, which is not correct anymore. So this needs a refresh.

```
#' @param version The serialization format version to use. The default, 2, was
#'   the default format from R 1.4.0 to 3.5.3. Version 3 became the default from
#'   R 3.6.0 and can only be read by R versions 3.5.0 and higher.
```

jennybc:
I think we *do* have a test that claims to probe default behaviour but it's clearly not working! Because it actually provides the `version` argument.

Would you also like to fix and update this test?

```
test_that("use_data() writes version 2 by default", {
  create_local_package()

  x <- letters
  use_data(x, internal = TRUE, version = 2, compress = FALSE)
  expect_identical(
    rawToChar(readBin(proj_path("R", "sysdata.rda"), n = 4, what = "raw")),
    "RDX2"
  )
})
```

laurabrianna:
Hi Jenny thanks for all your help (and patience!) I incorporated both your notes above!

jennybc:
Thanks!

PR #2045, opened 2024-08-15
Add `create_quarto_project()`

### LINKED ISSUE #1891: Add `create_quarto_project` or support Quarto-related parameters in `create_project`
In exploring options for creating a R Project file, I found `create_project()` and wondered if it might be enhanced by the addition of new parameters to better support the creation of Quarto projects.

A simple `create_quarto_project()` that includes the basic `_quarto.yml` settings that is created when selecting "Quarto Project" from the New Project dialog in RStudio could be one option.

Another option could be a project type parameter, e.g. "project", "website", "book", "blog", corresponding to those same default settings.

One last (but not mutually exclusive) option could be updating `create_project()` to take a `config_file` parameter (similar to `use_pkgdown()`) and offering some way to define a set of files to use as a starting place (maybe using a `repo_spec` parameter similar to `use_standalone()`).

If there is interest in this feature, I'd be happy to work on it and open a pull request.

focardozom:
Closes #1891 

~`quarto::quarto_create_project()` does not set a new directory; it only creates a folder containing the `_qmd` and `.qmd` files. I’m not sure if `usethis` should be used to set the new directory and restar the session as in `create_project()`.~ *This PR evolved into making this more like the other `create_*()` functions.*


R/create.R
```diff
@@ -2,12 +2,17 @@
 #'
 #' @description
 #' These functions create an R project:
-#'   * `create_package()` creates an R package
+#'   * `create_package()` creates an R package.
 #'   * `create_project()` creates a non-package project, i.e. a data analysis
-#'   project
+#'     project.
+#'   * `r lifecycle::badge("experimental")` `create_quarto_project()` creates a
```
jennybc, line 8: I'm marking this as experimental because, e.g., the behaviour when called inside an existing project is not so nice (mostly a limitation rooted in the quarto cli when called like this). I've also exposed the bare minimum of arguments. I want to make it clear that this might very well need some refinement.

R/create.R
```diff
@@ -19,17 +24,22 @@
...
 #'   Project](https://r-pkgs.org/workflow101.html#sec-workflow101-rstudio-projects).
-#'    If `FALSE` and a non-package project, a sentinel `.here` file is placed so
-#'   that the directory can be recognized as a project by the
-#'   [here](https://here.r-lib.org) or
-#'   [rprojroot](https://rprojroot.r-lib.org) packages.
+#'
+#'   If `FALSE`, the goal is to ensure that the directory can be recognized as
+#'   a project by, for example, the [here](https://here.r-lib.org) package. If
+#'   the project is neither an R package nor a Quarto project, a sentinel
+#'   `.here` file is placed to mark the project root.
```
jennybc, line 31: This is mildly aspirational. I have only recently added `rprojroot::is_quarto_project` to usethis's project-finding criteria and I have a PR in flight to have the here package do the same: https://github.com/r-lib/here/pull/130. But I think this will hopefully be true soon anough.

R/create.R
```diff
@@ -2,12 +2,17 @@
...
 #' @description
 #' These functions create an R project:
-#'   * `create_package()` creates an R package
+#'   * `create_package()` creates an R package.
 #'   * `create_project()` creates a non-package project, i.e. a data analysis
-#'   project
+#'     project.
+#'   * `r lifecycle::badge("experimental")` `create_quarto_project()` creates a
+#'     Quarto project. It is a simplified convenience wrapper around
+#'     `quarto::quarto_create_project()`, which you should call directly for
```
jennybc: I can't actually link to `quarto::quarto_create_project()` because quarto is only in `Suggests`.

jennybc:
@focardozom Will you take this for a test drive and see if it feels right?

NEWS.md
```diff
@@ -10,10 +10,15 @@
   is not specified (#2117, @salim-b).
 
 * usethis's criteria for recognizing a project have expanded to include (#2133):
-  - a `.vscode/` directory, which Positron or VS Code might create
+  - a `.vscode/settings.json` file, which Positron or VS Code might create
   - a `_quarto.yml` file, typical of a Quarto project
   - an `renv.lock` file, which renv creates
 
+* `create_quarto_project()` is a new experimental function that combines basic
+  usage of `quarto::quarto_create_project()` with some of the niceties of
```
jennybc, line 20: I think putting `quarto::quarto_create_project()` inside the `usethis::create_*()` envelope is the main contribution here. E.g. challenging a project-within-a-project and opening / switching to the newly created Quarto project.

R/create.R
```diff
@@ -114,6 +124,46 @@ create_project <- function(
   invisible(proj_get())
 }
 
+#' @rdname create_package
+#' @export
+create_quarto_project <- function(
+  path,
+  type = "default",
+  rstudio = rstudioapi::isAvailable(),
```
cderv, line 134: `rstudio` does not seem used. Is this for future usage? 

`quarto::quarto_create_project()` won't create an RStudio project.

cderv (approved):
Looks good! 

Just a question on argument to the wrapper

R/create.R
jennybc: Good point! I think I was thinking that we don't *need* to mark it as an RStudio Project since it will have `_quarto.yml`. But that's really getting ahead of myself and for now it makes more sense to handle more similar to what's done in `create_project()`. The whole matter of placing `.Rproj` needs a lot of thought in the Positron era.

PR #2046, opened 2024-08-16
Adding pre-commit hook for README even when not overwritten.

### LINKED ISSUE #2035: `use_readme_rmd()` does not create pre-commit hook if `README.Rmd` is not overwritten
If `use_readme_rmd()` is run in a project where `README.Rmd` already exists, it asks the user if they want to overwrite it with a template. If they answer no, the git pre-commit hook in `./.git/hooks/pre-commit` is not created.

To work around this I had to answer "yes" to overwriting and then run `git restore README.Rmd` which is not ideal.

jplecavalier:
Fixes #2035. @jennybc I don't know if you were planning to resolve this issue or not, feel free to reject my PR if you disagree with the proposal.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,9 @@
 
 ## Bug fixes and minor improvements
 
+* `use_readme_rmd()` now adds the pre-commit hook for making sure README.Rmd
+  was knited before commit (@jplecavalier, #2035).
+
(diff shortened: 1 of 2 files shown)


jennybc:
See #2035 for why I've decided not to pursue this. Thanks for the PR, in any case!

PR #2048, opened 2024-08-22
Fix some lints

etiennebacher:
Hello, I'm building a new package to find and automatically fix lints in R code: [`flint`](https://flint.etiennebacher.com/).

I'm using real-life, large packages to check its performance (both in speed and in correctness of the fixes) and `usethis` is one of them. Since I already test on this, there's almost no additional cost for me in proposing those changes.

FYI, those changes were generated with `flint::fix_package()`.

Most changes are trivial:
- replace `any(is.na())` by `anyNA()`
- replace `expect_identical(x, NULL)` by `expect_null(x)`
etc.

There's also likely a bug fix: `length(old > 0)` is replaced by `length(old) > 0`.

`flint` is quite new and linter rules are necessarily a bit opinionated (but I follow `lintr` most of the time), so some of those changes might not be to your taste, but I'd be happy to have some feedback on this.

### DIFF
--- a/tests/testthat/test-utils-github.R
+++ b/tests/testthat/test-utils-github.R
@@ -11,7 +11,7 @@ test_that("parse_github_remotes() works, on named list or named character", {
     bitbucket2 = "git@bitbucket.org:OWNER/REPO.git"
   )
   parsed <- parse_github_remotes(urls)
-  expect_equal(parsed$name, names(urls))
+  expect_named(urls, parsed$name)
   expect_equal(unique(parsed$repo_owner), "OWNER")
   expect_equal(
     parsed$host,
--- a/tests/testthat/test-utils-github.R
+++ b/tests/testthat/test-utils-github.R
@@ -11,7 +11,7 @@ test_that("parse_github_remotes() works, on named list or named character", {
     bitbucket2 = "git@bitbucket.org:OWNER/REPO.git"
   )
   parsed <- parse_github_remotes(urls)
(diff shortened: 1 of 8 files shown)


tests/testthat/test-utils-github.R
```diff
@@ -11,7 +11,7 @@ test_that("parse_github_remotes() works, on named list or named character", {
     bitbucket2 = "git@bitbucket.org:OWNER/REPO.git"
   )
   parsed <- parse_github_remotes(urls)
-  expect_equal(parsed$name, names(urls))
+  expect_named(urls, parsed$name)
```
jennybc, line 14: This one seems to change the intent of the expectation. As in, I just created `urls` by hand with specific names. So I'm testing something about `parsed`, not about `urls`. This probably still succeeds / fails appropriately, but it somehow feels off.

jennybc (commented):
Overall, the changes seem neutral to good and I agree this seems to have caught a bug.

I commented on one change that I don't love.

Thanks!

tests/testthat/test-utils-github.R
etiennebacher: Thanks for the feedback. I agree it's less readable, this is one of the defaults of `lintr` that I will maybe deviate from.

etiennebacher:
Done, the `pkgdown` failure is unrelated

jennybc:
Thanks!

Issue #1681, opened 2022-09-20
Function to create initial data docs


focardozom:
Hi! document data is always tricky, so having a function to help people document the dataset would be fantastic. This function could create an .R file in the R folder containing information gathered from the data set. For example, the information can be inserted in a roxygen template, and the @items values can be filled using `glue()`. The template could also include some descriptive information to help users understand the dataset better.

jennybc:
Would you like to make a PR for consideration? Full disclosure: I'm not 100% convinced that usethis should do this. But this is the topic of an issue I recently closed in R Packages, which contains some concrete ideas to start with.

https://github.com/hadley/r-pkgs/issues/707

jennybc:
As I continue to work on R Packages, I've learned there are packages that already offer functions to do this. One example is sinew (https://cran.r-project.org/web/packages/sinew/index.html). So given that there are solutions out there already, I don't think it's a priority for us to add this to usethis.

ngreifer:
Just wanted to add to this since I had the same thought. `sinew` does not fit well into the package development ecosystem for `usethis` users; it prints a string rather than creating a file, doesn't use Markdown syntax, doesn't check for existing Roxygen etc. `usethis` offers `use_package_doc()`, which is a major help, and I think the same could be done by a function, e.g., `use_data_doc()`, which takes in a dataset and creates a file with a Roxygen skeleton. It could also be an argument to `use_data()`, e.g., `use_data(., doc = TRUE)` which saves the dataset to `/data` and creates a documentation file. Thank you for considering and making such a useful package!

focardozom:
> Just wanted to add to this since I had the same thought. `sinew` does not fit well into the package development ecosystem for `usethis` users; it prints a string rather than creating a file, doesn't use Markdown syntax, doesn't check for existing Roxygen etc. `usethis` offers `use_package_doc()`, which is a major help, and I think the same could be done by a function, e.g., `use_data_doc()`, which takes in a dataset and creates a file with a Roxygen skeleton. It could also be an argument to `use_data()`, e.g., `use_data(., doc = TRUE)` which saves the dataset to `/data` and creates a documentation file. Thank you for considering and making such a useful package!

Hi, based on @jennybc suggestions, I created [this function](https://github.com/focardozom/DocumentData ) to help with the documentation. This is under construction, so any comments will be constructive. I like your idea of use_data(., doc = TRUE), so maybe they will re-open this issue in the future. I am also a big fan of use_this.

[...]

PR #2049, opened 2024-08-22
document data 

focardozom:
Potential fix for #1681. Here’s a summary of the relevant points discussed in the issue:

1. **Existing Tools**: The [sinew](https://github.com/yonicd/sinew) package provides functions for documenting, but its workflow might overlap with `usethis`. More details can be found in the [workflow guide](https://yonicd.github.io/sinew/articles/workflow.html).
   
2. **Alternative Solution**: Another approach is proposed [here](https://gist.github.com/friendly/14f3ee1464213bb0b9fbcb489468383b), which handles various R objects and offers customizable documentation. However, it may require style translation and simplification to make it more accessible.
   
3. **This Pull Request**: The `use_document_data` function is simpler and easier to read, though it doesn’t offer the same level of customization as the alternative.


jennybc:
This isn't a full review, but I came by quickly to see if this pkgdown failure is related to one seen in a different PR. It doesn't seem to be.

The pkgdown CI failure appears to be real because the new functions need to be added to the reference index. So that would be good to fix up to make this PR complete.

R/use_document_data.R
```diff
@@ -0,0 +1,99 @@
...
+#'         for the data set. This includes the data set's basic information and
+#'         placeholders for detailed descriptions of each variable.
+#'
+#' @keywords internal
+#'
+create_data_description <- function(dataset, name, description, source) {
+  data_info <- generate_data_info(dataset)
+  variable_descriptions <- generate_variable_descriptions(dataset)
+
+  description_template <- paste0(
```
ijlyttle, line 73: In time, this could be done using template file and the `use_template()` function. I would not worry about this now, but it would not surprise me if this were used, once things become more settled.

R/use_document_data.R
```diff
@@ -0,0 +1,99 @@
...
+    "#' ", paste(variable_descriptions, collapse = "\n#' "), "\n#' }\n",
+    "#' @source ", source, "\n",
+    "'", name, "'"
+  )
+
+  return(description_template)
+}
+
+generate_data_info <- function(dataset) {
+  paste0(
```
ijlyttle, line 89: Within usethis, newer functions use `glue()`, rather than `paste0()`, would you consider refactoring?

R/use_document_data.R
```diff
@@ -0,0 +1,99 @@
...
+    "A ", typeof(dataset), " [", class(dataset), "] with ",
+    nrow(dataset), " rows and ", length(names(dataset)), " variables:\n"
+  )
+}
+
+generate_variable_descriptions <- function(dataset) {
+  purrr::map_chr(names(dataset), function(var) {
+    paste0("\\item{", var, "} {", class(dataset[[var]]), ": Type label here}")
```
ijlyttle, line 97: same here for `glue()`, vs. `paste0()`

ijlyttle (commented):
As part of Tidyverse Dev Day, I am helping @jennybc move this PR forward. 
Having spoken with Jenny, we agree this is an important capability to bring into usethis  and that this PR is greatly appreciated.

We also realize that this PR has been around for a while, and the widespread adoption
of LLM (e.g. Positron) is changing a lot of workflows. For example, we might imagine providing a skeleton
containing material known to be true, which could provide an AI assistant with the opportunity to 
fill in the blanks on more-subjective things, like the description.

As such, my approach to this review is to focus on a minimal set of things we can merge to move the ball forward.

It's difficult to add a function to be exported, then to remove it in case we want to do things differently. 
So I think it could be useful to focus on merging the internal functions, using that as a starting point for further work.

Would you consider, for the purposes of this PR, removing the `use_document_data()` function, so that work can continue on 
`create_data_description()`, `generate_data_info()`, and `generate_variable_descriptions()`?

As well. would you consider contributing some tests for `generate_data_info()`, and `generate_variable_descriptions()`? 

Thanks again for taking a swing at this!

hadley:
I think I'm going to call it on this PR. Unfortunately we haven't made any progress in two years, and as @ijlyttle points out LLMs have really changed the game (even more in the last 12 months). So while I really appreciate all your work on this issue (and I hope you learned something along the way!) I'm going to close it.

Issue #1929, opened 2023-12-13
Some HTTP 403 errors offer context


aronatkins:
Using usethis 2.2.2, I struggled to understand why a refreshed GitHub token was not usable. Even after the token-update dance, I was still seeing errors like the following from `usethis::use_release_issue()`:

```
Error: Unable to get GitHub info for these remotes: 'origin'
Are we offline? Is GitHub down? Has the repo been deleted?
Otherwise, you probably need to configure a personal access token (PAT) for 'github.com'
See `?gh_token_help` for advice
```

The feedback from `usethis::git_sitrep()` was similar:

```
• Desc = 'origin' is a GitHub repo and 'upstream' is either not configured or is not a GitHub repo.
  
  We may be offline or you may need to configure a GitHub personal access
  token. `gh_token_help()` can help with that.
  
  Read more about what this GitHub remote configurations means at:
  'https://happygitwithr.com/common-remote-setups.html'
```

I did not uncover any actionable feedback until digging into the lower-level calls made by usethis. Directly calling `gh::gh()` gave me:

```
Error in `gh::gh()`:
! GitHub API error (403): Resource protected by organization SAML enforcement. You must
  grant your Personal Access token access to this organization.
ℹ Read more at
  <https://docs.github.com/articles/authenticating-to-a-github-organization-with-saml-single-sign-on/>
Run `rlang::last_trace()` to see where the error occurred.
```

This was enough of a reminder - our SSO needs tokens to be authorized (a separate step after creation / refresh).

Is it possible for context about the original error to be provided to some of the higher-level calls?

PR #2053, opened 2024-09-09
Some notes for discussion

hadley:


R/utils-github.R
```diff
@@ -171,67 +171,103 @@ github_remote_list <- function(these = c("origin", "upstream"), x = NULL) {
...
+        c(
+          "Can't reach {.url {api_url}}.",
+          "Are you offline? Is GitHub down?"
+        ),
+        parent = cnd,
+        call = error_call
+      )
+    },
+    httr2_http_404 = function(cnd) {
+      # Might be because it's a private repo & no PAT
```
hadley, line 251: https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#404-not-found-for-an-existing-resource

R/utils-github.R
```diff
@@ -171,67 +171,103 @@ github_remote_list <- function(these = c("origin", "upstream"), x = NULL) {
...
+            "Do you need to configure a GitHub token?",
+            "See {.run usethis::gh_token_help()} for advice."
+          )
+        ),
+        parent = cnd,
+        call = error_call
+      )
+    },
+    httr2_http_403 = function(cnd) {
+      # No token?
```
hadley, line 265: I don't _think_ it's possible to get this error?

jennybc:
Not a full review but glancing at this reminded me: do we need to use "http" here instead of "httr2"? If you look around, that's what's done in similar situations already. E.g. "http_error_403". Didn't we learn that's how gh classes things?

R/utils-github.R
hadley: Nope, see https://github.com/r-lib/usethis/issues/1929

hadley:
Giving up on this approach

PR #2055, opened 2024-09-10
Use `github_remote_list()` instead of `github_remotes()` when GitHub not consulted

hadley:
To reduce some of the usage of `github_remotes()` where it's not strictly necessary.

### DIFF
--- a/R/pr.R
+++ b/R/pr.R
@@ -662,7 +662,7 @@ pr_create <- function() {
   branch <- git_branch()
   tracking_branch <- git_branch_tracking(branch)
   remote <- remref_remote(tracking_branch)
-  remote_dat <- github_remotes(remote, github_get = FALSE)
+  remote_dat <- github_remote_list(remote)
   ui_bullets(c("_" = "Create PR at link given below."))
   view_url(glue_data(remote_dat, "{host_url}/{repo_spec}/compare/{branch}"))
 }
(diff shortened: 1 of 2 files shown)


jennybc:
My main reaction is: surely I had some reason for using `github_remotes(github_get = FALSE)` instead of `github_remote_list()` (?). Because obviously I had both fully uploaded into my brain when I did this. I'm just getting some Chesterton's Fence vibes. It's hard to know what will happen since none of this is under test.

hadley:
`new_*()`  are _only_ used for tests, so I think that's safe.

jennybc:
> `new_*()` are only used for tests, so I think that's safe.

I do have some memory that the whole reason I made the `new_*()` functions this way was so that I could authentically "shaped" github remote info.

I.e. sure it is github remote info that is fake, but I make it in such a way that it's guaranteed to have the same "shape" as when we actually do call github. I think this can't be detected today (meaning: tests aren't failing), because the assignments we go on to make happen to create all the right fields. The fields that are no longer there due to the switch from `github_remotes(github_get = FALSE)` to `github_remote_list()` all get created now, instead of just modified.

jennybc:
So I think this implies that the PR should probably only propose a change in `pr_create()`?

hadley:
Giving up on this approach

PR #2062, opened 2024-09-20
Where possible use existing github config when determining default branch

### LINKED ISSUE #2021: `pr_` performance
In my heart of hearts I feel like the performance of the `pr_` functions is slower than it should be. I don't really have any evidence to base this feeling on, but now that [dev profvis](https://r-lib.github.io/profvis/dev/) defaults to elapsed timing, we now have a way to profile the code and see which API calls are taking up the time.

hadley:
Fixes #2021

### DIFF
--- a/R/git-default-branch.R
+++ b/R/git-default-branch.R
@@ -93,39 +93,18 @@ NULL
 #' git_default_branch()
 #' }
 git_default_branch <- function() {
+  git_default_branch_(github_remote_config())
+}
+
+# If config is available, we can use it to avoid an additional lookup
+# on the GitHub API
+git_default_branch_ <- function(cfg) {
   repo <- git_repo()
 
-  # TODO: often when we call git_default_branch(), we already have a GitHub
-  # configuration or target repo, as produced by github_remote_config() or
-  # target_repo(). In that case, we don't need to start from scratch as we do
-  # here. But I'm not sure it's worth adding complexity to allow passing this
-  # data in.
-
-  # TODO: this critique feels somewhat mis-placed, i.e. it brings up a general
-  # concern about a repo's config (or the user's permissions and creds)
-  # related to whether github_remotes() should be as silent as it is about
-  # 404s
-  critique_remote <- function(remote) {
-    if (remote$is_configured && is.na(remote$default_branch)) {
-      ui_bullets(c(
-        "x" = "The {.val {remote$name}} remote is configured, but we can't
-               determine its default branch.",
-        " " = "Possible reasons:",
-        "*" = "The remote repo no longer exists, suggesting the local remote
-               should be deleted.",
-        "*" = "We are offline or that specific Git server is down.",
-        "*" = "You don't have the necessary permission or something is wrong
-               with your credentials."
-      ))
-    }
-  }
-
-  upstream <- git_default_branch_remote("upstream")
+  upstream <- git_default_branch_remote(cfg, "upstream")
   if (is.na(upstream$default_branch)) {
-    critique_remote(upstream)
-    origin <- git_default_branch_remote("origin")
+    origin <- git_default_branch_remote(cfg, "origin")
     if (is.na(origin$default_branch)) {
-      critique_remote(origin)
       db_source <- list()
     } else {
       db_source <- origin
@@ -186,7 +165,7 @@ git_default_branch <- function() {
 
 # returns a whole data structure, because the caller needs the surrounding
 # context to produce a helpful error message
-git_default_branch_remote <- function(remote = "origin") {
+git_default_branch_remote <- function(cfg, remote = "origin") {
   repo <- git_repo()
   out <- list(
     name = remote,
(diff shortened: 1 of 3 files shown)


R/git-default-branch.R
```diff
@@ -93,39 +93,18 @@ NULL
...
 git_default_branch <- function() {
-  repo <- git_repo()
+  git_default_branch_(github_remote_config())
+}
 
-  # TODO: often when we call git_default_branch(), we already have a GitHub
-  # configuration or target repo, as produced by github_remote_config() or
-  # target_repo(). In that case, we don't need to start from scratch as we do
-  # here. But I'm not sure it's worth adding complexity to allow passing this
-  # data in.
```
hadley, line 102: Deleted because that's what I did 😄

R/git-default-branch.R
```diff
@@ -93,39 +93,18 @@ NULL
...
-               determine its default branch.",
-        " " = "Possible reasons:",
-        "*" = "The remote repo no longer exists, suggesting the local remote
-               should be deleted.",
-        "*" = "We are offline or that specific Git server is down.",
-        "*" = "You don't have the necessary permission or something is wrong
-               with your credentials."
-      ))
-    }
-  }
```
hadley, line 121: Deleted because I think it's now more clear that it's the responsibility of `github_remote_config()` to report such problems, and `github_remote_config()` is now more consistently called before `git_default_branch()` where it matters.

R/github.R
```diff
@@ -61,7 +61,7 @@ use_github <- function(organisation = NULL,
   visibility <- match.arg(visibility)
   check_protocol(protocol)
   check_uses_git()
-  default_branch <- git_default_branch()
+  default_branch <- guess_local_default_branch()
```
hadley, line 64: I think this a reasonable substitution because I can't see why origin or upstream would be setup _before_ you're calling `use_github()`. Please let me know if I've forgotten some likely scenario.

R/pr.R
```diff
@@ -237,7 +237,7 @@ pr_resume <- function(branch = NULL) {
     ui_bullets(c(
       "i" = "No branch specified ... looking up local branches and associated PRs."
     ))
-    default_branch <- git_default_branch()
+    default_branch <- guess_local_default_branch()
```
hadley, line 240: I think it's ok to use a cheap check here because we're only using it to exclude branches, and this (along with the tweak below) allows `pr_resume()` to work offline.

R/pr.R
```diff
@@ -491,11 +492,11 @@ pr_view <- function(number = NULL, target = c("source", "primary")) {
 #' @export
 #' @rdname pull-requests
 pr_pause <- function() {
-  # intentionally naive selection of target repo
-  tr <- target_repo(github_get = FALSE, ask = FALSE)
+  cfg <- github_remote_config(github_get = NA)
```
hadley, line 495: This seems like a change, but isn't really because the call to `git_default_branch()` on line 498 would have used `github_get = NA`.

R/pr.R
```diff
@@ -629,14 +632,10 @@ pr_clean <- function(number = NULL,
 # we're in DEFAULT branch of a fork. I wish everyone set up DEFAULT to track the
 # DEFAULT branch in the source repo, but this protects us against sub-optimal
 # setup.
-pr_pull_source_override <- function(tr = NULL, default_branch = NULL) {
-  # naive selection of target repo; calling function should analyse the config
-  tr <- tr %||% target_repo(github_get = FALSE, ask = FALSE)
-
+pr_pull_source_override <- function(tr, default_branch) {
```
hadley, line 635: All callers of `pr_pull_source_override()` supply their arguments and I think we should continue to enforce that.

hadley (commented):
I think this is a more surgical approach to getting the speed up that I'm looking for. Hopefully you agree!

R/git-default-branch.R
hadley: For example, if you're offline, `pr_resume()` will emit this message even though you can still checkout the local branches that it lists.

R/pr.R
jennybc: I definitely remember that the only reason I had `NULL` defaults here was for development convenience.

R/pr.R
```diff
@@ -629,14 +632,10 @@ pr_clean <- function(number = NULL,
 # we're in DEFAULT branch of a fork. I wish everyone set up DEFAULT to track the
 # DEFAULT branch in the source repo, but this protects us against sub-optimal
 # setup.
-pr_pull_source_override <- function(tr = NULL, default_branch = NULL) {
-  # naive selection of target repo; calling function should analyse the config
-  tr <- tr %||% target_repo(github_get = FALSE, ask = FALSE)
```
jennybc, line 634: I definitely remember that the only reason I had `NULL` default here and did this was for development convenience.

R/pr.R
```diff
@@ -629,14 +632,10 @@ pr_clean <- function(number = NULL,
...
-
+pr_pull_source_override <- function(tr, default_branch) {
   # TODO: why does this not use a check_*() function, i.e. shared helper?
   # I guess to issue a specific error message?
   current_branch <- git_branch()
-  default_branch <- default_branch %||% git_default_branch()
```
jennybc, line 639: Ditto for this

jennybc (approved):
I don't spot any obvious problems, so LGTM. I think the main way we'll find out there's an issue is just through exercising it.

R/pr.R
hadley: My thinking was that it's worth a little development inconvenience in order to ensure that we get efficient code.

PR #2063, opened 2024-09-20
Specifically handle offline case in `pr_init()`

hadley:
I realised that we're checking whether or not the host is online further down, so we might as well use that a little earlier to give a more informative message.

### DIFF
--- a/R/pr.R
+++ b/R/pr.R
@@ -167,30 +167,41 @@ pr_init <- function(branch) {
   cfg <- github_remote_config(github_get = NA)
   check_for_bad_config(cfg)
   tr <- target_repo(cfg, ask = FALSE)
+  online <- is_online(tr$host)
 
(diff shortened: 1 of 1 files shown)


R/pr.R
```diff
@@ -213,10 +224,6 @@ pr_init <- function(branch) {
       ui_bullets(c("v" = "Pulling changes from {.val {remref}}."))
       git_pull(remref = remref, verbose = FALSE)
     }
-  } else {
-    ui_bullets(c(
```
hadley, line 217: Already warned that we can't do this above

jennybc: We didn't actually warn about *this*, i.e. that we can't make sure the base branch is current. So we could add something around line 175. Or not.

jennybc (approved):

R/pr.R
hadley: Ok, tweaked that message a little bit.

PR #2070, opened 2024-10-08
record upkeep year and use it for filtering checklist

### LINKED ISSUE #1916: Record last upkeep year
In `Config/usethis/upkeep` year? Read/write automatically in `use_tidy_upkeep_issue()`

thomasp85:
Fix #1916 

This PR adds two new functions: `record_upkeep_year()` which writes the year of the upkeep to `Config/usethis/upkeep` and `last_upkeep_year()` which retrieves this value or defaults to 2000 if it doesn't exist. These two functions are used in `use_tidy_upkeep_issue()` to automatically use the year of the last upkeep for filtering the produced checklist as well as incrementing the recorded year to the current year.

Further, this PR fixes what appears to be a bug/discrepancy in how `year` was used in `make_upkeep_issue()` depending on whether `use_upkeep_issue()` or `use_tidy_upkeep_issue()` was called


R/upkeep.R
```diff
@@ -16,13 +16,13 @@
...
 #' \dontrun{
-#' use_upkeep_issue(2023)
+#' use_upkeep_issue()
 #' }
 use_upkeep_issue <- function(year = NULL) {
   make_upkeep_issue(year = year, tidy = FALSE)
 }
 
-make_upkeep_issue <- function(year, tidy) {
+make_upkeep_issue <- function(year, last_year, tidy) {
```
jennybc, line 25: I wonder if it's better to frame this as "previous (upkeep) year"? Seems less vulnerable to confusion about what "last year" actually means. So I'm proposing `previous_year` or `prev_year`.

jennybc: Also, from #1916, do you think we should be incorporating this idea throughout?

> Recording year and month is probably worthwhile.

jennybc (commented):
My first / early reaction is that it feels like we have various `year` arguments lying around, some of which refer to the year of the upkeep and others which refer to the year of the previous upkeep. To be clear, this is not a situation of your creation -- it largely pre-existed.

Do you agree? Do you think this is a good occasion to make that situation less confusing?

Left a couple other comments.

thomasp85:
Yeah, I was a bit perplexed about the duality of `year` when I started digging into this. But I also came to the conclusion that I'd rather not break anything so I kept it as is and instead changed the code to actually do as advertised.

I kinda don't think an argument for "current year" is needed at all - I can't imagine a situation where it would be anything other than what can be extracted from `Sys.Date()`. So if we were to change things up I'd move towards deprecating that argument and perhaps change the name of year were it refers to last upkeep year to something more descriptive

R/upkeep.R
thomasp85: I wondered about this as well. On one hand we are currently not using month to anything (which is we I went the way I did), but on the other hand we may do upkeep biannually which means recording year is not enough.

I'm happy to add month as well and then we can always figure out if/how we want to use that information

thomasp85:
I have changed the code to record the full date rather than just the year. It still holds that we currently have nothing to use the added granularity for, but it is now recorded for the future.

As for the `year` argument, I'm afraid I don't have any strong feelings. The current situation is messy but I'm also hesitant to make a breaking change for the sake of cleaning it up

thomasp85:
I'm happy to deprecate the `year` argument in `use_upkeep_issue()` as I think that is the (most) offending argument, but that is up to you

NEWS.md
```diff
@@ -1,14 +1,18 @@
 # usethis (development version)
 
+* `use_tidy_upkeep_issue()` now records the year it is being run in the
+  `Config/usethis/upkeep` field in DESCRIPTION. If this value exists it is
```
hadley, line 4: Maybe this would be `Config/usethis/last-upkeep`?

R/upkeep.R
```diff
@@ -118,10 +118,12 @@ upkeep_checklist <- function(target_repo = NULL) {
 
 #' @export
 #' @rdname tidyverse
-#' @param year Approximate year when you last touched this package. If `NULL`,
-#'   the default, will give you a full set of actions to perform.
-use_tidy_upkeep_issue <- function(year = NULL) {
-  make_upkeep_issue(year = year, tidy = TRUE)
+#' @param year Approximate year when you last touched this package. Default will
```
hadley, line 121: I suspect few people supply this argument by name, so we could make it less confusing by calling it `last_upkeep`?

tests/testthat/_snaps/upkeep.md
```diff
@@ -60,50 +60,3 @@
       
       <sup>Created on 2023-01-01 with `usethis::use_tidy_upkeep_issue()`, using [usethis v1.1.0](https://usethis.r-lib.org)</sup>
 
-# upkeep bullets don't change accidentally
```
hadley, line 63: Why did these disappear?

hadley (approved):

R/upkeep.R
thomasp85: yeah... also, it dawned on me that everyone is calling these functions interactively (I hope) - changing argument names is basically a non-issue afaik

thomasp85:
@jennybc I think this is ready now. If we merge it in we can ask people to call `usethis:::record_upkeep_date(Sys.Date())` if they have already created the issue

jennybc:
Thanks!

PR #2071, opened 2024-10-08
Support `suggests` field of standalone file

### LINKED ISSUE #2067: Add support for `suggests` field in standalone files
As like #1789, I wonder `suggests` field can be added.

For example, https://github.com/r-lib/rlang/blob/3d48c13d052724cd5997a730ea41bc8e6991691b/R/standalone-cli.R includes `cli::`, so the `cli` package is needed in the `Suggests` filed of the DESCRIPTION file.

eitsupi:
Close #2067


R/use_standalone.R
```diff
@@ -84,20 +87,27 @@ use_standalone <- function(repo_spec, file = NULL, ref = NULL, host = NULL) {
...
+
+    for (i in seq_len(nrow(x))) {
+      pkgs <- x[i, , drop = FALSE]
+      if (is.na(pkgs$ver)) {
+        ver <- NULL
+      } else {
+        ver <- pkgs$ver
+      }
+      ui_silence(
+        use_package(pkgs$pkg, min_version = ver, type = type)
```
olivroy, line 103:
```suggestion
        use_dependency(pkgs$pkg, min_version = ver, type = type, allow_decrease = FALSE)
```

If #2077 gets merged

PR #2077, opened 2024-10-25
Fix `use_data()` + `use_standalone()` by not allowing decreasing R dependency

olivroy:
To fix a drawback from #2043

Basically, if I have a dependency on R 4.1 and call `use_data()`, it will decrease version to 3.5. Same thing if I require rlang 1.1.4 and call `usethis::use_standalone("r-lib/rlang", "types-check")`

In short if I have DESCRIPTION

```
Depends: R (>= 4.1)
```

And I call this:

```r
usethis::use_data(data, overwrite = TRUE)
✔ Setting active project to "usethis".
✔ Decreasing R version to ">= 3.5" in DESCRIPTION.
✔ Saving "data" to "data/data.rda".
```

But this is incorrect.

The PR fixes it.

However, I fixed it in the helper function only. If an external package relied on `use_package()`, it would error.

Agreed that if I call interactively

```r
usethis::use_package("R", "depends", "3.5")
```

it should decrease by default.

### DIFF
--- a/R/data.R
+++ b/R/data.R
@@ -45,9 +45,9 @@ use_data <- function(...,
   objs <- get_objs_from_dots(dots(...))
 
   if (version < 3) {
-    use_dependency("R", "depends", "2.10")
+    use_dependency("R", "depends", "2.10", allow_decrease = FALSE)
   } else {
-    use_dependency("R", "depends", "3.5")
+    use_dependency("R", "depends", "3.5", allow_decrease = FALSE)
   }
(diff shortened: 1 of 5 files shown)


jennybc:
Thanks for catching this! I'm going to solve it in a different way, but am glad you brought it to my attention. I have learned the hard way that `use_dependency()` is very easy to introduce bugs into, so I'll fix this in `use_data()`.

PR #2083, opened 2024-11-20
Never lower an R version requirement in `use_data()`

jennybc:
Closes #2077

### DIFF
--- a/R/data.R
+++ b/R/data.R
@@ -44,11 +44,13 @@ use_data <- function(...,
 
   objs <- get_objs_from_dots(dots(...))
 
-  if (version < 3) {
-    use_dependency("R", "depends", "2.10")
-  } else {
-    use_dependency("R", "depends", "3.5")
+  original_minimum_r_version <- pkg_minimum_r_version()
+  serialization_minimum_r_version <- if (version < 3) "2.10" else "3.5"
+  if (is.na(original_minimum_r_version) ||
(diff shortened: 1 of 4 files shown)


jennybc:
@olivroy Do you remember which standalone file caused the minimum R version to be downgraded for you?

Another option I was considering is a specialized `use_package()` helper for R itself, i.e. splitting that out so that `use_package()` can just focus on normal packages. I'll explore that solution as well.

jennybc:
`use_test()` only calls `use_testthat_impl()` if the package doesn't use testthat yet. So that doesn't seem like it would really contribute to this problem. I'm not saying you didn't experience it with some specific set of moves, but in the main use case, I don't see the connection there.

olivroy:
Ok! simple repro case..

If I have in DESCRIPTION, 


```
Imports:
  rlang (>= 1.1.4)
```

With dev usethis,
I call `usethis::use_standalone("r-lib/rlang", "types-check")`, which [requires rlang 1.1.0](https://github.com/r-lib/rlang/blob/38eb82d2a952fd219ded73a5a2aec894d95505aa/R/standalone-types-check.R#L7).

It will decrease the dependency of rlang silently.

jennybc:
I'm just going to worry about not downgrading the required version of R. I think `use_standalone()` is enough of an expert-use only function that said expert can also decide not to stage that particular change.

Issue #1897, opened 2023-09-22
`github_remote_list()` not picking up GHE host


tsahota:
Hi I'm having trouble using `usethis::use_github_release() on a GHE host`

```r
> usethis::use_github_release()
Error in `stop_bad_github_remote_config()` at usethis/R/utils-github.R:615:4:
! Unsupported GitHub remote configuration: 'no_github'
• Host = NA
• origin = <not configured>
• upstream = <not configured>
• Neither 'origin' nor 'upstream' is a GitHub repo.
```

Here is my remote (note no 'github' string is present in the hostname):

```r
> gert::git_remote_list()
# A tibble: 1 × 2
  name   url                                                  
* <chr>  <chr>                                                
1 origin https://ghe-gsk-prod.metworx.com/account/reponame
```

I believe the source of the issue is in `github_remote_list()` which contains the lines:

```r
parsed <- parse_github_remotes(set_names(x$url, x$name))
# TODO: generalize here for GHE hosts that don't include 'github'
is_github <- grepl("github", parsed$host)
```

and since 'github' doesn't match anything in `https://ghe-gsk-prod.metworx.com/account/reponame`, this causes the issue above.  The TODO comment indicates that the author was aware.

This was introduced in commit 43f6d2dc59096b7b6151bf926b403dcb093a298d and I'd be grateful if this useful function could be made to work on non 'github' matching GHE host names!

jennybc:
Yeah, basically I was waiting to see if anyone actually felt this pain. The simplest thing to do when introducing GHE support was to recognize such a host by looking for "github". I'm not entirely sure what I will do about this, but thanks for opening the issue. It's good to know that someone has banged their shins on this in real life.

In the meantime, have you considered renaming your GHE instance? 😜

tsahota:
OK thanks.  I've opted to make my own little version without the `is_github` test.  But hopefully there can be an easy solution to this in future, then I can get rid of my hack and... usethis

jennybc:
Merging #2098 is a step in the right direction, but it does not solve @tsahota's problem.

Discussion with @jameslairdsmith in #2098 reveals that there are 2 relevant GitHub Enterprise products:

* GitHub Enterprise Cloud (which usethis should now play nicely with)
* GitHub Enterprise Server (which remains unhandled)

The URLs for GHEC are very predictable and can therefore be handled as slight variations on what we do to recognize and process github.com URLs.

[...]

PR #2098, opened 2025-01-31
Handle URLs from GitHub Enterprise managed orgs

jameslairdsmith:
My organisation recently set itself up on a managed GitHub Enterprise organisation. Our new URL is of the form "https://my-cool-org.ghe.com". This PR makes the changes required for usethis to recognise this kind of URL as a GitHub one and parse it to find the appropriate API URL, which is of the form "https://api.my-cool-org.ghe.com".

My immediate use case was trying `usethis::use_pkgdown_github_pages()` and I can confirm that this function now works with the proposed changes.

### DIFF
--- a/R/utils-gh.R
+++ b/R/utils-gh.R
@@ -41,9 +41,17 @@ get_apiurl <- function(url) {
   prot_host <- strsplit(host_url, "://", fixed = TRUE)[[1]]
   if (is_github_dot_com(host_url)) {
     paste0(prot_host[[1]], "://api.github.com")
+  } else if(is_github_enterprise(host_url)) {
+    paste0(prot_host[[1]], "://api.", prot_host[[2]])
   } else {
     paste0(host_url, "/api/v3")
   }
+}
+
+is_github_enterprise <- function(url) {
+  url <- get_baseurl(url)
+  url <- normalize_host(url)
+  grepl("ghe.com$", url)
 }
 
 is_github_dot_com <- function(url) {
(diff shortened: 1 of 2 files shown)


R/utils-gh.R
```diff
@@ -41,11 +41,19 @@ get_apiurl <- function(url) {
...
+    paste0(prot_host[[1]], "://api.", prot_host[[2]])
   } else {
     paste0(host_url, "/api/v3")
   }
 }
 
+is_github_enterprise <- function(url) {
+  url <- get_baseurl(url)
+  url <- normalize_host(url)
+  grepl("ghe.com$", url)
```
jennybc, line 54: Do you have any concrete info on how universally this applies to GHE? Like, is this scheme for the hostname forced on you or is it a default or recommended? I've long wanted to better account for GHE and dragged my feet due to my perception that this could be anything.

jennybc: For example, in #1897, the host is `ghe-gsk-prod.metworx.com`. So I am inclined to believe that full GHE support will need to make this configurable.

jennybc: Thanks for the extra info! So I gather that you have GitHub Enterprise Cloud? It does seems like accomodating that would be considerably easier than full configurability. I'm about to be travelling, but I'll come back and review this PR.

jameslairdsmith: > So I gather that you have GitHub Enterprise Cloud?

That's correct yes.

> I'm about to be travelling, but I'll come back and review this PR.

Safe travels! And thanks for the prompt response!

PR #2105, opened 2025-02-15
docs: link to revdepcheck fn

salim-b:
plus minor tweaks

### DIFF
--- a/R/revdep.R
+++ b/R/revdep.R
@@ -1,11 +1,11 @@
 #' Reverse dependency checks
 #'
-#' Performs set up for checking the reverse dependencies of an R package, as
+#' Performs setup for checking the reverse dependencies of an R package, as
 #' implemented by the revdepcheck package:
-#' * Creates `revdep/` directory and adds it to `.Rbuildignore`
+#' * Creates `revdep/` directory and adds it to `.Rbuildignore`.
 #' * Populates `revdep/.gitignore` to prevent tracking of various revdep
-#' artefacts
-#' * Prompts user to run the checks with `revdepcheck::revdep_check()`
+#' artefacts.
+#' * Prompts the user to run the checks with [revdepcheck::revdep_check()].
 #'
 #' @export
 use_revdep <- function() {


R/revdep.R
```diff
@@ -1,11 +1,11 @@
...
-#' Performs set up for checking the reverse dependencies of an R package, as
+#' Performs setup for checking the reverse dependencies of an R package, as
 #' implemented by the revdepcheck package:
-#' * Creates `revdep/` directory and adds it to `.Rbuildignore`
+#' * Creates `revdep/` directory and adds it to `.Rbuildignore`.
 #' * Populates `revdep/.gitignore` to prevent tracking of various revdep
-#' artefacts
-#' * Prompts user to run the checks with `revdepcheck::revdep_check()`
+#' artefacts.
+#' * Prompts the user to run the checks with [revdepcheck::revdep_check()].
```
jennybc, line 8: I can't link to a topic in a package that's not on CRAN, which revdepcheck is not.

salim-b: Why can't you link to it? Is there a CRAN rule forbidding this?

jennybc: Technically you should have a formal dependency on a package to link to its help topics:

https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Cross_002dreferences-1

> Packages referred to by these ‘other forms’ should be declared in the DESCRIPTION file, in the ‘Depends’, ‘Imports’, ‘Suggests’ or ‘Enhances’ fields.

In the past, you can sort of get away with a link to something you don't depend on, sometimes. But I would not go so far for a package that is not even on CRAN.

Also CRAN is now making static HTML reference manuals, which has triggered a big wave of requests to re-release packages cleaning up various documentation loose ends. So if anything, this is all going to get even stricter.

salim-b: Oh, didn't know that. Thanks for the references and explanations! I guess your reluctance is wise.

In general, I think linking to help topics would always lead to better UX than not linking, regardless of whether a package is installed or not, is on CRAN or not, etc. But of course CRAN comes up with yet another silly obstacle to sensible dev workflows[^1] in 2025... 🥳

[^1]: After all, the R project is one of the last non-SVN related open-source projects that still hasn't switched to Git.

salim-b: Addendum: We could link to the [pkgdown documentation](https://revdepcheck.r-lib.org/reference/revdep_check.html) instead, WDYT?

PR #2136, opened 2025-07-09
use_air() also turns on "format on save" for qmd + other updates

### LINKED ISSUE #2135: `use_air()` should set `formatOnSave` for Quarto too
So that saving formats qmd files as well.

```ts
 "[quarto]": {
    "editor.formatOnSave": true
  },
```

jennybc:
Closes #2135

### DIFF
--- a/R/air.R
+++ b/R/air.R
@@ -13,11 +13,12 @@
 #'   used by the Air extension installed through either VS Code or Positron, see
 #'   the Installation section for more details. Specifically it:
 #'
-#'   - Sets `editor.formatOnSave = true` for R files to enable formatting on
-#'     every save.
+#'   - Sets `editor.formatOnSave = true` for R and Quarto files to enable
+#'     formatting on every save.
 #'
 #'   - Sets `editor.defaultFormatter` to Air for R files to ensure that Air is
-#'     always selected as the formatter for this project.
+#'     always selected as the formatter for this project. Likewise, sets the
+#'     default formatter for Quarto.
 #'
 #'   - Sets the Air extension as a "recommended" extension for this project,
 #'     which triggers a notification for contributors coming to this project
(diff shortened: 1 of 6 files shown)


R/air.R
```diff
@@ -31,20 +32,21 @@
...
 #' additional manual step you must take before using Air for the first time:
 #'
 #' - For RStudio, follow the [installation
 #'   guide](https://posit-dev.github.io/air/editor-rstudio.html).
 #'
-#' - For Positron, install the [OpenVSX
-#'   Extension](https://open-vsx.org/extension/posit/air-vscode).
+#' - For Positron, the [Air extension](https://open-vsx.org/extension/posit/air-vscode)
+#'   is installed by default and that already includes the Air binary. A typical
+#'   Positron user does not need to do anything about installing Air.
```
jennybc: Note: we needed this important update anyway!

DavisVaughan:
```suggestion
#'   Positron user does not need to do anything to install Air.
```

R/air.R
```diff
@@ -118,7 +120,7 @@ create_air_toml <- function(ignore = FALSE) {
 
 air_toml_regex <- function() {
   # Pre-escaped regex allowing both `air.toml` and `.air.toml`
-  "^[\\.]?air\\.toml$"
+  "^[.]?air[.]toml$"
```
DavisVaughan, line 123: I believe the `\\.` was purposeful. i.e. we have `^\.vscode$` in our Rbuildignore along with many other `\.` usage.

Unless the `[]` does something special, I think we should end up with `[\.]` as well, otherwise doesn't a lone `.` mean "anything" or something like that?

R/air.R
```diff
@@ -142,20 +144,19 @@ create_vscode_json_file <- function(name) {
...
-    # Mock it
-    settings_r <- set_names(list())
-  }
-
-  # Set these regardless of their previous values. Assume that calling
-  # `use_air()` is an explicit request to opt in to these settings.
-  settings_r[["editor.formatOnSave"]] <- TRUE
-  settings_r[["editor.defaultFormatter"]] <- "Posit.air-vscode"
-
-  settings[["[r]"]] <- settings_r
```
DavisVaughan, line 158: I was a little worried `modifyList()` would overwrite an existing `[r]` or `[quarto]` setting that _wasn't_ `formatOnSave` or `defaultFormatter`, but I see your new test with

```
    "[quarto]" = list(
      "editor.wordWrap" = "wordWrapColumn"
    )
```

And it looks like it preserves that, so that's a win for `modifyList()` I guess 👍

R/air.R
DavisVaughan: Copilot tells me they are functionally the same, but `\.` is probably more common for escaping a literal `.`, and that seems to align with our existing conventions in `.Rbuildignore`. Thoughts?

Screenshot 2025-07-10 at 8 41 30 AM

DavisVaughan (approved):
Looks good, just a question about the `^[\.]?air\.toml$` change

R/air.R
lionel-: `[]` is commonly used as a less heavy quote syntax for single characters. I like it as it's easier to read.

jennybc: It's perhaps a really small hill to die on, but, yes, I am a big fan of using `[]` instead of escaping, when possible. (But yes, you'll see some escaping in usethis because history.) I really don't care deeply about this.

R/air.R
jennybc: It really is designed for exactly this.

R/air.R
DavisVaughan: The [] is fine

PR #2138, opened 2025-07-10
Deprecate `use_tidy_style()`

### LINKED ISSUE #2110: Rename `use_tidy_style()` to `use_styler()`
Now that we have `use_air()` as well https://github.com/r-lib/usethis/pull/2109

Will probably require a little reorganizing of the file that `use_tidy_style()` lives in, and the pkgdown section it lives in

jennybc:
Closes #2110


R/usethis-deprecated.R
```diff
@@ -56,6 +65,20 @@ use_tidy_eval <- function() {
   )
 }
 
+#' @export
+#' @rdname tidy-deprecated
+use_tidy_style <- function(strict = TRUE) {
+  lifecycle::deprecate_warn(
```
hadley, line 71:
```suggestion
  lifecycle::deprecate_stop(
```

hadley (approved):

R/usethis-deprecated.R
jennybc: Hmm. In theory I'm on board with `deprecate_stop()`. But it felt like going from 0 to 60.

If anyone's got `use_tidy_style()` in teaching materials or personal workflow scripts or similar, it feels like an error versus warning might make a big difference. I went with `deprecate_warning()` since it feels like a state we should pass through.

Do you feel strongly that we go straight to `deprecate_stop()`?

jennybc: Based on a [GitHub search](https://github.com/search?q=use_tidy_style++-repo:cran/usethis+-repo:cran/styler+-repo:r-lib/usethis+-repo:r-lib/styler+-is:fork+&type=code&p=2), I do think we should at least pass through a `deprecate_warn()` phase.

PR #2139, opened 2025-07-11
Add more project criteria

### LINKED ISSUE #2133: Refresh proj detection criteria
https://github.com/r-lib/usethis/blob/8bd4ad5c53062fa0e66d07b6e6b5c37954296f45/R/proj.R#L188-L195

In particular, this should include quarto.

Here are the criteria built-in to rprojroot. I can't remember why usethis doesn't just use those defaults or if there even is a reason for that.

https://rprojroot.r-lib.org/reference/criteria.html

`is_quarto_project` is the main one that really feels like an oversight. Another intriguing candidate: `is_renv_project`.

jennybc:
Closes #2133

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,11 @@
 * `use_package(min_version = FALSE)` is treated the same as when `min_version`
   is not specified (#2117, @salim-b).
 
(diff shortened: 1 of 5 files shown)


hadley (approved):
Do we want to check for `air.toml` too?

PR #2142, opened 2025-07-24
Stop calling `link_path()` and let `is_dir()` do its thing

### LINKED ISSUE #2069: usethis::create_package() fails in temporary directory
```
> usethis::create_package("/tmp/project")
Error in `check_path_is_directory()`:
✖ private/tmp is not a directory.
```

```
> sessionInfo()
R version 4.4.1 (2024-06-14)
Platform: aarch64-apple-darwin20
Running under: macOS 15.0

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/Los_Angeles
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

loaded via a namespace (and not attached):
 [1] compiler_4.4.1  magrittr_2.0.3  cli_3.6.3       tools_4.4.1
 [5] pillar_1.9.0    fs_1.6.4        glue_1.8.0      tibble_3.2.1
 [9] utf8_1.2.4      fansi_1.0.6     vctrs_0.6.5     usethis_3.0.0
[13] lifecycle_1.0.4 pkgconfig_2.0.3 rlang_1.1.4     purrr_1.0.2
```

The problem here occurs because `/tmp` is a symlink to `private/tmp`, but it doesn't get resolved correctly. Here:

https://github.com/r-lib/usethis/blob/b4e84773d1e5e12c60fa2bfc85b19819393fc517/R/directory.R#L42-L44

But:

```
Browse[1]> fs::link_path("/tmp")
private/tmp
```

Note that the link path is relative, not absolute, and so needs to be interpreted relative to the directory holding the symlink.

```
kevin@MBP-P2MQ:/
$ ll | grep tmp
lrwxr-xr-x@  1 root  wheel    11B Sep  5 13:54 tmp@ -> private/tmp
```

jennybc:
Fixes #2069 

@gaborcsardi I tagged you for fs experise. Am I failing to take advantage of anything in fs? Does fs have the equivalent of what `readlink` is doing? Does fs need a function worthy of the name `link_target()`? Hmm, I guess that's just what `path_real()` is?

This guards specifically against a (somewhat peculiar?) situation on macOS.

First, I have this in `.Renviron` to get nicer looking temp file paths. I think this is pretty common among R users or R power users:

```
TMPDIR=/tmp
```

`/tmp` is actually a symlink on macOS, but it's a *relative* symlink:

```
~/rrr/usethis % ls -l /tmp
lrwxr-xr-x@ 1 root  wheel  11 May  3 22:39 /tmp -> private/tmp
```

So it has to be interpreted relative to this directory it lives in, namely `/`.

If you use `readlink` instead, you get the full resolution:

```
~/rrr/usethis % readlink -f /tmp
/private/tmp
```

I suppose I have always assumed `fs::link_path()` was returning output akin to `readlink`, whereas it's actually doing something more like `ls -l`. I guess I have not been punished for this because relative symlinks are rare?

In any case, I think this is a correct and minimal fix to make `check_path_is_directory()` more robust. My tendency towards minimalism comes from previous painful experience proving that all path manipulation in usethis is more tricky than it seems, mostly thanks to Windows (UNC paths vs drive paths, etc.).

### DIFF
--- a/R/directory.R
+++ b/R/directory.R
@@ -39,7 +39,7 @@ check_path_is_directory <- function(path) {
   }
 
   if (is_link(path)) {
-    path <- link_path(path)
+    path <- path_real(path)
   }
 
   if (!is_dir(path)) {
--- a/R/directory.R
+++ b/R/directory.R
@@ -38,12 +38,14 @@ check_path_is_directory <- function(path) {
     ui_abort("Directory {.path {pth(path)}} does not exist.")
   }
 
+  orig_path <- path
+
   if (is_link(path)) {
     path <- path_real(path)
   }
 
   if (!is_dir(path)) {
-    ui_abort("{.path {pth(path)}} is not a directory.")
+    ui_abort("{.path {pth(orig_path)}} is not a directory.")
   }
 }
 
--- a/R/directory.R
+++ b/R/directory.R
@@ -38,14 +38,8 @@ check_path_is_directory <- function(path) {
     ui_abort("Directory {.path {pth(path)}} does not exist.")
   }
 
-  orig_path <- path
-
-  if (is_link(path)) {
-    path <- path_real(path)
-  }
-
   if (!is_dir(path)) {
-    ui_abort("{.path {pth(orig_path)}} is not a directory.")
+    ui_abort("{.path {pth(path)}} is not a directory.")
   }
 }
 
--- a/tests/testthat/test-directory.R
+++ b/tests/testthat/test-directory.R
@@ -29,11 +29,22 @@ test_that("symlink to directory is directory", {
 
 # https://github.com/r-lib/usethis/issues/2069
 test_that("relative symlink to directory is directory", {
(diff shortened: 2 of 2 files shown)


gaborcsardi:
Yes, this makes sense to me. Maybe keep the original `path` argument and use that in the error message?

jennybc:
Ha! Upon further research, the real solution was to just delete the call to `link_path()` entirely. `is_dir()` handles the symlink case by default now.

PR #2144, opened 2025-07-25
Use finalized `format-suggest.yaml`

DavisVaughan:
Hadley convinced me it's not worth it to pin the shas. Everything else stayed the same.


jennybc (approved):
YOLO

Issue #1946, opened 2024-02-14
`use_release_issue()` changes


hadley:
Based on our discussion at work week, I'd propose that we too new arguments/questions to `use_release_issue()`: `use_pre_release` and `use_branch`.

If `use_pre_release` is true, then we create two issues a pre-release (the current "prepare for release bullets), and then a release issue, which would gain new bullets for locking the main branch at the start and unlocking at the end.

If `use_branch` is true, the bullets would replace `git pull` and `git push` with (e.g.) `pr_init("v1.4.1-release")` and merging the PR. (We'd recommend that the PR close the issue automatically with a comment).

(For `use_branch = TRUE` we still need to think through the post-release process to ensure that both dev and releases sites are updated, and we end up with a dev version on the main branch.)

hadley:
A couple more thoughts:

* We probably want to keep `use_prerelease = FALSE` the default, since it only really affects bigger releases.
* We might want to make `use_branch = TRUE` the default, since it's probably good practice for us even we can push directly to main.
* We might want to move the check runs to the release process, because particularly when using a branch, we can update the version earlier, meaning that we'd move to 0 notes when everything was good.

jennybc:
We took an initial step on this by adding `github_(un)lock_branch()` but then I had to revert those changes. Due to the API design, it's going to be pretty fiddly to lock/unlock a branch, without nuking all pre-existing branch protections or leaving it with new, undesired branch protections. I didn't want to put the half-baked version in a patch release and I don't want to do the job right at this time. I'll leave some notes below for our next attempt.

---

*I'll be back with notes.*

hadley:
Claude suggests we might be able to use (new-since-last-time-we-tried-this?) rulesets API

```R
gh("POST /repos/{owner}/{repo}/rulesets",
   owner = "tidyverse",
   repo = "purrr",
   name = "Freeze main",
   target = "branch",
   enforcement = "active",
   conditions = list(
     ref_name = list(include = list("refs/heads/main"), exclude = list())
   ),
   rules = list(
     list(type = "deletion"),
     list(type = "non_fast_forward"),
     list(type = "update")  # blocks all updates = read-only
   )
)```

And

> Rulesets are nicer for temporary freezes because you can flip enforcement between "active", "evaluate", and "disabled" without deleting and recreating, and you get an explicit ID to manage.

PR #2145, opened 2025-07-28
Lock and unlock branch as part of the release process

hadley:
This ensures that there's no way to accidentally forget that you're in the middle of the release process and make a change to the repo.

Part of #1946

@DavisVaughan it's been a long time since we discussed this, but does this make sense to you?

@jennybc mostly opening to start discussion; not planning to merge just yet.


R/github_token.R
```diff
@@ -294,3 +296,48 @@ scold_for_scopes <- function(scopes) {
   )
   ui_bullets(message)
 }
+
+#' Lock and unlock a branch on GitHub
+#'
+#' These functions lock and unlock a branch on GitHub so that it's not possible
+#' for anyone to make any changes. This is used as part of the release process
+#' to ensure that you don't accidentally merge any pull requests or push any
+#' commits while you are waiting for CRAN to get back to you.
```
DavisVaughan: Anything about permissions required to perform this action?

R/github_token.R
```diff
@@ -294,3 +296,48 @@ scold_for_scopes <- function(scopes) {
...
+#'
+#' @export
+#' @param branch The branch to lock/unlock. If not supplied, uses the
+#'   default branch with is usually "main" or "master".
```
DavisVaughan:
```suggestion
#'   default branch which is usually "main" or "master".
```

R/github_token.R
```diff
@@ -294,3 +296,48 @@ scold_for_scopes <- function(scopes) {
...
+gh_lock_branch <- function(branch = NULL) {
+  cfg <- github_remote_config(github_get = TRUE)
+  repo <- target_repo(cfg)
+  branch <- branch %||% git_default_branch_(cfg)
+
+  invisible(gh::gh(
```
DavisVaughan: What happens if you lock an already locked branch?

R/release.R
```diff
@@ -137,13 +138,15 @@ release_checklist <- function(version, on_cran, target_repo = NULL) {
     "",
     "Submit to CRAN:",
     "",
+    todo("`usethis::gh_lock_branch()"),
```
DavisVaughan:
```suggestion
    todo("`usethis::gh_lock_branch()`"),
```

R/release.R
```diff
@@ -137,13 +138,15 @@ release_checklist <- function(version, on_cran, target_repo = NULL) {
...
     todo("`usethis::use_version('{type}')`"),
     todo("`devtools::submit_cran()`"),
     todo("Approve email"),
     "",
     "Wait for CRAN...",
     "",
     todo("Accepted :tada:"),
+    todo("`usethis::gh_unlock_branch()"),
```
DavisVaughan:
```suggestion
    todo("`usethis::gh_unlock_branch()`"),
```

DavisVaughan (approved):
I think the positioning makes sense based on what I remember about this

thomasp85:
Don't have anything to add on top of Davis' comments

R/github_token.R
hadley: Looks like it's idempotent

R/github_token.R
hadley: You need to be an admin or owner, which I've documented. It looks like this is already in line with our governance tidyup where the maintainer has admin access: https://github.com/tidyverse/tidyups/blob/main/004-governance.md#maintainer.

NAMESPACE
```diff
@@ -32,7 +32,9 @@ export(edit_r_profile)
 export(edit_rstudio_prefs)
 export(edit_rstudio_snippets)
 export(edit_template)
+export(gh_lock_branch)
```
jennybc: What do we think of `github_lock_branch()`? Just want us to at least consider. I don't have a strong opinion.

R/github_token.R
```diff
@@ -296,3 +296,52 @@ scold_for_scopes <- function(scopes) {
...
+#'
+#' You must be an admin or an owner of the repo in order to lock/unlock
+#' a branch.
+#'
+#' @export
+#' @param branch The branch to lock/unlock. If not supplied, uses the
+#'   default branch which is usually "main" or "master".
+gh_lock_branch <- function(branch = NULL) {
+  cfg <- github_remote_config(github_get = TRUE)
+  repo <- target_repo(cfg)
```
jennybc: Right around here we could adapt this code we use in `use_release_issue()` and `use_upkeep_issue()` except I guess we just tell people that don't have enough permission to do this:

https://github.com/r-lib/usethis/blob/05920773ba2ab5bd85ade54d596fd6b55975834b/R/release.R#L38-L47

R/github_token.R
```diff
@@ -296,3 +296,52 @@ scold_for_scopes <- function(scopes) {
...
+  branch <- branch %||% git_default_branch_(cfg)
+
+  invisible(gh::gh(
+    "PUT /repos/{owner}/{repo}/branches/{branch}/protection",
+    owner = repo$repo_owner,
+    repo = repo$repo_name,
```
jennybc: We usually do a different move to create these `gh::gh()` calls for the target repo, but I am happy to do this consistency pass myself before we merge.

NAMESPACE
hadley: I don't have a strong opinion either. If you want user-facing functions to use a `github_` prefix I'm happy to change it.

R/github_token.R
hadley: I think it's probably fine to let the GitHub error bubble up, and then we can add more user-friendly errors if we discover it's causing substantial confusion.

R/github_token.R
hadley: Thanks!

R/github_token.R
jennybc: I went ahead and emitted our own error. It's a big improvement on what happens otherwise:

Letting the gh error bubble up, we get a mysterious 404:
```
> github_lock_branch()
✔ Setting active project to "/Users/jenny/tmp/nettskjemar".
Error in `gh::gh()` at usethis/R/utils-gh.R:6:5:
! GitHub API error (404): Not Found
✖ URL not found:
  <https://api.github.com/repos/cran/nettskjemar/branches/master/protection>
ℹ Read more at
  <https://docs.github.com/rest/branches/branch-protection#update-branch-protection>
```

If we check for admin permissions and message:

```
> github_lock_branch()
✔ Setting active project to "/Users/jenny/tmp/nettskjemar".
Error in `github_lock_branch()`:
✖ You don't seem to have admin permissions for the source repo
  "cran/nettskjemar", which is required to lock a branch.
```

jennybc:
The CI failure is because the Air `format-suggest` workflow is having a bad day. Not relevant.

PR #2147, opened 2025-07-28
Introduce mockable `testthat_version()`

hadley:
So the tests don't fail if you happen to be working on the development version of testthat.


R/test.R
```diff
@@ -95,6 +89,9 @@ check_edition <- function(edition = NULL) {
   }
 }
 
+testthat_version <- function() {
```
jennybc, line 93:
```suggestion
# wrapping so we can mock this in tests
testthat_version <- function() {
```

We've got a few other instances of this, all labelled like so, so we don't accidentally remove it later.

jennybc (approved):

Issue #2137, opened 2025-07-10
Assist with GHA setup around Air


jennybc:
What do we want to do around GitHub Actions and Air?

<https://posit-dev.github.io/air/integration-github-actions.html>

Two workflows already exist and can be placed like so:

```r
usethis::use_github_action(url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-suggest.yaml")
usethis::use_github_action(url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-check.yaml")
```

* `format-suggest.yaml`: "This action runs `air format .` on every pull request. If formatting is required, the check fails and suggestion comments are added directly to the pull request."
* `format-check.yaml`: "This runs `air format . --check` on every push to `main` and on every pull request. This is a very simple action that fails if any files would be reformatted."

I don't think we want to place these with `use_air()`, since that is usually just the _opening move_ for adopting Air. You would not want these actions in place until the reformatting was complete. 

Discussion points:

* [ ] We could message about adding GHA in `use_air()` and could give the code snippet above re: how to place the actions later in the `use_air()` docs.
* [ ] I feel like there are some implications for `r-lib/actions`. For example, it certainly feels nicer/easier to place workflows that live there. Would we want to copy the Air actions to `r-lib/actions`? Also, what are the implications for `/style` as currently defined [`examples/pr-commands.yaml`](https://github.com/r-lib/actions/blob/6b49fefb2846ed3e5e6e58366e7f7dfe01377f02/examples/pr-commands.yaml#L69). Note that we currently place/update this workflow in `use_tidy_github_actions()` and it's one of the few we offer in `use_github_action()`, so it's sort of "official".

DavisVaughan:
It is worth nothing that I very much dislike `/style` as a PR comment command for this reason

> Note that if you combine this with another action to commit and push these reformatted files back to a branch or pull request, the push will [not retrigger](https://github.com/orgs/community/discussions/25702) any of your GitHub Workflows. This means you won't get any feedback that the push has fixed any broken formatting. Instead, we recommend that you either reformat locally and push the changed files back up using a standard commit, or use the format-suggest.yaml example that adds suggestion comments to your pull request. Both of these methods will retrigger your checks.

Not getting a ✅ on your R CMD check workflow, or your style workflow, after running the formatter and pushing back results is _scary business_ and I dislike it.

[...]

PR #2150, opened 2025-07-30
Pivot away from pr commands

jennybc:
Progress towards #2137 

I think this is about as far as we can go at this moment.

### DIFF
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -249,7 +249,6 @@ use_tidy_github_actions <- function(ref = NULL) {
 
   use_github_action("check-full.yaml", ref = ref, badge = TRUE)
 
-  use_github_action("pr-commands", ref = ref)
   use_github_action("pkgdown", ref = ref)
 
   use_coverage(repo_spec = repo_spec)
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -233,9 +233,8 @@ use_github_actions_badge <- function(
 #'     and Windows).
 #'   - Report test coverage.
 #'   - Build and deploy a pkgdown site.
-#'   - Provide two commands to be used in pull requests: `/document` to run
-#'     `roxygen2::roxygenise()` and update the PR, and `/style` to run
-#'     `styler::style_pkg()` and update the PR.
+#'   - Check the formatting of incoming pull requests with Air and suggest
+#'     fixes as necessary.
 #'
 #'     This is how the tidyverse team checks its packages, but it is overkill
 #'     for less widely used packages. For `R CMD check`, consider using the more
@@ -254,6 +253,12 @@ use_tidy_github_actions <- function(ref = NULL) {
   use_coverage(repo_spec = repo_spec)
   use_github_action("test-coverage", ref = ref)
 
+  use_github_action(
+    url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-suggest.yaml"
+  )
+
+  # TODO: give `pr-commands` similar treatment once we have a full replacement,
+  # i.e. the aspirational `document-suggest`
   old_configs <- proj_path(c(".travis.yml", "appveyor.yml"))
   has_appveyor_travis <- file_exists(old_configs)
 
--- a/R/github-actions.R
+++ b/R/github-actions.R
@@ -2,13 +2,13 @@
 #'
 #' @description
 #' Sets up continuous integration (CI) for an R package that is developed on
-#' GitHub using [GitHub Actions](https://github.com/features/actions). CI can be
-#' used to trigger various operations for each push or pull request, e.g.
+#' GitHub using [GitHub Actions](https://github.com/features/actions) (GHA). CI
+#' can be used to trigger various operations for each push or pull request, e.g.
 #' running `R CMD check` or building and deploying a pkgdown site.
 #'
-#' ## Workflows
+#' ## Core workflows
 #'
-#' There are four particularly important workflows that are used by many
+#' There are three particularly important workflows that are used by many
 #' packages:
 #'
 #' * `check-standard`: Run `R CMD check` using R-latest on Linux, Mac, and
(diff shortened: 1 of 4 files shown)


R/github-actions.R
```diff
@@ -150,7 +171,8 @@ choose_gha_workflow <- function(error_call = caller_env()) {
   workflows <- c(
     "check-standard" = "Run `R CMD check` on Linux, macOS, and Windows",
     "test-coverage" = "Compute test coverage and report to https://about.codecov.io",
-    "pr-commands" = "Add /document and /style commands for pull requests"
+    "posit-dev/setup-air/format-suggest" = "Suggest fixes to code formatting in pull requests",
+    "posit-dev/setup-air/format-check" = "Checks code formatting with a pass/fail mentality"
```
jennybc: I decided to promote these to this menu. How do we feel about the way I've "namespaced" them, to make their provenance more clear?

R/github-actions.R
```diff
@@ -227,12 +248,17 @@ use_tidy_github_actions <- function(ref = NULL) {
...
 
   use_coverage(repo_spec = repo_spec)
   use_github_action("test-coverage", ref = ref)
 
+  use_github_action(
+    url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-suggest.yaml"
+  )
+
+  # TODO: give `pr-commands` similar treatment once we have a full replacement,
+  # i.e. the aspirational `document-suggest`
```
jennybc, line 265: I don't think `use_tidy_github_actions()` should proactively remove `pr-commands` just yet, but eventually we will.

R/github-actions.R
```diff
@@ -17,25 +17,45 @@
...
+#' ## Other workflows
+#' Other specific workflows are worth mentioning:
+#' * `pr-commands`: `r lifecycle::badge("superseded")` Enables the use of two
+#'    R-specific commands in pull request issue comments: `/document` to run
+#'    `roxygen2::roxygenise()` and `/style` to run `styler::style_pkg()`. Both
+#'    will update the PR with any changes once they're done.
+#'
+#'    We don't recommend new adoption of the `pr-commands` workflow. Better
+#'    alternatives exist for code formatting (see below) and we plan to
+#'    re-implement documentation updates using a similar approach.
```
DavisVaughan: Move to the bottom of the list? and remove `(see below)`

R/github-actions.R
```diff
@@ -17,25 +17,45 @@
...
+#' * `format-suggest` and `format-check` from
+#'   [Air](https://posit-dev.github.io/air):
+#'   `r lifecycle::badge("experimental")` These workflows are a great way to
+#'   keep your code well-formatted, once you adopt Air in a project
```
DavisVaughan:
```suggestion
#'   keep your code well-formatted once you adopt Air in a project
```

R/github-actions.R
```diff
@@ -17,25 +17,45 @@
...
+#' Other specific workflows are worth mentioning:
+#' * `pr-commands`: `r lifecycle::badge("superseded")` Enables the use of two
+#'    R-specific commands in pull request issue comments: `/document` to run
+#'    `roxygen2::roxygenise()` and `/style` to run `styler::style_pkg()`. Both
+#'    will update the PR with any changes once they're done.
+#'
+#'    We don't recommend new adoption of the `pr-commands` workflow. Better
+#'    alternatives exist for code formatting (see below) and we plan to
+#'    re-implement documentation updates using a similar approach.
+#' * `format-suggest` and `format-check` from
```
DavisVaughan:
```suggestion
#' * `format-suggest` or `format-check` from
```

I view them more as an either or situation, depending on your preferences

R/github-actions.R
DavisVaughan: To be clear this doesn't actually work right? Because the url you build from this snippet points to r-lib/actions

DavisVaughan: I'm actually ok with not putting them in the menu if it makes things less complicated

R/github-actions.R
```diff
@@ -227,12 +248,17 @@ use_tidy_github_actions <- function(ref = NULL) {
...
 
-  use_github_action("pr-commands", ref = ref)
   use_github_action("pkgdown", ref = ref)
 
   use_coverage(repo_spec = repo_spec)
   use_github_action("test-coverage", ref = ref)
 
+  use_github_action(
+    url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-suggest.yaml"
+  )
```
DavisVaughan, line 262: Random thought that possibly it should be guarded with `if (uses_air()) {`? Which would look for an `air.toml` at root level?

I'm not sure, I think what is here is also fine as it encourages you to opt in to air with `use_air()` as well.

R/github-actions.R
jennybc: Oh, right, I forgot how hard the assumption of `r-lib/actions` is here. I'll either fix that up or remove. A menu with just 2 options feels sort of odd, though 🤔

jennybc: I've removed them. In addition to targetting the correct repo, there's also the matter of getting the right ref, which is "latest release" for those in `r-lib/actions`, and getting the associated README.

I still think it would be nice if `format-suggest` and `document-suggest` showed up here eventually, but we'll just have to let things develop further and revisit this.

R/github-actions.R
jennybc: Yeah I'll do this.

R/air.R
```diff
@@ -89,7 +87,9 @@ use_air <- function(vscode = TRUE) {
 
   ui_bullets(c(
     "_" = "Read the {.href [Air editors guide](https://posit-dev.github.io/air/editors.html)}
```
DavisVaughan:
```suggestion
    "_" = "Read {.href [Air's editors guide](https://posit-dev.github.io/air/editors.html)}
```

For symmetry

R/github-actions.R
DavisVaughan: Also look for `.air.toml` with a leading `.`

jennybc: I've added this now. It sort of gets lost in the other output about writing files, but I think it's still good to include.

```
> use_tidy_github_actions()
✔ Saving "r-lib/actions/examples/check-full.yaml@v2" to .github/workflows/R-CMD-check.yaml.
☐ Learn more at <https://github.com/r-lib/actions/blob/v2/examples/README.md>.
✔ Adding "R-CMD-check badge" to README.md.
✔ Adding "Codecov test coverage badge" to README.md.
✔ Saving "r-lib/actions/examples/pkgdown.yaml@v2" to .github/workflows/pkgdown.yaml.
☐ Learn more at <https://github.com/r-lib/actions/blob/v2/examples/README.md>.
✔ Saving "r-lib/actions/examples/test-coverage.yaml@v2" to .github/workflows/test-coverage.yaml.
☐ Learn more at <https://github.com/r-lib/actions/blob/v2/examples/README.md>.
! Can't find an air.toml file. Do you need to run use_air()?
✔ Saving "posit-dev/setup-air/examples/format-suggest.yaml@main" to
  .github/workflows/format-suggest.yaml.
```

jennybc: In a setting where such hyperlinks work, `use_air()` is clickable/runnable here:

> ! Can't find an air.toml file. Do you need to run use_air()?

R/github-actions.R
```diff
@@ -251,6 +251,12 @@ use_tidy_github_actions <- function(ref = NULL) {
...
 
+  if (!uses_air()) {
+    ui_bullets(c(
+      "!" = "Can't find an {.file air.toml} file. Do you need to run
+             {.run [use_air()](usethis::use_air())}?"
+    ))
+  }
   use_github_action(
     url = "https://github.com/posit-dev/setup-air/blob/main/examples/format-suggest.yaml"
   )
```
DavisVaughan: Oh interesting I was originally suggesting that we'd only ever add the `format-suggest.yaml` if we detected usage of air, like

```r
if (uses_air()) {
  use_github_action(".../format-suggest.yaml")
}
```

I don't mind being a little pushy about using air though

jennybc: I think it's ok to be pushy in `use_tidy_github_actions()`, since the only folks who are pseudo-obligated to do this is us.

PR #2151, opened 2025-07-31
Assume the folder from use_course() should be treated as a project

### LINKED ISSUE #2127: `use_course()` should open projects in Positron without an `.Rproj` file
`use_course()` relies on the presence of an `.Rproj` file to trigger opening the project via `rstudioapi::openProject()`:

https://github.com/r-lib/usethis/blob/4aa55e72ccca131df2d98fcd84fff66724d6250a/R/course.R#L398-L409

For a project designed to be used in Positron, this file may not be present, but it would be nice if the project still opened in a new Positron window.

jennybc:
Fixes #2127

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,10 +11,16 @@
 
 * `use_r_universe_badge()` is a new function that creates a README badge indicating your package is available on [R-universe](https://r-universe.dev) and reporting the latest version (@olivroy, #1883).
 
-* usethis's criteria for recognizing a project have expanded to include (#2133):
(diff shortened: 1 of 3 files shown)


jennybc:
@cwickham I'm merging this and will start a usethis release process very soon. But if you have a chance to take `use_course()` out for a spin, please do and let me know if you have any feedback!

`use_course()` should open the new folder in the current IDE (Positron or RStudio) essentially all the time now. And will also write a `.here` file if none of the other project root criteria are satisfied.

PR #2165, opened 2025-09-04
Verifying fixes for "no Quarto CLI"

jennybc:
I've adapted a GHA workflow from the quarto R package. I want to reproduce the CRAN failure in this branch, then apply recent changes from `main` and see that it's fixed. I will never merge this.


jennybc:
I'm convinced!

Before:

Screenshot 2025-09-04 at 9 40 27 AM

After:
Screenshot 2025-09-04 at 9 41 30 AM

PR #2173, opened 2025-09-19
Allow `use_github_file()` to work with non-text files

### LINKED ISSUE #1782: Make `use_github_file()` more general
As it stands, we've built-in the assumption that we're dealing with a plain text file, but I don't see why it should be limited to that.

ateucher:
Since text files are read with `read_utf8()` and written to the final destination with `write_over()`, I changed `read_github_file()` to `get_github_file()` and instead of returning the text lines, it returns a path to a temp file, which requires using `read_utf8()` in places it was used before.

Fixes #1782.

### DIFF
--- a/tests/testthat/test-use_github_file.R
+++ b/tests/testthat/test-use_github_file.R
@@ -59,3 +59,13 @@ test_that("parse_file_url() errors when it should", {
     "https://gitlab.com/OWNER/REPO/path/to/file"
   ))
 })
+
+test_that("use_github_file works with non-text files", {
+  create_local_project()
+  use_github_file(
+    "https://github.com/r-lib/usethis/blob/main/man/figures/logo.png",
+    save_as = "logo.png"
+  )
+
+  expect_proj_file("logo.png")
+})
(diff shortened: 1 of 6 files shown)


tests/testthat/test-use_github_file.R
```diff
@@ -59,3 +59,13 @@ test_that("parse_file_url() errors when it should", {
     "https://gitlab.com/OWNER/REPO/path/to/file"
   ))
 })
+
+test_that("use_github_file works with non-text files", {
+  create_local_project()
+  use_github_file(
+    "https://github.com/r-lib/usethis/blob/main/man/figures/logo.png",
```
ateucher, line 98: I think this is probably stable, but I can put a small png in `tests/testthat/` if that would be better?

ateucher:
Hi @jennybc, I got Posit Assitant to help me clean this up and make it more robust, and also avoid adding `mime` as a new dependency.

-   Check file content, not extension to find binaries: `is_binary_file()` checks for NUL-bytes (same as `grep -I`/git). This fixes YAML/JSON files (`application/*` MIME types) and extensionless files being misclassified as binary, and drops the `mime` dependency.
-   Fixed `get_release_news()`: Moved `read_utf8()` inside the `tryCatch`.
-   Binary files now get same treatment overwrite as text in `write_file()`, used `can_overwrite()`.
-   Added tests for the detection helper and binary overwrite behavior, a NEWS bullet, and `skip_if_offline()` on the download test.

PR #2174, opened 2025-09-19
Add the use-upkeep_issue checklist to the manual page #1862

### LINKED ISSUE #1862: Could the manual page of use_upkeep_issue show the checklist?
I'd find it nice to see the list of items in the function documentation, without having to open an issue. My use case would be to send the page link to potential users, being confident they'd get a good idea of what the good practice list could look like.

In any case, thanks for a very useful function. :smile_cat:

ndjrt:
Fixes #1862


hadley:
Closing since it's now being tackled in #2279.

PR #2180, opened 2025-09-19
Add `use_snapshot()`

### LINKED ISSUE #2156: How to browse snapshots from a test file?
(Maybe rather an IDE question)

If I'm reading tests to gain an understanding of what the code is doing, how can I easily access the snapshots? Currently I do that by clicking.

More often of course I interact with snapshots when they fail or when I add some. :sweat_smile:

jonthegeek:
Implement a `use_snapshot()` helper to go with `use_r()` and `use_test()`.

I ran out of time and didn't implement tests or NEWS for this yet (and there's a solid chance that this doesn't quite nail it), but I wanted to get it checked in.

Fixes #2156.


jonthegeek:
This should be good to go now!

R/r.R
```diff
@@ -88,6 +88,39 @@ use_test <- function(name = NULL, open = rlang::is_interactive()) {
...
+    use_testthat_impl()
+  }
+
+  path_root <- proj_path("tests", "testthat", "_snaps")
+  if (!is.null(variant)) {
+    path_root <- path(path_root, variant)
+  }
+  # I can't pass "md" to `compute_name()` because we're fine with them giving us
+  # "R" as the extension here.
+  path <- path(path_root, fs::path_ext_set(compute_name(name), "md"))
```
hadley: It seems weird to not follow the same pattern as all the other `compute_name()` using functions.

R/r.R
```diff
@@ -88,6 +88,39 @@ use_test <- function(name = NULL, open = rlang::is_interactive()) {
   invisible(TRUE)
 }
 
+#' Edit the snapshot file associated with an R or test file
+#'
+#' This function opens the snapshot file associated with an R or test file, if
+#' one exists.
+#'
+#' @inheritParams use_r
+#' @param variant An optional subdirectory within the `_snaps/` directory.
```
hadley: I think you can drop the variant argument; it's not important for this use case.

R/r.R
jonthegeek: I over-complicated. This fails:

``` r
usethis:::compute_name("x.R", ext = "md")
#> Error:
#> ! `name` must have extension "md", not "R".
```

<sup>Created on 2025-09-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

...but so does this:
``` r
usethis:::compute_name("x.md", ext = "R")
#> Error:
#> ! `name` must have extension "R", not "md".
```

<sup>Created on 2025-09-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Because what actually gets sent to `compute_name()` in the context we're dealing with is usually a bare name, or (even more often) `NULL`, in which case the "What file is open?" code kicks in. So... yup, I can simplify this to the same as everywhere else, with the `ext` arg. Fixing!

R/r.R
jonthegeek: Fair enough!

jennybc:
@jonthegeek and/or @maelle What is the workflow in which you regularly want to open snapshot files in an editor? I mostly interact with them in the `testthat::snapshot_review()` app or in my git client. On the occasion I want to open a snapshot file, I think I just fall back to whatever usual ways I open files in, e.g., my IDE. It feels like there's some workflow context here that I'm missing.

jonthegeek:
> @jonthegeek and/or @maelle What is the workflow in which you regularly want to open snapshot files in an editor?

I think my original interest in this was more for the other direction, where I'm looking at a snapshot (on a rare occasion when I edited one directly for an anticipated result or something) and want to hop to the test or R file with my hotkeys. I've had situations where I want to edit the file to define what I think it should look like, though (before working out exactly how to do that), so that's where this would come in for me. Not super common but it will be nice when it comes up.

jonthegeek:
@jennybc Merge conflicts resolved. Should I re-mark this as "done" as far as tidy-dev-day?

hadley:
Thank you!

maelle:
:tada: :pray:

PR #2186, opened 2025-09-29
`create_from_github()` tweaks

hadley:
* Don't need to force RStudio project creation if in positron
* Make it easier to install the dev dependencies

The second bullet is the most important here — I realised that 100% of the time I call `create_from_github()` the next function I call is `pak::pak()` in order to be able to actually run the tests. I think this is probably generally true, and installing a few extra packages shouldn't be harmful.


hadley:
I'm finding the auto-install super helpful as I work through the testthat revdeps.

DavisVaughan (approved):
Both of these feel like wins to me for normal day to day usage of this function

I have definitely gotten in the habit of having to clear out RStudio specific info when I use `create_from_github()` from Positron on a non-RStudio project, so that feels like a win.

Notably, I do not think I would have used `install_dependencies` for my Claude Code adventures with dplyr revdeps. I specifically wanted to control the libpath where the dependencies were installed to, so I had to:
- `create_from_github()` to make `pkg/`
- Add `pkg/.Rprofile` which added a custom path to `.libPaths()`
- Then `pak::pak()`

So I would have needed to inject a step in the middle there.

But that's a rare use case, and I would still have been able to achieve this because I would have been able to just set `install_dependencies = FALSE`.

R/create.R
```diff
@@ -388,7 +392,12 @@ create_from_github <- function(
     )
   }
 
-  rstudio <- rstudio %||% rstudio_available()
+  if (install_dependencies) {
+    ui_bullets(c("v" = "Installing missing dependencies."))
+    pak::local_install_dev_deps(repo_path, upgrade = FALSE)
+  }
+
+  rstudio <- rstudio %||% (rstudio_available() && !is_positron())
```
jennybc, line 400: Just to put this down in writing: in some workflows for "make this new thing", leaving `.Rproj` behind is still somewhat load-bearing in terms of advertising a project root and there's no Positron equivalent. And it doesn't hurt anything.

But we don't have to fret over this here, because, by definition, if you're using `create_from_github()`, we're talking about a git repo and *that* is a strong signal of the project root. And obviously it's nicer to not have a vestigial, unnecessary `.Rproj` file.

jennybc (approved):
LGTM!

Vague thought: do we want to describe or qualify these to-be-installed dependencies as "development dependencies" anywhere? Not a strong opinion, it's just occurred to me since I definitely think of  them this way. Maybe because the function we use is `pak::local_install_dev_deps()` and not `pak::local_install_deps()`.

R/create.R
```diff
@@ -237,6 +238,8 @@ create_quarto_project <- function(
 #'   pre-existing `.Rproj` file. Defaults to `FALSE` otherwise (but note that
 #'   the cloned repo may already be an RStudio Project, i.e. may already have a
 #'   `.Rproj` file).
+#' @param install_dependencies Install package dependencies? Defaults to `TRUE`,
+#'   so that you can immediately work with the package.
```
jennybc:
```suggestion
#'   so that you can immediately do development work on the package.
```

PR #2210, opened 2026-02-02
Don't auto-format quarto docs

hadley:
Because it doesn't work correctly.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,7 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(Rscript:*)"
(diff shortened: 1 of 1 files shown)


.claude/settings.local.json
jennybc, line 1: Do we want to track a *local* claude settings file here? I would have expected `.claude/settingsjson`.

jennybc (approved):
I'm happy to go along with this, but do hope we can bring formatting inside `.qmd` back.

.claude/settings.local.json
hadley: Gah I didn't mean the commit this.

PR #2211, opened 2026-02-12
add tidy principled `expect_` usage to CLAUDE.md

EmilHvitfeldt:
I found that using this CLAUDE.md files gives me a lot of `expect_true()` and almost no `expect_snapshot()`. This is my attempt to fix this.

Done by going back and forth

- generate new tests 
- look at and verify old tests
- update CLAUDE.md
- delete changes

repeat.

The language feels quite hard, but without it, it doesn't pick it up. Example below:

```md
# Me
you still generated some expect_error and expect_warning dispite being told not to in
.claude/CLAUDE.md. what went wrong?                                                 

# Claude
You're right. The CLAUDE.md instructions say:                                           
                                                                                          
Use expect_snapshot(error = TRUE) to test error messages (and similarly for warnings and
other conditions).                                                                     
                                                                                          
I should have used expect_snapshot(error = TRUE) instead of expect_error() and
expect_snapshot() for warnings.                                                         
                                                            
What went wrong: I didn't properly internalize that instruction when writing the tests. 
I defaulted to the more common expect_error() pattern rather than following the
project-specific convention for snapshot testing of error/warning messages.
```

I personally think this is an improvement from before. There are still some patterns it doesn't pick up. But the alternative would be a really long list of "do this, not this" that would be hard to maintain

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -45,7 +45,7 @@ air format .
 - If there are existing tests, place new tests next to similar existing tests.
 - Strive to keep your tests minimal with few comments.
 - Avoid `expect_true()` and `expect_false()` in favour of a specific expectation which will give a better failure message. A few expectations in newer releases that you might not know about are `expect_all_true()`, `expect_all_equal()`, and `expect_r6_class()`.
-- When testing errors and warnings, don't us `expect_error()` or `expect_warning()`. Instead, use `expect_snapshot(error = TRUE)` for errors and `expect_snapshot()` for warnings because these allow the user to review the full text of the output.
+- When testing errors and warnings, don't use `expect_error()` or `expect_warning()`. Instead, use `expect_snapshot(error = TRUE)` for errors and `expect_snapshot()` for warnings because these allow the user to review the full text of the output.
 - Avoid the `.package` argument to `local_mocked_bindings()`; this modifies the namespace of another package which is not good practice. Instead create a mockable version of the function in the current package. See `?local_mocked_bindings` for more details.
 
 ### Documentation


.claude/CLAUDE.md
```diff
@@ -44,6 +44,8 @@ air format .
 - All new code should have an accompanying test.
 - If there are existing tests, place new tests next to similar existing tests.
 - Strive to keep your tests minimal with few comments.
+- Avoid `expect_true(fn(...))` or `expect_false(fn(...))` - there's almost always a specific expectation that gives better failure messages.
```
hadley:
```suggestion
- Avoid `expect_true()` and `expect_false()` in favour of a specific expectation which will give a better failure message.
```

Maybe worth explicitly including a fewer from newer versions. Something like:

> A few expectations in newer releases that you might not know about are `expect_all_true()`, `expect_all_equal()`, and `expect_r6_class()`

.claude/CLAUDE.md
```diff
@@ -44,6 +44,8 @@ air format .
...
+- Never use `expect_error()` or `expect_warning()`. Instead, use `expect_snapshot(error = TRUE)` for errors and `expect_snapshot()` for warnings.
```
hadley:
```suggestion
- When testing errors and warnings, don't us `expect_error()` or `expect_warning()`. Instead, use `expect_snapshot(error = TRUE)` for errors and `expect_snapshot()` for warnings because these allow the user to review the full text of the output.
```

.claude/CLAUDE.md
```diff
@@ -44,6 +44,8 @@ air format .
...
```
hadley, line 50:
```suggestion
* Avoid the `.package` argument to `local_mocked_bindings()`; this modifies the namespace of another package which is not good practice. Instead create a mockable version of the function in the current package. See `?local_mocked_bindings` for more details.
```

.claude/CLAUDE.md
```diff
@@ -44,6 +44,9 @@ air format .
 - All new code should have an accompanying test.
 - If there are existing tests, place new tests next to similar existing tests.
 - Strive to keep your tests minimal with few comments.
+- Avoid `expect_true()` and `expect_false()` in favour of a specific expectation which will give a better failure message. A few expectations in newer releases that you might not know about are `expect_all_true()`, `expect_all_equal()`, and `expect_r6_class()`.
+- When testing errors and warnings, don't us `expect_error()` or `expect_warning()`. Instead, use `expect_snapshot(error = TRUE)` for errors and `expect_snapshot()` for warnings because these allow the user to review the full text of the output.
```
jennybc:
```suggestion
- When testing errors and warnings, don't use `expect_error()` or `expect_warning()`. Instead, use `expect_snapshot(error = TRUE)` for errors and `expect_snapshot()` for warnings because these allow the user to review the full text of the output.
```

PR #2222, opened 2026-03-11
More options for uncommitted changes

### LINKED ISSUE #1300: Expose alternative strategies for resolving uncommitted changes
e.g. this:

```
There are uncommitted changes, which may cause problems or be lost when we push, pull, switch, or compare branches.
Do you want to proceed anyway?
1: Nope
2: Yes
3: No way
```
Could instead be:

```
What do you want to do?
1. Stash changes (restore with git_stash_pop())
2. Abort this operation
3. Proceed anyway
```

hadley:
Fixes #1300

Currently 100% vibe coded. It seems straightforward enough, but I'm suspicious of the use of `on.exit()` here (although it is a pragmatic way to avoid bigger API changes).


jennybc:
@olivroy I'll still have a look at including some detail about the affected files. But I wanted to go ahead and take this improvement.

PR #2230, opened 2026-04-29
More informative diagnostics from `github_remotes()`

### LINKED ISSUE #1929: Some HTTP 403 errors offer context

### LINKED ISSUE #2229: Better offline diagnostics
Feels like we should be able to tease about some of these situations a little bit more:

```
Error in `github_remotes()`:
✖ Unable to get GitHub info for these remotes: "origin".
ℹ Are we offline? Is GitHub down? Has the repo been deleted?
ℹ Otherwise, you probably need to configure a personal access token (PAT) for "github.com".
```

hadley:
We more carefully generate an error based on the specific error condition, always including the original error for context.

Fixes #2229. Fixes #1929.

### DIFF
--- a/R/github_token.R
+++ b/R/github_token.R
@@ -118,6 +118,11 @@ code_hint_with_host <- function(function_name, host = NULL, arg_name = NULL) {
   glue_chr("{function_name}({arg_hint(host, arg_name)})")
 }
 
+has_pat <- function(api_url = "https://api.github.com") {
+  pat <- tryCatch(gh::gh_token(api_url = api_url), error = function(cnd) "")
+  nzchar(pat)
+}
+
 # workhorse behind gh_token_help() and called, possibly twice, in git_sitrep()
 # hence the need for `scold_for_renviron = TRUE/FALSE`
 # scope determines if "global" or "de_facto" email is checked
--- a/R/github_token.R
+++ b/R/github_token.R
@@ -118,7 +118,7 @@ code_hint_with_host <- function(function_name, host = NULL, arg_name = NULL) {
   glue_chr("{function_name}({arg_hint(host, arg_name)})")
 }
 
-has_pat <- function(api_url = "https://api.github.com") {
+has_pat <- function(api_url = default_api_url()) {
   pat <- tryCatch(gh::gh_token(api_url = api_url), error = function(cnd) "")
   nzchar(pat)
 }
--- a/R/utils-github.R
+++ b/R/utils-github.R
@@ -233,16 +233,17 @@ get_gh_repo <- function(
   remote,
   owner,
   name,
-  api_url = "https://api.github.com",
+  api_url = default_api_url(),
   github_get = TRUE,
   call = caller_env()
 ) {
-  # Don't even try to get parent info for missing remote
+  # Don't even try to get info if we don't know owner/name
+  # This happens when we try and get the parent info for remote that wasn't found
   if (is.na(owner) || is.na(name)) {
     return(list())
   }
 
-  if (identical(github_get, TRUE)) {
+  if (isTRUE(github_get)) {
     # Fail with informative message
     withCallingHandlers(
       gh_repo(owner, name, api_url),
--- a/R/utils-github.R
+++ b/R/utils-github.R
@@ -283,7 +283,7 @@ gh_repo_fail_message <- function(cnd, api_url) {
   } else if (gh_is_rate_limited(cnd)) {
     c(
       "GitHub API rate limit exceeded.",
-      "i" = if (has_path(api_url)) {
+      "i" = if (!has_pat(api_url)) {
         "An authenticated request has a much higher limit than an anonymous
              one; see {.run usethis::gh_token_help()} for advice."
       }
@@ -325,7 +325,7 @@ gh_is_rate_limited <- function(cnd) {
   } else if (inherits(cnd, "http_error_403")) {
     # GitHub also returns 403 (not 429) when you exhaust the primary hourly
     # quota, distinguished from a permissions failure by `x-ratelimit-remaining`.
-    remaining <- cnd$response$headers[["x-ratelimit-remaining"]]
+    remaining <- cnd$response_headers[["x-ratelimit-remaining"]]
     !is.null(remaining) && remaining == "0"
   } else {
     FALSE
--- a/R/utils-github.R
+++ b/R/utils-github.R
@@ -178,46 +178,26 @@ github_remote_list <- function(these = c("origin", "upstream"), x = NULL) {
 github_remotes <- function(
   these = c("origin", "upstream"),
   github_get = NA,
-  x = NULL
+  x = NULL,
+  call = caller_env()
 ) {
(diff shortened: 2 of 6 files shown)


R/utils-github.R
```diff
@@ -178,46 +178,26 @@ github_remote_list <- function(these = c("origin", "upstream"), x = NULL) {
...
-      repo = repo_name,
-      .api_url = api_url
-    )
-  }
-  repo_info <- purrr::pmap(
-    grl[c("repo_owner", "repo_name", "api_url")],
-    get_gh_repo
-  )
+
+  repo_info <- unwrap_purrr(purrr::pmap(
```
hadley, line 186: Explicitly unwrapping the purrr error here since `get_gh_repo()` already includes all the context needed.

R/utils-github.R
```diff
@@ -178,46 +178,26 @@ github_remote_list <- function(these = c("origin", "upstream"), x = NULL) {
...
   # NOTE: these can be two separate matters:
   # 1. Did we call the GitHub API? Means we know `is_fork` and the parent repo.
   # 2. If so, did we call it with auth? Means we know if we can push.
   grl$github_got <- map_lgl(repo_info, \(x) length(x) > 0)
-  if (isTRUE(github_get) && !all(grl$github_got)) {
-    oops <- which(!grl$github_got)
-    oops_remotes <- grl$remote[oops]
-    oops_hosts <- unique(grl$host[oops])
-    ui_abort(c(
-      "Unable to get GitHub info for these remotes: {.val {oops_remotes}}.",
```
hadley, line 214: The new behaviour is to error on the first remote that fails, which is maybe a small downgrade in experience, but I don't think it's significant.

R/utils-github.R
```diff
@@ -233,18 +213,125 @@ github_remotes <- function(
...
+  } else if (identical(github_get, NA)) {
+    # Proceed anyway
+    tryCatch(gh_repo(owner, name, api_url), error = function(cnd) list())
+  } else {
+    # Don't even try
+    list()
+  }
+}
+
+gh_repo_fail_message <- function(cnd, api_url) {
```
hadley, line 270: These errors (and accompanying snapshots) are probably the most important place to spend your review energy.

R/utils-github.R
```diff
@@ -339,9 +426,9 @@ new_github_remote_config <- function() {
   )
 }
 
-github_remote_config <- function(github_get = NA) {
+github_remote_config <- function(github_get = NA, call = caller_env()) {
   cfg <- new_github_remote_config()
-  grl <- github_remotes(github_get = github_get)
+  grl <- github_remotes(github_get = github_get, call = call)
```
hadley, line 432: Ensures that (e.g.) `pr_pull()` reports `pr_pull()` in the error message.

R/github_token.R
```diff
@@ -118,6 +118,11 @@ code_hint_with_host <- function(function_name, host = NULL, arg_name = NULL) {
   glue_chr("{function_name}({arg_hint(host, arg_name)})")
 }
 
+has_pat <- function(api_url = "https://api.github.com") {
```
jennybc:
```suggestion
has_pat <- function(api_url = default_api_url()) {
```

It's been an ongoing process to _not_ hardwire github.com everywhere, so this is more consistent with those efforts. Given how `has_pat()` is used in this PR it doesn't matter, but this feels better as a way to leave things for the future.

R/utils-github.R
```diff
@@ -233,18 +213,125 @@ github_remotes <- function(
...
     map_lgl(parent_info, c("permissions", "push"), .default = NA)
 
   grl
 }
 
+get_gh_repo <- function(
+  remote,
+  owner,
+  name,
+  api_url = "https://api.github.com",
```
jennybc:
```suggestion
  api_url = default_api_url(),
```

As discussed elsewhere

R/utils-github.R
```diff
@@ -233,18 +213,125 @@ github_remotes <- function(
...
+  github_get = TRUE,
+  call = caller_env()
+) {
+  # Don't even try to get parent info for missing remote
```
jennybc: This comment seems to not really describe the `if ()` condition, meaning `remote` != `owner, name`.

Also looks like we can early exit here for `github_get` is `FALSE`.

R/utils-github.R
```diff
@@ -233,18 +213,125 @@ github_remotes <- function(
...
+  if (is.na(owner) || is.na(name)) {
+    return(list())
+  }
+
+  if (identical(github_get, TRUE)) {
```
jennybc:
```suggestion
  if (isTRUE(github_get)) {
```

Nit.

R/utils-github.R
```diff
@@ -233,18 +213,125 @@ github_remotes <- function(
...
+    # Fail with informative message
```
jennybc, line 247: Speaking for myself, maybe a comment about why we need `withCallingHandlers()` (vs. `tryCatch()`) is more useful than this? I did have to read up on this FWIW. Or no comment.

jennybc (approved):
LGTM. I made a few inline comments.

My only source of some discomfort is kind of hard-wiring specific interpretations of specific HTTP codes. Of course everything looks/feel right and we are surfacing the more nitty gritty details of the error. I just wouldn't be surprised if we say something a bit off base or confusing now or in the future. But this is absolutely more and better info than usethis currently provides!

R/utils-github.R
hadley: Comment improved. I think it's clearer to handle the three `github_get()` values in one place below.

R/utils-github.R
hadley: I'm not sure where we document that but you should always use `withCallingHandlers()` when wrapping errors.

PR #2232, opened 2026-05-11
Tickle git more frequently

hadley:
I've been noticing this in Positron during the release process

### DIFF
--- a/R/github.R
+++ b/R/github.R
@@ -220,6 +220,7 @@ use_github_links <- function(overwrite = FALSE) {
     untracked = TRUE,
     paths = "DESCRIPTION"
   )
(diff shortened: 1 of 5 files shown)


jennybc:
Have you noticed that this helps? I feel like I remember some earlier thread about Positron/VS Code's git extension being prone to staleness 🤔

hadley:
I haven't tried it yet, but I had a whole convo with Julia where I thought it was a positron issue and this is supposed to work.

hadley:
I have now tried it and made some changes and confirmed it works 😄

jennybc:
OK I'm happy to take your testimony!

Issue #2226, opened 2026-03-23
Use new roxygen2 `DESCRIPTION` syntax


hadley:
https://github.com/r-lib/roxygen2/pull/1835

JesseAlderliesten:
Using `devtools::document()` after upgrading `roxygen2` from 7.3.3 to 8.0.0 removed `RoxygenNote: 7.3.3` from the `DESCRPTION` and added `Config/roxygen2/version: 8.0.0` instead.

However, `usethis::use_roxygen_md()` re-inserts `RoxygenNote: 8.0.0` to the `DESCRPTION`. I **suspect** that is a bug but I am not sure and came across this issue when looking if someone filed this bug already, so asking here first without filing a new issue.

Re-inserting `RoxygenNote: 8.0.0` happens because `usethis::use_roxygen_md()` calls `usethis:::uses_roxygen()` to check for the use of `roxygen2`, which looks for the just-removed `RoxygenNote` field and does **not** recognise the newly added `Config/roxygen2/version` field, nor the `Roxygen` field that is still in the `DESCRPTION`.

If it is a bug, adjusting `usethis:::uses_roxygen()` from `proj_desc()$has_fields("RoxygenNote")` to something like
`any(usethis:::proj_desc()$has_fields(c("Roxygen", "RoxygenNote", "Config/roxygen2/version")), na.rm = TRUE)` might solve it, but I'm **not** sure *which* of these fields should be checked, or if `markdown = TRUE` should be required in the `Roxygen` field (e.g., as implemented in `usethis:::uses_roxygen_md()`).

I'm using `R` 4.6.0 with `roxygen2` 8.0.0, `usethis` 3.2.1 and `devtools` 2.5.2.

hadley:
Yes, this is a bug. But it feels like a minor bug since you only need to call `use_roxygen_md()` when porting a package that uses very old roxygen2.

jonthegeek:
This also breaks `use_import_from()`

JesseAlderliesten:
> This also breaks `use_import_from()`

The new roxygen style indeed also breaks `use_import_from()` which *is* used a lot (running `devtools::document()` did not solve it):
```
Error in `check_uses_roxygen()`:
✖ Package develcoder does not use roxygen2.
ℹ `use_import_from()` can not work without it.
ℹ You might just need to run devtools::document() once, then try again.
```
So should I make a bugfix and do a pull request @hadley ?

jennybc:
[...]

PR #2234, opened 2026-05-12
`use_import_from()` now also works in packages using `roxygen2` 8.0.0

JesseAlderliesten:
#2226, report by @jonthegeek; fix by @JesseAlderliesten

### DIFF
--- a/R/roxygen.R
+++ b/R/roxygen.R
@@ -78,8 +78,10 @@ uses_roxygen_md <- function() {
 }
 
 uses_roxygen <- function() {
-  any(proj_desc()$has_fields(c("RoxygenNote", "Config/roxygen2/version")),
-      na.rm = TRUE)
+  any(
+    proj_desc()$has_fields(c("RoxygenNote", "Config/roxygen2/version")),
+    na.rm = TRUE
+  )
 }
 
 roxygen_ns_append <- function(tag) {
--- a/R/roxygen.R
+++ b/R/roxygen.R
@@ -72,8 +72,10 @@ uses_roxygen_md <- function() {
 
   roxygen <- desc$get_field("Roxygen", "")
   config_roxygen <- desc$get_field("Config/roxygen2/markdown", "")
-  if (grepl("markdown\\s*=\\s*TRUE", roxygen) ||
-      grepl("TRUE", config_roxygen)) {
+  if (
+    grepl("markdown\\s*=\\s*TRUE", roxygen) ||
+      grepl("TRUE", config_roxygen)
+  ) {
     TRUE
   } else {
     NA
--- a/R/roxygen.R
+++ b/R/roxygen.R
@@ -59,21 +59,18 @@ use_roxygen_md <- function(overwrite = FALSE) {
   invisible()
 }
 
-# FALSE: no Roxygen field and no Config/roxygen2/markdown field
-# TRUE: Roxygen field matches regex targeting 'markdown = TRUE', with some
-#   whitespace slop or Config/roxygen2/markdown field matches '"TRUE"'
+# FALSE: no Roxygen field
+# TRUE: matches regex targetting 'markdown = TRUE', with some whitespace slop
 # NA: everything else
 uses_roxygen_md <- function() {
   desc <- proj_desc()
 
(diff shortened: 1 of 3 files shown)


R/roxygen.R
```diff
@@ -59,26 +59,36 @@ use_roxygen_md <- function(overwrite = FALSE) {
   invisible()
 }
 
-# FALSE: no Roxygen field
-# TRUE: matches regex targetting 'markdown = TRUE', with some whitespace slop
+# FALSE: no Roxygen field and no Config/roxygen2/markdown field
+# TRUE: Roxygen field matches regex targeting 'markdown = TRUE', with some
+#   whitespace slop or Config/roxygen2/markdown field matches '"TRUE"'
 # NA: everything else
 uses_roxygen_md <- function() {
```
hadley, line 65: This is unrelated to the goal of the PR, right? I'd suggest either removing it from this PR, or finishing it off by adding a test.

R/roxygen.R
```diff
@@ -59,26 +59,36 @@ use_roxygen_md <- function(overwrite = FALSE) {
...
   } else {
     NA
   }
 }
 
 uses_roxygen <- function() {
-  proj_desc()$has_fields("RoxygenNote")
+  any(
+    proj_desc()$has_fields(
+      c("Roxygen", "RoxygenNote", "Config/roxygen2/version")
```
hadley: I think this could just be RoxygenNote + Config/roxygen2/version

tests/testthat/test-roxygen.R
```diff
@@ -31,3 +31,17 @@ test_that("use_roxygen_md() finds 'markdown = TRUE' in presence of other stuff",
   expect_no_error(use_roxygen_md())
   expect_true(uses_roxygen_md())
 })
+
+test_that("`uses_roxygen()` recognizes the `Config/roxygen2/version` field", {
+  skip_if_not_installed("roxygen2")
+
+  pkg <- create_local_package()
+  desc::desc_del(keys = c("Roxygen", "RoxygenNote"))
```
hadley: Rather than relying on `create_local_package()` and deleting fields, I think it would be better to create an empty temporary directory and then specifically a minimal description with just the fields you want. Otherwise, when we update `create_local_package()`, this test will be confusing.

Since there isn't an existing test for the RoxygenNote field, I'd suggest testing that as well.

hadley (commented):
Thanks for working on this!

R/roxygen.R
JesseAlderliesten: My reasoning was that `uses_roxygen_md()` would be `FALSE` when `Roxygen: list(markdown = TRUE)` is replaced by `Config/roxygen2/markdown: TRUE`. I now see that **both fields** are present in roxygen 8.0.0-style `DESCRIPTION` files, such that the change to `uses_roxygen_md()` is indeed not needed. So I will cut that out.

tests/testthat/test-roxygen.R
JesseAlderliesten: Hm, `create_local_package()` is also used in the other tests in `test-roxygen.R`, so using that seemed better than trying to get an non-yet-existing, empty writeable temporary directory from scratch (I tried for my own [package](https://github.com/JesseAlderliesten/progutils/blob/master/R/create_tempdir.R) but writing files to it in a way that works both interactively and with devtools::check() gives errors).

How would you set up such a temporary directory?

hadley: `withr::local_tempdir()` 😄

JesseAlderliesten: Thanks! I will check that out and see if I can get that implemented tomorrow

tests/testthat/test-roxygen.R
JesseAlderliesten, line 1: I'm afraid I just wrote over this code when replacing, sorry! If you can revert it, that's fine.

JesseAlderliesten:
@hadley: I reverted to your commit, which is clearer and more compact than mine. One nitpick: the linter complains that `file.path()` should be replaced with `fs::path()`, so you might want to change that.

JesseAlderliesten:
All tests pass :tada: And looks good to me. Thanks for the guidance!

hadley:
Thanks @JesseAlderliesten!

PR #2243, opened 2026-06-10
Add `use_env_var()` helper for setting env vars in `.Renviron`

### LINKED ISSUE #2201: Helper for saving an env var
In my recent LLM workshop, folks had a hard time following by instructions to add environment variables to `.Renviron`. This is possible due to my poor instructions, but I think it would be useful to have a helper that could add a new value. The signature would look something like like `use_env_var("API_KEY")` and the value would be captured using {askpass} or similar.

This would mainly be aimed at people who don't have a lot in the env var (e.g. in my workshop I didn't see anyone with an existing `.Renviron`) so would need relatively little intelligence. I I think it would be reasonable to start by just appending a new`NAME=value` line at the bottom of the file.

gadenbuie:
Closes #2201

## Summary

`use_env_var(name, value = NULL, scope = NULL)` makes it easy to add or update an environment variable in `.Renviron` — the main target audience is workshop attendees and R newcomers who struggle to locate and edit the file manually.

What it does:

- Prompts for the value securely via `askpass::askpass()` when no value is supplied, keeping secrets out of `.Rhistory` and scripts.
- Writes the entry using proper `.Renviron` quoting so the value round-trips correctly across R sessions (double-quoted for most values; unquoted with doubled backslashes for values ending in `\`, such as Windows paths).
- Detects and offers to overwrite existing entries in-place (matching `NAME=value`, `NAME = value`, and `  NAME = value` forms).
- Calls `Sys.setenv()` so the variable is active in the current session immediately, without requiring a restart.
- `scope` defaults to `"project"` when an active project is detected, otherwise `"user"`. This differs from other `edit_r_*()` functions that have a fixed default — the reasoning is that being inside a project during a workshop or AI-assisted session likely signals intent to scope the variable there. Advanced users who want explicit control can pass `scope` directly or use `edit_r_environ()`.

Values that cannot be encoded in `.Renviron` produce a clear error: newlines, `${VAR}` references (expanded by `readRenviron()` with no escape mechanism), and values that end in `\` with surrounding whitespace.

`askpass` is added to `Suggests` (it is a transitive dependency of `gert` and `gh`, so it is effectively always installed).


R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+#' Adds or updates an environment variable in `.Renviron` and immediately
+#' makes it available in the current session via [Sys.setenv()]. The
+#' `scope` defaults to `"project"` when an active project is detected,
+#' otherwise `"user"`.
+#'
+#' @param name Name of the environment variable. Must contain only letters,
+#'   digits, and underscores, and must start with a letter or underscore.
+#' @param value Value to set. If `NULL` (the default), you are prompted to
+#'   enter the value securely using [askpass::askpass()].
+#' @param scope Edit globally for the current **user** (`"user"`), or locally
```
hadley: My concern with writing to the project file is that as soon as you have a single entry in the project file then _nothing_ in the user file will be loaded.

R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+      "x" = "{.val {name}} contains invalid characters.",
+      "i" = "Valid names start with a letter or underscore and contain only
+             letters, digits, and underscores."
+    ))
+  }
+
+  if (is.null(scope)) {
+    scope <- if (possibly_in_proj()) "project" else "user"
+  } else {
+    scope <- match.arg(scope, c("user", "project"))
```
hadley: `arg_match`?

R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+
+  path <- scoped_path_r(scope, ".Renviron", envvar = "R_ENVIRON_USER")
+  lines <- if (file_exists(path)) read_utf8(path) else character()
+  existing_idx <- grep(paste0("^\\s*", name, "\\s*="), lines)
+
+  if (length(existing_idx) > 0) {
+    ui_bullets(c(
+      "i" = "{.envvar {name}} is already defined in {.file {pth(path)}}."
+    ))
+    overwrite <- if (getOption("usethis.overwrite", FALSE)) {
```
hadley: Move `overwrite <- ` inside each branch?

R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+      TRUE
+    } else if (is_interactive()) {
+      ui_yep("Overwrite the existing value for {.envvar {name}}?")
+    } else {
+      FALSE
+    }
+    if (!overwrite) {
+      return(invisible(path))
```
hadley: Should this be an error? I don't have a strong sense either way..

R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+    }
+  }
+
+  if (is.null(value)) {
```
hadley, line 189: Should we allow `value` to ever be non-NULL? I don't love that this might end up recorded in `.Rhistory`.
But maybe this is less dangerous in positron since it doesn't have a history pane?  Either way, I'd love to hear your thoughts.

R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+    ))
+  }
+
+  new_line <- paste0(name, "=", renviron_quote(value))
+
+  if (length(existing_idx) > 0) {
+    lines[existing_idx] <- new_line
+    write_utf8(path, lines)
+  } else {
+    create_directory(path_dir(path))
```
hadley: Under what circumstances would the directory not already exist?

R/edit.R
```diff
@@ -125,6 +125,102 @@ edit_r_environ <- function(scope = c("user", "project")) {
...
+    write_utf8(path, c(lines, new_line))
+  }
+
+  args <- list(value)
+  names(args) <- name
+  do.call(Sys.setenv, args)
```
hadley:
```suggestion
  do.call(Sys.setenv, set_names(list(value), name))
```

hadley: It's a nice touch to set it!

R/edit.R
```diff
@@ -266,3 +362,28 @@ edit_pkgdown_config <- function() {
     invisible(edit_file(path))
   }
 }
+
+renviron_quote <- function(value) {
+  if (grepl("\\$\\{", value)) {
+    ui_abort(c(
+      "{.arg value} contains {.code ${{VAR}}}, which {.file .Renviron} expands on re-read and cannot be stored literally.",
```
hadley, line 369: TIL!

R/edit.R
gadenbuie: I changed this so that we now default to `scope = "user"`. The docs also focus on nudging toward user-level changes.

I had forgotten about the fact that project `.Renviron` causes the user file to be completely ignored; I think you're right that changing the user-level file will lead to the least amount of confusion in the long run for most people.

R/edit.R
gadenbuie: I rewrote this so that we just consult `usethis.overwrite` or call `ui_yep()`, which aborts in a non-interactive use case, which ended up simplifying this code a bit.

R/edit.R
gadenbuie: Yeah it should. I updated this to follow the pattern of `write_over()` and to throw if choosing not to overwrite -- we're not using that helper because I think it's better UX to ask about overwriting before asking for the env var value. It's awkward to paste the value into the message and _then_ be told you're about to overwrite.

R/edit.R
gadenbuie: I included this primarily for situations where the value of the environment variable is found programmatically, e.g. through an API call or something like `keyring::key_get()`. Or in the odd chance that `askpass` isn't available or the user wants to wire up a different collection mechanism.

I updated the docs to explicitly call out that users shouldn't set this value in plain-text code.

I'd find it convenient, but I'm not tied to it and I'd be okay with removing `value` if you have a strong opinion.

hadley: I hadn't thought about the programmatic case. That makes sense to me.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # usethis (development version)
 
+* New `use_env_var()` sets an environment variable in `.Renviron`, prompting for the value securely via `askpass::askpass()` and immediately activating it in the current session via `Sys.setenv()` (#2201, @gadenbuie).
```
hadley: Can you rebase and then put this in the right place? Should be good to merge then.

gadenbuie: 👍 all done!

PR #2244, opened 2026-06-11
Allow `use_git()` to work in new project

### LINKED ISSUE #2242: Allow `use_git()` to work without an active project?
Since `.git` is one of the directories used to establish if something is a project, it's sometimes hard to bootstrap a simple project.

hadley:
Fixes #2242


R/git.R
```diff
@@ -11,17 +15,15 @@
...
     ui_bullets(c("v" = "Initialising Git repo."))
     git_init()
-    # hacky but helps prevent a pop-up in Positron, where early attempts to
-    # interact with a newly created repo lead to:
-    # Git: There are no available repositories
-    # https://github.com/r-lib/usethis/pull/2011#issue-2380380721
-    if (is_positron()) {
-      Sys.sleep(1)
-    }
+    rstudio_git_tickle()
```
hadley, line 26: Since I was in here, I verified that this hack was no longer necessary.

jennybc: I worry that this is an intermittent issue and we're going to get reports of it again. But we can roll the dice.

PR #2250, opened 2026-06-12
Broad reworking of `agents.md`

hadley:
* Try a homegrown skills approach to avoid so much duplication
* Reduce `settings.json` since I think auto mode largely obviates it
* Add a "This package" section that's preserved across `use_tidy_agents()` re-runs
* Give some info about posit assistant.


R/tidy-agents.R
```diff
@@ -52,6 +53,60 @@ use_tidy_agents <- function() {
...
+#'
+#' @param name Name of the skill:
+#'   * `"arg-checking"`: add input checking to a function.
+#'   * `"deprecate"`: deprecate a function or argument.
+#' @export
+#' @examples
+#' \dontrun{
+#' learn_tidy_skill("deprecate")
+#' }
+learn_tidy_skill <- function(name) {
```
jennybc, line 74: Interesting to think about looking up a skill on, say, the usethis pkgdown site. Then falling back to a local file. I'm thinking about ways to combat staleness.

jennybc (approved):
I didn't do a super-thorough review, but in principle LGTM.

jennybc (approved):
I didn't do a super-thorough review, but in principle LGTM.

R/tidy-agents.R
hadley: Yeah, I think we might try that in the future too.

Issue #2238, opened 2026-05-16
`Contributing` uses deprecated `install_dev_deps()` and `devtools::check()` before changes


JesseAlderliesten:
The [`Contributing` document](https://github.com/r-lib/usethis/edit/main/.github/CONTRIBUTING.md) instructs to use `devtools::install_dev_deps()` to install the needed dependencies, but for `devtools >= 2.5.0` this issues the  warning
```
`install_dev_deps()` was deprecated in devtools 2.5.0.
ℹ Please use `pak::local_install_dev_deps()` instead.
```

So `pak::local_install_dev_deps()` could be added.

Also, the guide says to run `devtools::check()` **before** any change is made and does not recommend running it **after** changes are made.
I suppose running `devtools::check()` before making any changes is a good habit to ensure the baseline situation is fine, but running it after making changes seems helpful to prevent R cmd check failing when making a pull requests.

PR #2252, opened 2026-06-15
Prefer `pak::local_install_dev_deps()` over `devtools::install_dev_deps()`

JesseAlderliesten:
Prefer `pak::local_install_dev_deps()` over `devtools::install_dev_deps()` to install development dependencies because the latter was deprecated in `devtools` 2.5.0. Fixes first part of #2238.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -20,7 +20,7 @@ If you’ve found a bug, please file an issue that illustrates the bug with a mi
 
 *   Fork the package and clone onto your computer. If you haven't done this before, we recommend using `usethis::create_from_github("r-lib/usethis", fork = TRUE)`.
 
-*   Install all development dependencies with `devtools::install_dev_deps()`, and then make sure the package passes R CMD check by running `devtools::check()`. 
+*   Install all development dependencies with `pak::local_install_dev_deps()` (or, for `devtools` prior to version 2.5.0, with `devtools::install_dev_deps()`), and then make sure the package passes R CMD check by running `devtools::check()`. 
     If R CMD check doesn't pass cleanly, it's a good idea to ask for help before continuing. 
 *   Create a Git branch for your pull request (PR). We recommend using `usethis::pr_init("brief-description-of-change")`.
 


jennybc (commented):
The primary file to edit in this "area" is the template:

https://github.com/r-lib/usethis/blob/main/inst/templates/tidy-contributing.md

This what gets placed by `use_tidy_contributing()`. So you'd update the template and then re-apply `use_tidy_contributing()` to usethis itself.

(As you may have guessed, the contributing instructions don't seem to be that load-bearing these days and are somewhat neglected ... but that doesn't mean we shouldn't improve them!)

JesseAlderliesten:
Should running `devtools::check()` also be indicated **after** a change is made?

Currently the file only says to run it **before** any change is made and ask for help if it does not pass (presumably because then it will also fail after changes are made). But testing once locally after making changes (but before pushing) seems more efficient to me than running the complete GitHub Actions on the pull request to discover they all fail if, e.g., a bug was introduced.

> The primary file to edit in this "area" is the template:
> 
> https://github.com/r-lib/usethis/blob/main/inst/templates/tidy-contributing.md
> 
OK, then I can close this pull request and create another one where I change the template you indicate, and include the instruction to also run `devtools::check()` **after** a change is made if that makes sense to you.

> (As you may have guessed, the contributing instructions don't seem to be that load-bearing these days and are somewhat neglected ... but that doesn't mean we shouldn't improve them!)

Searching GitHub for `"For a detailed discussion on contributing to this and other tidyverse packages"` gives 114 hits BTW, so this file (or its template) has quite some reach ;)

jennybc:
> Searching GitHub for "For a detailed discussion on contributing to this and other tidyverse packages" gives 114 hits BTW, so this file (or its template) has quite some reach ;)

Yeah, it's probably been placed in almost every tidyverse/r-lib package. But I just mean that I'm not sure how often it gets *read* and followed 😅 

Yes adding a reminder to check the package before making the PR is also good. I think advice to check before and after making changes is good for the typical audience.

PR #2255, opened 2026-06-27
fix: prevent `use_standalone()` from downgrading version constraints

eitsupi:
## Problem

`use_standalone()` could inadvertently lower a package's version constraint in `DESCRIPTION`. When the standalone file's `imports:` field specifies a lower minimum version than what the package already requires, `use_dependency()` overwrites the existing (higher) constraint.

For example, if `DESCRIPTION` already contains `rlang (>= 1.1.0)` and a standalone file declares `imports: rlang (>= 1.0.0)`, calling `use_standalone()` would silently downgrade the constraint to `>= 1.0.0`.

This happens because `use_dependency()` always writes the new version when the version differs and the dependency type is unchanged — regardless of whether the new version is higher or lower than the existing one.

## Fix

Add a `downgrade` argument to `use_dependency()` and `use_package()`:

- `downgrade = TRUE` (default) preserves the existing behaviour, allowing the version to be lowered.
- `downgrade = FALSE` silently ignores a `min_version` that is lower than the currently recorded version, emitting an informational message instead.

`use_standalone()` now passes `downgrade = FALSE` when recording its `imports:` dependencies.

## Changes

- `use_dependency()`: new `downgrade` argument (with `check_bool()` validation)
- `use_package()`: new `downgrade` argument, threaded through to `use_dependency()`
- `use_standalone()`: passes `downgrade = FALSE` to `use_package()`
- New test and snapshot for the `downgrade = FALSE` behaviour

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # usethis (development version)
 
+* `use_package()` gains a `downgrade` argument. When set to `FALSE`, a `min_version` lower than the currently recorded version is silently ignored rather than being written to `DESCRIPTION`. `use_standalone()` now passes `downgrade = FALSE` when recording its `imports:` dependencies, preventing a standalone file from inadvertently lowering a version constraint the package author has intentionally set higher.
 * `create_from_github()` now installs package dependencies by default, so you're set up to immediately start working on the package. Use `install_dependencies = FALSE` to suppress (#2186).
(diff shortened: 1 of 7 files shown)


jennybc:
Thanks for this PR! In the process of reviewing it, I ended up working out a slightly different approach, so I'm merging that. But I appreciate the motivation provided by this PR.

PR #2261, opened 2026-09-02
Agent/skill tweaks

hadley:
* List available skills if no input to `learn_tidy_skill()`
* Add a setup skill
* Remind agents that air is bundled with Positron

Our own agents file is getting kind of long, but these are all hard won lessons 😐

### DIFF
--- a/inst/AGENTS.md
+++ b/inst/AGENTS.md
@@ -1,5 +1,7 @@
 ## This package
 
+<!-- Insert package-specific content here. use_tidy_agents() will preserve this section when updating the rest of the file. -->
+
 ## Package development
 
 ### Key commands
@@ -90,7 +92,7 @@ There are three possible ways to run code, listed in rough order of desirability
 
 - Do you need to deprecate a function or argument? Read `usethis::learn_tidy_skill("deprecate")`.
 - Are you adding input checking to an existing function or writing a new exported function? Read `usethis::learn_tidy_skill("arg-checking")`.
-- Are you creating a new package? Read `usethis::learn_tidy_skill("setup")`.
+- Are you creating a new package? Read `usethis::learn_tidy_skill("package-setup")`.
 
 ## Git
 
--- a/inst/AGENTS.md
+++ b/inst/AGENTS.md
@@ -1,7 +1,5 @@
 ## This package
 
-<!-- Insert package-specific content here. use_tidy_agents() will preserve this section when updating the rest of the file. -->
-
 ## Package development
 
 ### Key commands
@@ -48,7 +46,7 @@ There are three possible ways to run code, listed in rough order of desirability
 ### Code style
 
 - Follow the tidyverse style guide
-- Always run `air format .` after generating code.
+- Always run `air format .` after generating code. (air is bundled with Positron so look there if you can't otherwise find it.)
 - Use the base pipe operator (`|>`), not the magrittr pipe (`%>%`).
 - Use `\() ...` for single-line anonymous functions. For all other cases, use `function() {...}`.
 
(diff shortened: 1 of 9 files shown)


inst/skills/setup.md
```diff
@@ -0,0 +1,14 @@
+# Package setup
+
+When setting up a package for the first time, prefer using usethis functions rather than memorised solution. 
+
+In general, you will want to follow these steps.
+
+* Call `usethis::use_testthat()` to set up unit tests
+* Call `devtools::document()` to generate `NAMESPACE` and update docs
+* Ask the use which license they want to use, then call `usethis::use_mit_license()`, `usethis::use_proprietary_license()` etc.
```
jennybc:
```suggestion
* Ask the user which license they want to use, then call `usethis::use_mit_license()`, `usethis::use_proprietary_license()` etc.
```

inst/skills/setup.md
```diff
@@ -0,0 +1,14 @@
...
+* Create package documentation with `usethis::use_package_doc()`.
+* Take a depedency on rlang with `usethis::use_package("rlang")`. Import all functions from rlang into the `NAMESPACE` with `@import rlang`.
```
jennybc:
```suggestion
* Take a dependency on rlang with `usethis::use_package("rlang")`. Import all functions from rlang into the `NAMESPACE` with `@import rlang`.
```

Do we want to specify that `@import rlang` belongs in the fenced parking lot in that package-level documentation file? E.g.

```
## usethis namespace: start
#' @import rlang
## usethis namespace: end
```

inst/skills/setup.md
```diff
@@ -0,0 +1,14 @@
...
+* Call `use_news_md()` to set up a news file.
+* Check the package with `devtools::check()` and ensure there are no issues
+* Make an initial git commit
```
jennybc: I personally am pretty firm that I don't want the agent making commits for me FWIW. I guess they are constantly reconciling contradictory instructions (and ignoring instructions), so there's that.

inst/AGENTS.md
```diff
@@ -48,7 +46,7 @@ There are three possible ways to run code, listed in rough order of desirability
 ### Code style
 
 - Follow the tidyverse style guide
-- Always run `air format .` after generating code.
+- Always run `air format .` after generating code. (air is bundled with Positron so look there if you can't otherwise find it.)
```
jennybc, line 51: I'm so glad we're making it more and more likely that the Air inside Positron can and will be used (vs. a 2nd external installation). The air extension putting the cli on the PATH for Positron-owned terminals/processes is definitely helping.

inst/skills/package-setup.md
```diff
@@ -0,0 +1,14 @@
+# Package setup
```
jennybc, line 1: Weakly held opinion: name after `create_package()`? I.e. `create-package.md` and "Create a package"? I'm just thinking that "setup" is sort of generic and doesn't distinguish between a package and other sorts of projects.

jennybc (approved):

inst/AGENTS.md
```diff
@@ -1,7 +1,5 @@
 ## This package
 
-<!-- Insert package-specific content here. use_tidy_agents() will preserve this section when updating the rest of the file. -->
-
```
hfrick, line 4: Is this removal intentional?

inst/skills/setup.md
hadley: It seems to figure that out with out prompting for me.

inst/skills/setup.md
hadley: I'd be happy to weaken this to "initialise a git repo"

inst/AGENTS.md
hadley: Doh

inst/skills/package-setup.md
hadley: Good idea

PR #2270, opened 2026-09-09
Suggest writing a temporary R script on Windows

### LINKED ISSUE #2259: Recommendations for AGENTS.md aimed at Windows users
Hi, I've been playing with your AGENTS.md a bit and had two suggestions, based on using this on Windows.

1. "Otherwise, use Rscript -e "code"." is frequently bad advice for Windows. It will very often segfault if code is more than two lines. Instead, the following seems to work well:

> On Windows, do not use `Rscript -e '...'` for multiline or complex R code because it may segfault. Instead, write a temporary `.R` file and run `Rscript path/to/file.R`.

2. Rtools no longer needs to be updated at the same frequency, so there is no Rtools 4.6. Often, if there's a compiler failure of any kind, Codex and Claude both point to mismatched toolchains. Adding the following seems to suppress that search:

> R 4.6.1 and Rtools 4.5 are correctly paired; there is no Rtools 4.6.

Both of these are oriented around fixing expensive (token-wise) side quests.

jennybc:
Closes #2259

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,7 +41,9 @@ There are three possible ways to run code, listed in rough order of desirability
   available, use that. Note that `mcp__r__repl` uses a sandbox that blocks
   network requests and reads/writes outside of the current directory.
 
-- Otherwise, use `Rscript -e "code"`.
+- Otherwise, use `Rscript -e "code"`. On Windows, `Rscript -e` can segfault on
+  multiline or complex code; in that case, write it to a temporary `.R` file
+  and run `Rscript path/to/file.R`.
 
 ### Code style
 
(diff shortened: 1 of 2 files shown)


jennybc:
I'm not acting on the point about Rtools in #2259. Not because I have particular objections, but more because it does not naturally fit anywhere in the current `AGENTS.md`. So I'm going to let that idea age a bit, but I'd be happy to revisit in the future. I need to release usethis and I find the point about `Rscript` very compelling.

In the meantime, anyone who finds this ("R 4.6.1 and Rtools 4.5 are correctly paired; there is no Rtools 4.6.") especially advantageous could put it in the section for package-specific content or in a user-level file of agent instructions.

PR #2278, opened 2026-09-17
Challenge user when setting up pkgdown site on non-main branch

### LINKED ISSUE #2065: Empty github pages: Improve documentation on use_pkgdown_github_pages()
**Problem:** Running `use_pkgdown_github_pages()` in a different branch than root leads to correct github workflow which however create an empty site and leads to a 404 error message when accessing the site.

**Solution:** The documentation in `use_pkgdown_github_pages()` should inform the user that the changes need to be pushed to the main branch of the repository. 

**How to reproduce** Run `use_pkgdown_github_pages()` in a branch other than the root branch for a package where pkdown is not yet configured.

Usethis is great because it enables users without deep understanding of the underlying processes - like me - to do really useful things. In order to test out pkdown deployment, I first wanted to do so in a different branch to prevent any hiccups. Of course this lead to a missing `_pkgdown.yml` in the main branch so that the github workflow created an empty site. Because the process did not produce an error message and the github workflow ran successful, it wasnt immediately obvious what the problem was to me. Pushing the changes to the main branch fixed the problem.

Thanks a lot for your great work!!

mhausinger:
Closes #2065.
Closes #2172. *Oops, another PR from a dev day!*

Copies over functionality from pr_init to provide warning message when running use_pkgdown (and correspondingly for use_pkgdown_github_pages) that warns user that this will run into problems when running the function in a non-main branch. 

Tested manually by creating a dummy package. When checking out a branch and running use_pkgdown_github_pages(), it displays a challenge. When switching back to main and running the function again, it runs normally.


R/pkgdown.R
```diff
@@ -31,6 +31,21 @@ use_pkgdown <- function(config_file = "_pkgdown.yml", destdir = "docs") {
...
+  cfg <- github_remote_config(github_get = NA)
+  check_for_bad_config(cfg)
+  tr <- target_repo(cfg, ask = FALSE)
+  online <- is_online(tr$host)
+
+  default_branch <- if (online) {
+    git_default_branch_(cfg)
+  } else {
+    guess_local_default_branch()
+  }
```
hadley: Can you please put the assignment inside the if branches?

hadley (commented):
Thanks for working on this! I spotted one small code style issue.

jennybc:
There's a bigger issue here, which I now think this logic should be in `use_pkgdown_github_pages()` *before* the call to `use_pkgdown()`. This observation is based on me exploring why the tests fail and what that tells us.

jennybc:
The format-suggest workflow is just broken right now, so we can set that failure aside. The other failures are about downloading Quarto, which is also not our problem. I think this is good now, otherwise.

Thanks!

PR #2279, opened 2026-09-17
Add checklist to docs for use_upkeep_issue() and use_release_issue()

### LINKED ISSUE #1862: Could the manual page of use_upkeep_issue show the checklist?

trevinflick:
Fixes #1862

Programmatically generates roxygen chunks containing the checklist for `use_upkeep_issue()` and `use_release_issue()`.

Also added a parameter for `upkeep_checklist()` to remove the footer.

I didn't include any tests since the main changes were documentation.


hadley:
Thanks for working on this! I think these would be a little better if they appeared as code in the rendered docs. Does that make sense?

trevinflick:
That makes sense. Let me see if I can make those changes.

